PR #76, opened 2015-01-06
Fix one failing test case

Randy-Ma:
<h4>Summary</h4>

In a recent change, GetParameterValueByName will return Element other than an element id. Consequently, the related test case need to be changed.

This submission fixes one failing test case by getting the id from the returned element.

@Benglin 
PTAL

### DIFF
--- a/test/Libraries/RevitNodesTests/Elements/ElementTests.cs
+++ b/test/Libraries/RevitNodesTests/Elements/ElementTests.cs
@@ -36,9 +36,9 @@ namespace RevitNodesTests.Elements
 
             DocumentManager.Regenerate();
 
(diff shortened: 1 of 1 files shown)


test/Libraries/RevitNodesTests/Elements/ElementTests.cs
```diff
@@ -36,9 +36,9 @@ public void SetParameterByName_Element_CanSuccessfullySetMaterialByElement()
 
             DocumentManager.Regenerate();
 
-            var elemId1 = ele.GetParameterValueByName(paramName);
+            var elemId1 = ele.GetParameterValueByName(paramName) as Element;
```
sharadkjaiswal: good to rename elemId1 to something like element

PR #77, opened 2015-01-06
Fix failing test cases

Randy-Ma:
<h4>Summary</h4>

For nodes derived from ElementsQueryBase, the first time the nodes are read in from .dyn files. The ForceReExecuteOfNode is set to false which is not correct as at this time, because the nodes are new and need to be executed. 

This submission makes the change to use the base class property.

@Benglin 
PTAL


Benglin:
Thanks for the explanation @Randy-Ma, as discussed this LGTM!

PR #80, opened 2015-01-06
Fix more failing test cases

Randy-Ma:
<h4>Summary</h4>

This submission fixes the failing test cases caused by recent restructuring. It will modify the code slightly to accommodate to the new functions. This will fix quite a lot of remaining failing test cases.

@Benglin 
PTAL


Benglin:
Thanks @Randy-Ma, changes LGTM.

PR #83, opened 2015-01-08
Categorize 4 test cases to the Failure category

Randy-Ma:
<h4>Summary</h4>

This is to categorize 4 failing test cases to the Failure category.

@Benglin 
PTAL


Benglin:
Thanks @Randy-Ma. As discussed this is per your discussion with @pboyer and that we agreed to have these tests marked as `Failure` for now. LGTM!

PR #89, opened 2015-01-13
Fix MAGN-940 [Localization] user gets exception in sample on OverrideColorInView node

Randy-Ma:
<h4>Summary</h4>

The original implementation is trying to compare the pattern name with a string in English. But as the name will be different in different localization versions, which will not always be "Solid Fill", the obtained pattern element may be null. Then an exception will be thrown later if it is null and used.

This submission fixes the issue by using the IsSolidFill property to check whether the pattern element is a kind of solid fill.

@ke-yu 
PTAL


ke-yu:
LGTM.

jnealb:
@Randy-Ma @ke-yu @kronz @RodRecker Why is this merged into Revit2014 only?

PR #90, opened 2015-01-14
2014: Update RaaS Binaries for changes to NodeModel, Re-enable tests

pboyer:
### Purpose

There are a number of `NodeModel` subclasses in DynamoRaaS.  When `NodeModel` was modified to no longer require a `WorkspaceModel` in the constructor, these nodes would no longer compile, nor did they have run-time compatibility with Dynamo when DynamoRaaS is dynamically loaded.

### Fix

Centrally, I've updated these nodes to remove the `WorkspaceModel` reference.  This required a number of internal changes to these nodes in order to obtain the required dependencies.  

For `CloudRender`, for example, access to the WPF `Dispatcher` was required.  This was obtained by using a `INodeViewCustomization` class, who obtains a reference to the `View.`  

You can see the code changes in the closed source repository here:  

https://github.com/DynamoDS/DynamoRaaS/commit/b9988b1d97c517910bd18d9f83d250bd8e5d7234

### Tests

All the tests now run:

renderingtests

I've also experimentally done a few renders with success.

### Reviewer

@Randy-Ma


Randy-Ma:
Great! LGTM!

PR #93, opened 2015-01-14
2016: Update RaaS Binaries for changes to NodeModel, Re-enable tests

pboyer:
See #90


Randy-Ma:
How about only update the binaries for now?

Randy-Ma:
I mean we need to remove other lines similar with the one line removed. Other than this, LGTM!

PR #95, opened 2015-01-15
Updated Revit codes to align with node re-execution fixes in Dynamo

aparajit-pratap:
Refer PR: https://github.com/DynamoDS/Dynamo/pull/3523

### Reviewers ###
@Benglin 
@Randy-Ma

### DIFF
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -426,9 +426,9 @@ namespace Dynamo.Applications.Models
             foreach (var ws in Workspaces.OfType<HomeWorkspaceModel>())
             {
                 foreach (var node in ws.Nodes)
-                    node.ForceReExecuteOfNode = true;
+                    node.MarkNodeAsModified();
 
-                ws.OnAstUpdated();
+                ws.OnNodesModified();
                 
                 foreach (var node in ws.Nodes)
                 {
@@ -482,8 +482,7 @@ namespace Dynamo.Applications.Models
             var nodes = ElementBinder.GetNodesFromElementIds(deleted, CurrentWorkspace, EngineController);
             foreach (var node in nodes)
             {
-                node.ForceReExecuteOfNode = true;
-                node.OnAstUpdated();
+                node.OnNodeModified(forceExecute:true);
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -426,9 +426,9 @@ private void ResetForNewDocument()
             foreach (var ws in Workspaces.OfType<HomeWorkspaceModel>())
             {
                 foreach (var node in ws.Nodes)
-                    node.ForceReExecuteOfNode = true;
+                    node.MarkNodeAsModified();
```
Benglin: I think resetting for a new Revit document we need all nodes to be force re-executed, no? In which case we need `MarkNodeAsModified(forceExecute: true)`. I'm just being afraid that removing that will cause the nodes to not re-execute.

Benglin:
Other than that one point above, LGTM.

Benglin:
Need another PR for R2015 too (which is to go in almost simultaneously).

src/DynamoRevit/Models/RevitDynamoModel.cs
Randy-Ma: Is the function - MarkNodeAsModified - defined anywhere?

test/Libraries/RevitIntegrationTests/SelectionTests.cs
```diff
@@ -84,7 +84,7 @@ public void SelectionDocModificationSync()
 
             TransactionManager.Instance.ForceCloseTransaction();
 
-            Assert.AreEqual(true, selectNode.ForceReExecuteOfNode);
+            Assert.AreEqual(true, selectNode.NeedsForceExecution);
```
Randy-Ma: Is NeedsForceExecution defined?

Randy-Ma:
Except for the two points mentioned, LGTM!

Randy-Ma:
@aparajit-pratap 
I think there may be one pending request which need to be merged before this one is merged. So be careful about this! Otherwise, it may cause a build error.

src/DynamoRevit/Models/RevitDynamoModel.cs
Randy-Ma: agree

Benglin:
You're right, @Randy-Ma, this pull request must only be merged after [its prerequisite](https://github.com/DynamoDS/Dynamo/pull/3523). Also, @aparajit-pratap you'll need to create another PR for Revit 2015 since that branch is exercised by the build machine. When all three pull requests are validated, they should be merged almost simultaneously, with 3253 going in first.

src/DynamoRevit/Models/RevitDynamoModel.cs
aparajit-pratap: Prior to Steve's changes, the `RequiresRecalc` flag was set to `true`, which is why I let it be. However, I'll make the change as suggested. Don't see any harm in doing so.

PR #102, opened 2015-01-19
Fix for MAGN-5804, Changed String.FromObject to "String From Object".

riteshchandawar:
<h3>Problem & Solution</h3>
After opening sample Help>Samples>Revit>Revit_PlaceFamiliesByLevel_Set Parameters, **String.FromObject** was throwing "Warning: This node is obsolete, please use "String from Object" node.,   so replaced old node with new "String From Object".

http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-5804
**Old Node:** Dynamo.Nodes.DSFunction type="Dynamo.Nodes.DSFunction"
**New Node:** DSCoreNodesUI.StringNodes.FromObject type="DSCoreNodesUI.StringNodes.FromObject"

**Note:** Old node was having Lacing option and new node is not having Lacing option, is that a defect?

<h3>Reviewer</h3>
- [x] @ikeough 

<h3>Need merge to Branches</h3>
- [ ] Revit2015 
- [ ] Revit2016

For more details...
image


riteshchandawar:
@kronz : Please take a look at changes.

ikeough:
@riteshchandawar LGTM. Please speak with @ke-yu about the lacing issue. I believe he updated these nodes.

riteshchandawar:
@ke-yu PTAL.

ke-yu:
Now "String From Object" is a UI node, instead of a ds function node, so it doesn't have lacing. LGTM.

kronz:
@ke-yu can you say a little more what you mean by it's a UI node,not a ds function and therefore doesn't have lacing?  It doesn't have lacing or CAN'T have lacing? Either way, the behavior seems right, and there isn't any way that a single input node would be able to do anything with lacing anyway.  thanks
Z

PR #103, opened 2015-01-21
remove NXLT tags for Revit

zora-wang:
Please help to merge the untagged changed files  Thanks


Benglin:
Thanks for removing these, @zora-wang!

PR #104, opened 2015-01-21
remove tags in localLization2015

zora-wang:
update untagged files


Benglin:
Thanks for fixing these, @zora-wang.

PR #114, opened 2015-01-23
Fix MAGN-5923

Randy-Ma:
<h4>Summary</h4>

The reason for MAGN-5923 is the same as MAGN-5769 which is fixed by:
https://github.com/DynamoDS/Dynamo/pull/3531

I am taking the fix there manually to fix the issue here.

@Benglin 
PTAL


Benglin:
LGTM!

PR #115, opened 2015-01-23
Fix MAGN-5839

Randy-Ma:
<h4>Summary</h4>

This is to make a similar fix for MAGN-5839 which is done in the following pull request:
https://github.com/DynamoDS/Dynamo/pull/3568

@Benglin 
PTAL


Benglin:
I said it before and I will say it again -- LGTM! :)

PR #123, opened 2015-01-26
Fix MAGN-5925

Randy-Ma:
<h4>Summary</h4>

This is to make the similar change based on the following pull request:
https://github.com/DynamoDS/Dynamo/pull/3554

@Benglin 
PTAL


Benglin:
LGTM!

PR #125, opened 2015-01-27
Adjustment to the dispose of RevitServicesUpdater

Randy-Ma:
<h4>Summary</h4>

The singleton instance is to be disposed when the addin application is shutdown. If it is disposed when the Dynamo view is closed, the next time either an error will occur or the singleton instance need to be reinitialized.

@Benglin 
PTAL


Benglin:
As per discussed, LGTM!

PR #126, opened 2015-01-27
Build fix to correct the reference path

sharadkjaiswal:
DynamoCore components should be referenced using $(DYNAMOAPI)
environment variable rather than using relative path, because everyone
may not have the same folder structure for Dynamo and DynamoRevit.


sharadkjaiswal:
@Benglin Please take a look.

Benglin:
Thanks for fixing this, LGTM!

PR #135, opened 2015-01-30
MAGN-6095 Placing Get Family Parameter node crashes Dynamo

ikeough:
This PR is the Revit component of [MAGN-6095](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6095).

This PR depends on [this Dynamo Core PR](https://github.com/DynamoDS/Dynamo/pull/3655).

PTAL:
- [x] @pboyer

FYI:
@Randy-Ma

Requires Merge:
- [ ] Revit2014
- [ ] Revit2016

### DIFF
--- a/src/Libraries/RevitNodesUI/FamilyInstanceParametersNodeViewCustomization.cs
+++ b/src/Libraries/RevitNodesUI/FamilyInstanceParametersNodeViewCustomization.cs
@@ -1,11 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using DSCoreNodesUI;
-using Dynamo.Wpf;
-using Dynamo.Wpf.Nodes;
-using DSRevitNodesUI;
+﻿using DSRevitNodesUI;
 
 namespace Dynamo.Wpf.Nodes.Revit
 {
--- a/src/Libraries/RevitNodesUI/RevitDropDown.cs
+++ b/src/Libraries/RevitNodesUI/RevitDropDown.cs
@@ -212,29 +212,22 @@ namespace DSRevitNodesUI
 
         private Element GetInputElement()
         {
-            /*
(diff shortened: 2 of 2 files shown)


ikeough:
@pboyer @lukechurch I'm vaguely remembering why this was problematic now. Because we store the EngineController on the node model, we're in a weird place if the EngineController gets nulled for some reason. 

I propose we change this node, from taking a family type element to taking a string, which is the name of the family type. That way we can avoid all of this hassle, and just call a lookup method in Revit to find the family type, and get its parameters.

kronz:
@ikeough Wait, are you saying that users will have to do something like this to actually use the node?

Or that you are going to do a "to string" operation behind the scenes?

lukechurch:
@ikeough I guess we could do that - we'll just need to make sure that the selection node internally remembers the typeID so that if the file gets used in a different locale the family type node will generate the appropriate newly localised string?

jnealb:
@lukechurch @ikeough @pboyer QA is watching this!

lukechurch:
@ikeough IIRC, Fix tests here => Get the tests to compile so we can start running them.

And I think that as the other changes in this PR. I think this one was removed in order to get the main library to compile. At this stage I was still trying to put the basic infrastructure back together again. Should be pretty easy to verify - if you remove the comment blocks, does the node compile?

kronz:
Just to be clear . . . requiring the user to put a String From Object into the workflow is not ok.

ikeough:
@pboyer Please review this as a fix to the node AS IS, not as a change to the node's functionality per the discussion above. Let's get this fix in to solve the regression.

pboyer:
Keep in mind that `NodeModel`, in the general case, does not hold an `EngineController` reference.  Only this node does and it is set by the `NodeViewCustomization`.  This feels like a hack, but many of the alternatives (as we discussed when you were last in Waltham) are painful to implement.

pboyer:
LGTM!

PR #139, opened 2015-02-03
MAGN-6033 Exception on opening Revit Samples files on Revit 2016.

ikeough:
Revit 2016 is much more stringent about calling API methods outside of an "API context." File opening was not being handled inside a Transaction, and the API was throwing an exception at the first Revit API call inside the file opening operation. This PR does the following:

- Wraps the DynamoModel's OpenFileImpl method in an IdlePromise to ensure that file opening is done in the "API context."

- Adds an EnsureInTransaction call to the Element.SafeInit method. Although not the focus of this PR, i could not test any Revit element creation through Dynamo as all element creation failed due to the out-of-context InvalidOperation exception when attempting to create elements.

This PR requires:
https://github.com/DynamoDS/Dynamo/pull/3672

PTAL:
- [x] @Randy-Ma

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -32,6 +32,8 @@ namespace Revit.Elements
         /// <param name="init"></param>
         protected void SafeInit(Action init)
         {
+            TransactionManager.Instance.EnsureInTransaction(DocumentManager.Instance.CurrentDBDocument);
+
             var elementManager = ElementIDLifecycleManager<int>.GetInstance();
             var element = ElementBinder.GetElementFromTrace<Autodesk.Revit.DB.Element>(Document);
             int count = 0;
--- a/src/Libraries/RevitServices/Transactions/TransactionManager.cs
+++ b/src/Libraries/RevitServices/Transactions/TransactionManager.cs
@@ -199,19 +199,19 @@ namespace RevitServices.Transactions
     {
         public TransactionHandle EnsureInTransaction(TransactionWrapper wrapper, Document document)
         {
-            TransactionManager.Log("EnsureInTransaction - AUTO STRAT: Starting new Transaction");
+            TransactionManager.Log("EnsureInTransaction - AUTO START: Starting new Transaction");
             return !wrapper.TransactionActive ? wrapper.StartTransaction(document) : wrapper.Handle;
         }
 
         public void TransactionTaskDone(TransactionHandle handle)
         {
-            TransactionManager.Log("TransactionTaskDone - AUTO STRAT: Preserving Transaction");
+            TransactionManager.Log("TransactionTaskDone - AUTO START: Preserving Transaction");
             //Do nothing in automatic, continue using the same transaction.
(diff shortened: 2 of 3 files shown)


src/Libraries/RevitServices/Transactions/TransactionManager.cs
```diff
@@ -199,19 +199,19 @@ public class AutomaticTransactionStrategy : ITransactionStrategy
     {
         public TransactionHandle EnsureInTransaction(TransactionWrapper wrapper, Document document)
         {
-            TransactionManager.Log("EnsureInTransaction - AUTO STRAT: Starting new Transaction");
+            TransactionManager.Log("EnsureInTransaction - AUTO START: Starting new Transaction");
```
Randy-Ma: +1

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -32,6 +32,8 @@ public abstract class Element : IDisposable, IGraphicItem, IFormattable
         /// <param name="init"></param>
         protected void SafeInit(Action init)
         {
+            TransactionManager.Instance.EnsureInTransaction(DocumentManager.Instance.CurrentDBDocument);
```
Randy-Ma: @ikeough 
I am more curious about why there is not an active transaction even when we are going to create an element. Does this mean when a task is running, no transactions have been opened? Or does this mean some one has closed the transaction without reopening it?

But as EnsureIntransaction is just ensuring that there is an active transaction, it will make the Revit nodes work. I will agree with the changes with the questions in mind.

src/Libraries/RevitServices/Transactions/TransactionManager.cs
Randy-Ma: Do we need to localize the messages?

src/Libraries/RevitNodes/Elements/Element.cs
ikeough: @Randy-Ma This is a good point. I'll double check that the task is opening a transaction, but we will get into trouble if another operation closes the transaction before we get here.

src/Libraries/RevitServices/Transactions/TransactionManager.cs
ikeough: @Randy-Ma Not sure. I don't know that this is a user facing message. If it is, I'll localize it.

Randy-Ma:
@ikeough 

The change to update the logging message is an improvement and adding a check to ensure there is a live transaction is also an improvement. I think my concern is minor and should not block the pull request because the exception is more important! Base on this, LGTM!

Randy-Ma:
But be sure to first merge https://github.com/DynamoDS/Dynamo/pull/3672

src/Libraries/RevitServices/Transactions/TransactionManager.cs
ikeough: @Randy-Ma This string goes into the log. It's not visible in the UI. I'm leaving it un-localized.

PR #142, opened 2015-02-04
MAGN-4858 Adaptive components placed at origin when points are null

kiki-zhang:
The change includes a fix of MAGN-4858.
Root Cause: For creating a adaptive component, multiple RevitAPI are
called in one transaction. The first step is to create a original
component, and then call another RevitAPI to set its position. The issue
happens because an exception throws out in the process of setting
position due to the null point, which finally leads to the uncompleted
creation.
Solution: Using a subtransaction to roll back the creation if an
exception happens.

### DIFF
--- a/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
+++ b/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
@@ -13,7 +13,6 @@ using RevitServices.Persistence;
 using RevitServices.Transactions;
 using Point = Autodesk.DesignScript.Geometry.Point;
 using Reference = Autodesk.Revit.DB.Reference;
-using Surface = Autodesk.DesignScript.Geometry.Surface;
 
 namespace Revit.Elements
 {
--- a/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
+++ b/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
@@ -76,7 +76,6 @@ namespace Revit.Elements
         /// <param name="fs">FamilySymbol to place</param>
         private void InitAdaptiveComponent(Point[] pts, FamilySymbol fs)
         {
-
             // if the family instance is present in trace...
             var oldFam =
                 ElementBinder.GetElementFromTrace<Autodesk.Revit.DB.FamilyInstance>(Document);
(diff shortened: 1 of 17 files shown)


sharadkjaiswal:
@Randy-Ma you might be interested in it.

kiki-zhang:
@ikeough @Randy-Ma @DynamoDS/michelangelo Please help do the code review, thanks.

ikeough:
@kiki-zhang Really excited to see PRs from the Revit team. I've added one comment that applies to all the places where we construct ACs.

Thanks!

kiki-zhang:
Ian, thanks for your comments.
Regarding the performance concern for huge amounts of ACs creation, we have introduced an API change in Revit2016 to support the batch creation, it should be more efficient than placing AC one by one. If needed, we can consider to extend the batch creation in Dynamo.

kronz:
Hi @kiki-zhang .  There is actually a task in the Youtrack backlog about this, not sure if you have access:
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-3427
I'm planning on talking to Alley about this soon.

kiki-zhang:
@kronz I can access to this link, and add myself as watcher now. ACs batch creation is supported since Revit2016, I believe we cannot make it in DynamoRevit 2015 branch.

kiki-zhang:
I am not familiar with Github and got involving several commits, sorry for that. I updated the fix according to @ikeough comments, trying to throw out more detail information about failure. The detail info we get from Revit as it has already handles intern errors and throw them out, we can show them directly.
With a null value, it throws with message such as "The value is null..". With a non-adaptive family instance, it throws such as "The Symbol famSymb is not an Adaptive Family Symbol..."...

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -88,24 +87,35 @@ private void InitAdaptiveComponent(Point[] pts, FamilySymbol fs)
...
+                    st.Start();
+                    var fam = AdaptiveComponentInstanceUtils.CreateAdaptiveComponentInstance(Element.Document, fs.InternalFamilySymbol);
+                    InternalSetFamilyInstance(fam);
+                    InternalSetPositions(pts.ToXyzs());
+                    st.Commit();
+                }
+                catch (Exception)
+                {
+                    st.RollBack();
+                    throw new ArgumentException("An adaptive component could not be found or created.");
```
Randy-Ma: The messages need to be added to the resource file to support localization.

kiki-zhang:
Thanks @Randy-Ma .
I updated the exception message in subsequent commit to display more detail error info, and now the new message includes two parts, one part is the original text "An adaptive component could not be created.", and the other part extracts from RevitAPI or System Exception.
We may have bilingual message if having the original text translated and leave others. Is it acceptable?

ikeough:
@kiki-zhang Thanks for the updates, and sorry for the delayed reply. 

We should move the first part of the string to a resource file, so we don't have a dual language message. We will cover that in a separate fix.

The fix looks sufficient to cover my original requests. Please merge and move the associated YouTrack issue to "Testing". LGTM.

PR #124, opened 2015-01-26
Updated Greg dll's and improve dependency injection of IAuthProvider

pboyer:
Updates for https://github.com/DynamoDS/Dynamo/pull/3619

@lukechurch PTAL

PR #144, opened 2015-02-04
Revit2016 : Updated Greg dll's and improve dependency injection of IAuthProvider

pboyer:
See #124


jnealb:
@pboyer Do I need to update Greg dll?

PR #145, opened 2015-02-04
(DO NOT MERGE) MAGN-6141 Change the host application unit conversion in Revit to use the Revit unit display settings.

ikeough:
## What
This PR addresses [MAGN-6141](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6141), [MAGN-4049](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-4049), and [MAGN-3960](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-3960) . Previously, numerical inputs used to convey units of measure in Dynamo on Revit were assumed to be in SI units, and converted to their Revit API equivalent when being passed into Revit. As such, users needed to use nodes in the DynamoUnits library to assist in inputting units of measure in their system of choice. here we update the conversion methods currently used to provide a conversion factor based on the current Revit document's display unit settings. This effectively allows users to enter values in Dynamo in the display units specified in Revit.

## How
- DynamoRevit no longer references the DynamoUnits library. The call during initialization to set the static host unit conversion factors on the `BaseUnit` class have been removed. 
- The `DynamoToHostFactor` and `HostToDynamoFactor` properties on `UnitConverter` have been replaced with to the methods `DynamoToHostFactor(UnitType unitType)` and `HostToDynamoFactor(UnitType unitType)`. By passing a `UnitType` as a parameter, these conversion methods can now be used to convert things other than lengths.
- The internals of the `DynamoToHostFactor` method now retrieve the conversion factor from the current Document. This conversion factor is based on the document's `DisplayUnitType` for a given `UnitType`.
- Inlcudes all changes from [this PR](https://github.com/DynamoDS/DynamoRevit/pull/154)  

PTAL:
- [x] @hlp 

revitdisplayunits


src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
```diff
@@ -1,28 +1,31 @@
...
     {
-        public static double DynamoToHostFactor
+        public static double DynamoToHostFactor(UnitType unitType)
         {
-            get { return Length.FromDouble(1.0).ConvertToHostUnits(); }
+            var revitDisplayUnits =
+                DocumentManager.Instance.CurrentDBDocument.GetUnits()
+                    .GetFormatOptions(unitType)
+                    .DisplayUnits;
+            return UnitUtils.ConvertToInternalUnits(1, revitDisplayUnits);
```
hlp: Personally, I'm not a fan of "magic numbers" (http://en.wikipedia.org/wiki/Magic_number_%28programming%29#Unnamed_numerical_constants) in code. Could the 1 be broken out into a variable describing what it is / does?

src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
```diff
@@ -1,28 +1,31 @@
...
         }
 
-        public static double HostToDynamoFactor
+        public static double HostToDynamoFactor(UnitType unitType)
         {
-            get { return 1 / DynamoToHostFactor; }
+            return 1/DynamoToHostFactor(unitType);
```
hlp: Ditto above: can we remove this magic number?

hlp:
Besides my two points above, LGTM

src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
ikeough: That "magic" number is actually the number I want. In order to use this API, we need to provide a value to be converted. That value, in this case, is 1.0.

src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
ikeough: Again, not a "magic" number. We are inverting the host factor.

src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
hlp: Maybe add a comment? I just seems odd to have a number, 1, being fed into a function, if it's never going to change.

ikeough:
@hlp. Added some comments to address your concerns above. I also marked this as DO NOT MERGE, until we can get your changes for [MAGN-4049](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-4049) in place and we can send documentation out to users.

PR #150, opened 2015-02-07
MAGN-6032 [CRASH] while placing Select Model Element node in Canvas while running Dynamo on top of Revit-2016 only.

ikeough:
This PR addresses [MAGN-6032](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6032).

## What
When placing a Select Model Element(s) node in Revit, a crash would occur.

## How
- Wraps the event hooks in the RevitSelection constructor in an Idle action in order to avoid the Revit API warning that calls are being made outside the Revit context.
- Fixes an issue when selecting reference planes where the InternalUniqueId was not set.

PTAL:
- [x] @Randy-Ma 

Merge:
- [ ] Revit2015

### DIFF
--- a/src/Libraries/RevitNodes/Elements/ReferencePlane.cs
+++ b/src/Libraries/RevitNodes/Elements/ReferencePlane.cs
@@ -73,7 +73,7 @@ namespace Revit.Elements
         /// <param name="referencePlane"></param>
         private void InitReferencePlane(Autodesk.Revit.DB.ReferencePlane referencePlane)
         {
-            InternalReferencePlane = referencePlane;
+            InternalSetReferencePlane(referencePlane);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/Elements/ReferencePlane.cs
```diff
@@ -73,7 +73,7 @@ private ReferencePlane(XYZ bubbleEnd, XYZ freeEnd, XYZ normal, View view )
         /// <param name="referencePlane"></param>
         private void InitReferencePlane(Autodesk.Revit.DB.ReferencePlane referencePlane)
         {
-            InternalReferencePlane = referencePlane;
+            InternalSetReferencePlane(referencePlane);
```
Randy-Ma: Just curious whether this is changed because of a defect? I am asking because I remember once it is like what you have changed, but then someone changed it to be like what is before your change. But I have just looked into the code and it should be ok anyway.

Randy-Ma:
LGTM

ikeough:
@Randy-Ma This fixes MAGN-6032, and I do remember making this change before as well. Perhaps we had already branched 2016 at this point and this change never got picked over.

PR #151, opened 2015-02-09
MAGN-4679: Wrap structural foundation as FamilyInstance in Dynamo.

LianweiHuang:
MAGN-4679:  Previously structural foundation was wrapped as structural
framing in Dynamo.  And framing in Dynamo expects a Curve instead of
Point when getting its location. But location of foundation does return
a Point since its point-based.


LianweiHuang:
@ikeough @Randy-Ma Could you please take a look at the review?

LianweiHuang:
Close this one and will file another review.

PR #152, opened 2015-02-09
MAGN-4679 : Wrap structural foundation instance as FamilyInstance in Dyn...

LianweiHuang:
...amo.

Previously, foundation instance was wrapped as structural framing
instance in Dynamo which is assumed to be curve-based, but foundation is
Point-based. So StructuralFraming::Location does not for foundation.

@ikeough @Randy-Ma , please take a look at this review.


LianweiHuang:
Any comments about the fix?

Some discussion is here 
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-4679

Randy-Ma:
Hi @LianweiHuang, this looks great! 

one minor point is that the title is not complete. I guess this is because some part of the title is placed in the summary.

One more point, can you help check whether you can create a foundation using the StructuralFraming node? If that is the case, we do not want inconsistency: a foundation will be wrapped as a family instance but you can create a foundation as a structural framing.

LianweiHuang:
@Randy-Ma , we can't create a foundation with a framing node, as you can see in that node only curved-based family instance(beam/brace/slantedcolumn) could be created.

Randy-Ma:
Thanks for checking this! LGTM!

LianweiHuang:
Thanks,  @Randy-Ma !

PR #153, opened 2015-02-09
Build fix for compiling error in RevitDropDown.cs

LianweiHuang:
It seem method AddIdleAction in DynanoRevit was somehow not committed.

Currently, I just revert the change RevitDropDownBase() to the one before AddIdleAction was committed.

@ikeough @Randy-Ma please help review.


LianweiHuang:
FYI @DynamoDS/michelangelo

ikeough:
@LianweiHuang Sorry. This was my fault. I cherry-picked that fix into the Revit2015 branch, without realizing that the AddIdleAction method wasn't there. Thanks for the fix!

PR #155, opened 2015-02-10
MAGN-4858 Adaptive components placed at origin when points are null

kiki-zhang:
The change includes a fix of MAGN-4858.
Root Cause: For creating a adaptive component, multiple RevitAPI are
called in one transaction. The first step is to create a original
component, and then call another RevitAPI to set its position. The issue
happens because an exception throws out in the process of setting
position due to the null point, which finally leads to the uncompleted
creation.
Solution: Using a subtransaction to roll back the creation if an
exception happens.


kiki-zhang:
@ikeough I am confused by merging fix in the previous branch so i create a new one, please help re-view. Sorry for the redundant commit.

ikeough:
@kiki-zhang Seems that this can't be merged automatically. Once you resolve the merge conflict, this looks good to me. And thank you for moving those messages to the resource files.

Randy-Ma:
It also looks good to me now!

kiki-zhang:
@ikeough This PR can be merged automatically now, please feel free to let me know if anything else I need to follow up.

PR #162, opened 2015-02-17
Added Test Case for All Selection Nodes.

riteshchandawar:
This is very long test but if it fails it will not take long time to debug. I have given clear names to each verification point so that we will be easily point out what is the cause of Failure.

Task related to this work [MAGN-4362](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-4362)

<h3>Reviewer</h3>
- [x] @Randy-Ma 

Will issue separate PR for **Revit2015** and **Revit2016**

Test Case is passing on latest build.
image

### DIFF
--- a/test/Libraries/RevitIntegrationTests/SelectionTests.cs
+++ b/test/Libraries/RevitIntegrationTests/SelectionTests.cs
@@ -2,24 +2,17 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-
 using Autodesk.DesignScript.Geometry;
 using Autodesk.Revit.DB;
-
 using DSRevitNodesUI;
-
 using Dynamo.Interfaces;
 using Dynamo.Models;
 using Dynamo.Nodes;
 using Dynamo.Tests;
-
 using NUnit.Framework;
-
 using RevitServices.Persistence;
 using RevitServices.Transactions;
-
 using RTF.Framework;
-
 using ReferencePoint = Revit.Elements.ReferencePoint;
 
 namespace RevitSystemTests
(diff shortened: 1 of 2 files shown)


test/Libraries/RevitIntegrationTests/SelectionTests.cs
```diff
@@ -362,6 +355,126 @@ public void SelectionInSyncWithDocumentOperations_References()
...
+            var selectFacesNodeID = "00a2189b-88f4-4e15-b9f0-efefa6357b37";
+            AssertPreviewCount(selectFacesNodeID, 5);
+            for (int i = 0; i <= 4; i++)
+            {
+                var surface = GetPreviewValueAtIndex(selectFacesNodeID, i) as Surface;
+                Assert.IsNotNull(surface);
+            }
+
+            // StructuralFramingTypes
+            var structuralFramingTypesNodeID = GetPreviewValue("b786fdf5-b6fa-4aab-a2fd-b40ac12b8b87");
```
Randy-Ma: Is the value of this node an ID? If not, why not change the variable name to be a name easier to understand.

Randy-Ma: Similarly, there are few names which can be changed after this line.

riteshchandawar: @Randy-Ma , those are Node GUID so I have appended node name to understand ID coming fomr which node

Randy-Ma:
@riteshchandawar Thanks for the fix!
It will be easier for others to understand the code if the name of some variables are changed. But other than this, LGTM!

PR #175, opened 2015-02-23
Updated batch file to include VS2013 location

ramramps:
**Issue**: 
Running the batch file - transform_all.bat does not create RevitTestConfiguration.xml file in VS2013.

**Cause**:
TextTransform.exe which creates the xml file from the template is in different location for VS2013. For VS2012 this exe is in C:\Program Files (x86)\Common Files\microsoft shared\TextTemplating\11.0 and for VS2013 this file is in C:\Program Files (x86)\Common Files\microsoft shared\TextTemplating\12.0 folder.

**Fix**:
Updated the batch file to look for TextTransform.exe in both folders.

- [ ] @ikeough

### DIFF
--- a/src/transform_all.bat
+++ b/src/transform_all.bat
@@ -24,6 +24,8 @@ set file_name=!file_name:~0,-3!.%extension%
 echo:  \--^> !file_name!    
 echo "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\11.0\TextTransform.exe" -out !file_name! %%d
 "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\11.0\TextTransform.exe" -out !file_name! %%d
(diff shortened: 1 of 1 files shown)


src/transform_all.bat
```diff
@@ -24,6 +24,8 @@ set file_name=!file_name:~0,-3!.%extension%
 echo:  \--^> !file_name!    
 echo "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\11.0\TextTransform.exe" -out !file_name! %%d
 "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\11.0\TextTransform.exe" -out !file_name! %%d
+echo "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\12.0\TextTransform.exe" -out !file_name! %%d
```
sharadkjaiswal: Can we specify version 12.0 as a parameter, and pass the correct parameter based on Visual studio version from the cs project.

ramramps:
@ikeough - Do we need a parameter ?

PR #176, opened 2015-02-24
Refactor Revit test base classes.

ikeough:
This PR refactors the Revit base classes as follows:
- Merges the SystemTest class into RevitSystemTestBase.
- Changes all former classes which inherited from SystemTest to inherit from RevitSystemTestBase.
- Adds logic to set the DocumentManager's UIApplication and UIDocument. This logic was previously set inside RTF, but all Dynamo dependencies have been removed from RTF.

PTAL:
- [ ] @Randy-Ma 

Requires Merge:
- [ ] Revit_2015
- [ ] Revit_2016


test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -59,7 +162,46 @@ public override void TearDown()
             TransactionManager.Instance.ForceCloseTransaction();
         }
 
-        public override void StartDynamo()
+        #endregion
+
+        #region protected methods
+
+        protected override void SetupCore()
```
Randy-Ma: Shall we keep the access modifiers of SetupCore?

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -59,7 +162,46 @@ public override void TearDown()
...
+                emptyModelPath1 = Path.Combine(workingDirectory, "emptyV1.rfa");
+            }
+            else
+            {
+                emptyModelPath = Path.Combine(workingDirectory, "empty.rfa");
+                emptyModelPath1 = Path.Combine(workingDirectory, "empty1.rfa");
+            }
+        }
+
+        protected override void StartDynamo()
```
Randy-Ma: Shall we keep the access modifiers of StartDynamo?

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
Randy-Ma: Sorry, I see you have changed the modifier of the method in the base class.

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
Randy-Ma: Ignore this comment

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -36,17 +142,14 @@ public void Shutdown()
...
+        private string defsPath;
+        protected string emptyModelPath1;
+        protected string emptyModelPath;
 
-        [SetUp]
-        public override void Setup()
-        {
-            base.Setup();
-
-            DocumentManager.Instance.CurrentUIApplication.ViewActivating += CurrentUIApplication_ViewActivating;
```
Randy-Ma: @ikeough 
Is this required to fix something? I have an impression ViewActivating event is used for multiple test cases from a test source.

Randy-Ma: I mean those test cases in the "RegressionTest" category in RevitSystemTests.

ikeough: I didn't remove this hook completely. I just moved it into SetupCore. See below. Once I moved the ViewActivating hook, the Setup override was no longer needed.

Randy-Ma:
@ikeough Thanks for the fix! Except for the one point about the ViewActivating event, LGTM! I am merging this so that at least most test cases pass (if I am wrong with the event, then hopefully all impacted test cases will pass again). As the access modifiers of two methods have changed, this need to be merged so that there will be no issues with the build.

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
ikeough: @Randy-Ma I see what you mean now :( It needs to call the Setup method before each regression test. Nice catch!

PR #182, opened 2015-02-27
MAGN-6313 Create conversion nodes for new unit system

ramramps:
New node "Convert units" was added to Revit, as part of Units project. 

This node has three dropdowns: 
1. The top dropdown refers to metric units - Length,Area,Volume. 
2. The left dropdown refers to Revit Display unit settings.
3. The right dropdown refers to SI units

Selecting the metric unit dropdown will select default values in other dropdowns.
Left dropdown (default unit settings from Revit) is always disabled. 

- [ ] @ikeough


ikeough:
Please rename to "Convert Units". "Translate" has another meaning in geometric terms.

ikeough:
@ramramps You need to make this pull request against the units branch.

PR #185, opened 2015-03-01
"DynamoPathManager" related update for Revit 2016

Benglin:
These changes in `Revit` are required for `DynamoPathManager` changes made in [another pull request](https://github.com/DynamoDS/Dynamo/pull/3824).

Hi @ikeough, I believe all the necessary parts in Revit add-on side have been updated. The logic around test for `Revit 2016` is a little unique (that it has to preload 221 ASM), I think the build around that area fails as of right now. I'll get that fixed tomorrow morning when I get into the office. This PR should have fixed the main build errors.

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -85,6 +85,7 @@ namespace Dynamo.Applications
     public class DynamoRevit : IExternalCommand
     {
         private static readonly List<Action> idleActions = new List<Action>();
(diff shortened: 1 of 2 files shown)


ikeough:
@Benglin This looks fine. Yes, 221 is required for 2016, and the tests right now have to overload the entire setup method just to set that which is unfortunate. If we can make it so that the RevitSystemTests have a common setup method that loads the right version of LibG, it would be much preferred.

PR #187, opened 2015-03-02
Fix the failed test cases

Randy-Ma:
As there are changes to set the path, we need to update our test framework to set the path correctly as well.
The test cases are failing because the geometry factory path is not set at all in the test framework.

@Benglin


Benglin:
Wow, that's awesome! Thanks for helping with the fix, @Randy-Ma! LGTM!

PR #189, opened 2015-03-02
Fix the build error and the failing test cases

Randy-Ma:
This is to merge the following pull request to fix the failing test cases:
https://github.com/DynamoDS/DynamoRevit/pull/187

It also makes some changes to remove the line that is causing the compiling errors because it is not required to preload. I have verified one geometry test cases in SampleTests.

@Benglin

### DIFF
--- a/test/Libraries/RevitTestServices/RevitSystemTestBase.cs
+++ b/test/Libraries/RevitTestServices/RevitSystemTestBase.cs
@@ -170,8 +170,6 @@ namespace RevitTestServices
                 workingDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
             }
 
-            DynamoPathManager.PreloadAsmVersion("221", DynamoPathManager.Instance);
-
             CreateTemporaryFolder();
 
             // Setup Temp PreferenceSetting Location for testing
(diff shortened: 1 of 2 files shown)


test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -170,8 +170,6 @@ public override void Setup()
                 workingDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
             }
 
-            DynamoPathManager.PreloadAsmVersion("221", DynamoPathManager.Instance);
```
Randy-Ma: @ikeough 
I think this is not required and have verified this with one test case. But just in case anything I have missed, please help double check this.

Benglin:
Thanks @Randy-Ma, this is of great help! LGTM!

Anyway, @ikeough, I'm merging this in to have the test run through the changes. Let's see if this breaks anything, you can always comment on it later. Thanks in advance.

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
ikeough: @Randy-Ma that line was the only thing that was different about the Setup method in RevitSystemTestBase. I believe we can remove the Setup method and let it use the one on the base class.

PR #202, opened 2015-03-05
add SimpleRaaS.xml in Revit

zora-wang:
@sharadkjaiswal 
SimpleRaas.xml is copied to output path in  src/DynamoRevit/DynamoRevit.csproj

How this .xml comes from:
1.run DynamoRaas project-> SimpleRaas.xml is generated automatically in DynamoRevit project;
2.this .xml contains the node description info, and these nodes are in Analyze/DayLighting, Render ,etc.


sharadkjaiswal:
Since there are no code changes to DynamoRaaS project, let's not submit binaries. Create a new pull request which add SimpleRaaS.xml and update DynamoRevit.csproj to copy this XML to bin folder.

PR #207, opened 2015-03-06
MAGN-6433 [CRASH] while placing *SiteLocation* node in canvas, if active document is Family.

ikeough:
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6433

The Site Location is not available in a Family document. When a user attempted to place the node, the Site Location could not be found, a null was returned, and we were not gracefully handling this condition. This PR add a check for whether the document is a Family document. If it is, a warning is set on the node and the Location property is set to null. A null value is returned from the node.

The new look:

image

Requires Merge:
- [ ] Revit2014
- [ ] Revit2015

PTAL @Randy-Ma


ikeough:
This branch should be called MAGN-6433.

ikeough:
@Randy-Ma ping!

Randy-Ma:
@ikeough 
Thanks for the fix! LGTM!

PR #208, opened 2015-03-06
Fix for removal of OnNodesModified

pboyer:
See https://github.com/DynamoDS/Dynamo/pull/3909

PR #209, opened 2015-03-06
Fix for removal of OnNodesModified

pboyer:
See https://github.com/DynamoDS/DynamoRevit/pull/208


ikeough:
LGTM pending one small issue on the core implementation.

PR #211, opened 2015-03-06
MAGN-6582 Switching Workspaces in Run Auto all connectors become invisible, locks workspace

ikeough:
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6582

Previously we were only wrapping the call to ResetEngine in a scheduler task. This meant that when a workspace was opened, the task went off for processing and the UI tried to update immediately. Wrapping the entire file opening process in a scheduler task means that resetting of the engine controller and population of the UI are properly ordered and called in Revit's "API Context".

PTAL:
@pboyer 

Requires Merge:
- [ ] Revit2014


pboyer:
LGTM!

PR #212, opened 2015-03-09
Removed "RevitTestConfiguration.DefinitionsPath" property

Benglin:
Hi @ikeough, please let me know what you think. `DefinitionsPath` makes use of `DynamoPathManager`, and since I'm removing `DynamoPathManager`, it needs to be reimplemented. Good thing though, it is not used anywhere, so I choose the easier way out -- to remove it entirely. Does this look okay to you?


ikeough:
It it's not used, by all means, get rid of it! LGTM.

Benglin:
Thanks @ikeough!

PR #215, opened 2015-03-10
byNameNode Description updated

zora-wang:
@sharadkjaiswal 

"byName" node description is updated in Revit


src/Libraries/RevitNodes/Elements/Category.cs
```diff
@@ -57,7 +57,7 @@ internal Autodesk.Revit.DB.Category InternalCategory
         #region public static constructors
 
         /// <summary>
-        /// Get a Revit category by by the built in category name.
+        /// Get a Revit category by the built in category name.
```
sharadkjaiswal: Gets a Revit category by the built-in category name.

PR #221, opened 2015-03-10
Override the configuration file to fix the failing test cases

Randy-Ma:
Right now the configuration file - "TestServices.dll.config" is used but the RequestedLibraryVersion property is not set. As a result, libg_219 will always be used. In a computer where Revit 2014 is not installed, it will not work.

This fix overrides this configuration file so that the libg library of a correct version is used.

@ikeough 
PTAL

### DIFF
--- a/test/Libraries/RevitTestServices/RevitTestServices.csproj
+++ b/test/Libraries/RevitTestServices/RevitTestServices.csproj
@@ -110,6 +110,9 @@
     </Reference>
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <PropertyGroup>
+    <PostBuildEvent>xcopy /Y /R $(ProjectDir)TestServices.dll.config $(TestOutputPath)\..\</PostBuildEvent>
+  </PropertyGroup>
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
   <Target Name="BeforeBuild">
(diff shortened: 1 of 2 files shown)


Randy-Ma:
@sharadkjaiswal 
Could you help do a quick review so that I can merge this so that the fix can be picked?

sharadkjaiswal:
Are these tests going to run in the context of Revit or standalone mode? If it's running in the context of Revit then we should get this information based on Revit version in use.

ikeough:
@sharadkjaiswal They might not be run in the context of Revit. This config file is used by all the testing base classes for tests including those not related to Revit. 

@Randy-Ma Thanks for this Randy. I hadn't considered the case where all versions of Revit are not available. The only problem now is that this config file is basically required. The logic for reading the config file doesn't assume this. It assumes that if the config file is missing it'll fall back to using the executing assembly directory and 219 (not a great assumption we now know). I think we should either throw an exception if the config file is missing or use the old method we had to "FindASM" if the version isn't specified. Thanks very much for working on this. We keep fixing one configuration problem only to reveal another. If you want me to pick up this issue in the morning Id be happy to.

ikeough:
@Randy-Ma Looks like @Benglin has removed the FindASM method (probably for good reason). I like this solution because it allows us to forego having inherited classes that override setup methods, which we've found is very difficult to debug. I'm going to merge these in, and add one to Revit2014 just for consistency.

Benglin:
`FindASM` fails just as bad when there's no Revit installed, and it's removed now for a good reason (Dynamo can't care less about finding ASM) ;)

PR #222, opened 2015-03-10
Override the configuration file to fix the failing test cases

Randy-Ma:
This is to merge the following pull request to fix the failing test cases:
https://github.com/DynamoDS/DynamoRevit/pull/221

The only difference is that the RequestedLibraryVersion property is changed accordingly.

@ikeough 
PTAL


Randy-Ma:
@sharadkjaiswal 
Could you help do a quick review so that I can merge this so that the fix can be picked?

PR #225, opened 2015-03-11
Set up the paths when launching the application

Randy-Ma:
There are two changes here:
1).  An specific assembly resolver is created which will not depend on the path manager. This is in line with the ongoing changes with the path manager. The changes here is to make sure RTF will run correctly with a correct resolver.
2). SetupDynamoPaths() is moved from DynamoRevitApp to DynamoRevit so that the second time the plugin is started, the path manager will be created correctly. This is also in line with the ongoing changes with the path manager.

@ikeough 
@Benglin 
PTAL

### DIFF
--- a/src/DynamoRevit/DynamoRevitApp.cs
+++ b/src/DynamoRevit/DynamoRevitApp.cs
@@ -42,8 +42,6 @@ namespace Dynamo.Applications
         {
             try
             {
-                SetupDynamoPaths(application);
-
                 SubscribeAssemblyResolvingEvent();
 
                 ControlledApplication = application.ControlledApplication;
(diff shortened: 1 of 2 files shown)


ikeough:
LGTM. Thanks @Randy-Ma

Randy-Ma:
Ian, I will merge this first and later change the string.

src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -120,38 +118,47 @@ private static void RegisterAdditionalUpdaters(UIControlledApplication applicati
...
+                if (File.Exists(assemblyPath))
+                {
+                    return Assembly.LoadFrom(assemblyPath);
+                }
+                return null;
+            }
+            catch (Exception ex)
+            {
+                throw new Exception(string.Format("There location of the assembly, {0} could not be resolved for {1loading.", assemblyPath), ex);
+            }
```
Benglin: There's a typo around `{1loading`, fixed that in my branch ;)

PR #228, opened 2015-03-11
Update the exception message

Randy-Ma:
As @ikeough mentioned, there are some errors in the exception message. This submission to to update the message so that it is correct.

@Benglin 
PTAL


Benglin:
LGTM! Thanks for fixing

PR #232, opened 2015-03-11
Adding one used namespace

Randy-Ma:
This is to add one missing namespace used by the path manager.

@Benglin 
PTAL


Benglin:
LGTM!

PR #234, opened 2015-03-11
Set up paths for the test cases

Randy-Ma:
For the test cases, we also need to set up the paths before running them. The paths won't be set because right now we have moved the setting logic to DynamoRevit.

This is pretty much similar to what is done by Ben right now.

@Benglin 
PTAL


Benglin:
Thanks for the fix, @Randy-Ma. LGTM!

PR #241, opened 2015-03-12
[DO NOT MERGE] Fix MAGN-6634

Randy-Ma:
@Benglin 
Could you help review this and merge and cross-merge this?

The cause is similar to:
https://github.com/DynamoDS/Dynamo/pull/2112

This submission fixes the crash when Dynamo tries to read trace data using SoapFormatter. The formatter will require the type information related to the object to be desialized and it will try to load the assembly. But the resolver does not contain the path information. As a result, the assembly will not be found and the formatter will throw an exception. This submission fixes the issue by adding another path to the resolver.

### DIFF
--- a/test/Libraries/RevitNodesTests/GeometryConversion/RevitToProtoSolidTests.cs
+++ b/test/Libraries/RevitNodesTests/GeometryConversion/RevitToProtoSolidTests.cs
@@ -38,7 +38,7 @@ namespace RevitNodesTests.GeometryConversion
             }
         }
 
-        [Test]
+        [Test, Category("Failure")]
         [TestModel(@".\Solids.rfa")]
         public void ToProtoType_Boolean_ShouldConvertAllFormsInDocument()
         {


RodRecker:
Thank you Randy. Please be sure to include the defect title in addition to number.


Rod

src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -153,6 +153,11 @@ public static Assembly ResolveAssembly(object sender, ResolveEventArgs args)
                 {
                     return Assembly.LoadFrom(assemblyPath);
                 }
+                assemblyPath = Path.Combine(parentDirectory.FullName, @"Revit_2015\" + new AssemblyName(args.Name).Name + ".dll");
```
sharadkjaiswal: Shouldn't we use assemblyDirectory instead of hardcoding the folder name Revit_2015??

Randy-Ma:
@Benglin 
I will temporarily mark this one as "DNM". But if your changes related to this come in first, be careful about this.

Randy-Ma:
As Ben has addressed the issue in: 
https://github.com/DynamoDS/DynamoRevit/pull/245
https://github.com/DynamoDS/DynamoRevit/pull/246

I am closing this one.

Benglin:
Thanks @Randy-Ma!

PR #242, opened 2015-03-12
(WIP) Revit fixes 031215

ikeough:
This PR fixes a number of Revit system test failures which were due to the merge of the units branch.

As part of this PR, I've also moved application event handling from `DynamoRevit` to `RevitDynamoModel`. This assures us that the application event handlers align with the life-cycle of the `RevitDynamoModel` object, and it makes it much clearer when application event handlers are being bound and when they are being unbound.

PTAL:
@pboyer 

Requires Merge:
- [ ] Revit2014
- [ ] Revit2016


src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -30,6 +34,25 @@
...
+        public IPreferences Preferences { get; set; }
+        public bool StartInTestMode { get; set; }
+        public IUpdateManager UpdateManager { get; set; }
+        public ISchedulerThread SchedulerThread { get; set; }
+        public string GeometryFactoryPath { get; set; }
+        public IAuthProvider AuthProvider { get; set; }
+        public string PackageManagerAddress { get; set; }
+        public ExternalCommandData ExternalCommandData { get; set; }
+    }
+
```
pboyer: How about we move this to inside of the class like this:

    public class Foo
    {
        public interface IFooStartConfig
        {
            int Bar { get; set; }
        }
    }

pboyer: Also, if you're going to refactor this, we should probably shorted `xStartConfiguration` to `xStartConfig`.

ikeough:
@pboyer I updated the interfaces both here and in [this PR](https://github.com/DynamoDS/Dynamo/pull/3970) per our discussion. PTAL.

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -195,6 +223,92 @@ private void UnsubscribeDocumentManagerEvents()
...
+            commandData.Application.Application.DocumentClosing += OnApplicationDocumentClosing;
+            commandData.Application.Application.DocumentClosed += OnApplicationDocumentClosed;
+            commandData.Application.Application.DocumentOpened += OnApplicationDocumentOpened;
+
+            hasRegisteredApplicationEvents = true;
+        }
+
+        private void UnsubscribeApplicationEvents(ExternalCommandData commandData)
+        {
```
Benglin: Missed out a check:

```csharp
if (!hasRegisteredApplicationEvents)
{
    return;
}
```

PR #243, opened 2015-03-12
MAGN-5854 : Crash on certain Revit geoemtry

hlp:
This PR fixes the C# aspect of MAGN-5854. Ruled Surfaces were not correctly extracted out of Revit RuledFace objects: the ruled Surface must completely contain the trimming edge loops. The fix extends the ruled Surfaces in the instances when the ruled surface is too small.

@Randy-Ma PTAL


src/Libraries/RevitNodes/GeometryConversion/SurfaceExtractor.cs
```diff
@@ -238,15 +238,84 @@ public static Surface ExtractSurface(Autodesk.Revit.DB.RevolvedFace face, IEnume
...
+            bool success = EdgeLoopsTouchSurface(result, edgeLoops);
+
             c0.Dispose();
             c1.Dispose();
-            return result;
+            c0Move.Dispose();
+            c1Move.Dispose();
+
+            if (success)
+                return result;
```
Randy-Ma: @hlp 
But does this mean the returned surface does not match the original face? because it is extended.

hlp: It will match the original face, because later it will be trimmed by the edge loops. If you look at ExtractSurface with a PlanarFace argument, the returned face is a lot larger than the input face. Like this method, it is assumed it will be trimmed downstream.

Randy-Ma:
LGTM

Randy-Ma:
@hlp 
Also could you help create pull requests for other branches as well?

PR #244, opened 2015-03-13
Path Manager: Revit specific changes

Benglin:
Background
-----
`DynamoPathManager` has been [completely removed in `DynamoDS/Dynamo:master` branch](https://github.com/DynamoDS/Dynamo/pull/3934), this pull request updates Revit side of things to make use of the new path management system.

Changes
-----
1. References to `DynamoPathManager` are now removed.

2. `class RevitPathResolver : IPathResolver` interface implementation is introduced. Its job is to add `Revit_20xx` into additional resolution path collection, and to include both `RevitNodes.dll` and `SimpleRaaS.dll` for preloading.

3. `DynamoRevit.InitializeCoreModel` now supplies a `RevitPathResolver` object to `RevitDynamoModel`. Its job is to provide additional information about resolution path and preloaded node assemblies.

4. We no longer create a `PreferenceSettings` object and hand it off to `RevitDynamoModel` at start up. The `Preferences` reference is left as `null` so that an instance is created within `DynamoModel` constructor.

5. `class RevitTestPathResolver : IPathResolver` class is introduced. It is equivalent to `RevitPathResolver`, except that it is meant for Revit test cases. In addition to that, it also allows assembly directory (the one that contains `RevitNodes.dll` and `SimpleRaaS.dll`) to be specified.

Notes to Reviewer
-----
Hi @ikeough, please have a look. Note that `AssemblyResolver` is no longer a `static` class. Based on your recent experience with Revit test failures, could you please suggest any particular tests to run in Revit Test Framework? Thanks!

Hi @Randy-Ma, this is for your information, thanks again for the help along the way!


src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -97,14 +98,6 @@ public new static RevitDynamoModel Start(IRevitStartConfiguration configuration)
...
             if (string.IsNullOrEmpty(configuration.Context))
                 configuration.Context = Core.Context.REVIT_2014;
-            if (string.IsNullOrEmpty(configuration.DynamoCorePath))
-            {
-                var asmLocation = Assembly.GetExecutingAssembly().Location;
-                configuration.DynamoCorePath = Path.GetDirectoryName(asmLocation);
-            }
-
-            if (configuration.Preferences == null)
-                configuration.Preferences = new PreferenceSettings();
```
Benglin: Doing `new PreferenceSettings()` here is as good as doing it in `DynamoModel` base class. Unless `RevitDynamoModel` wants to supply a different implementation, it does not have to do this here (it's making something that is defined in `DynamoCore`, and then pass it back into `DynamoCore`, which is weird). Passing `null` here results in the same `new PreferenceSettings()` call in `DynamoModel`.

PR #246, opened 2015-03-14
Fixed issue with "SoapFormatter.Deserialize" exception

Benglin:
This fixes [an internally tracked defect](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6634).

When `SoapFormatter.Deserialize` tries to load `RevitServices.dll`. The assembly resolution logic in `DynamoRevitApp.ResolveAssembly` was called, but it attempts to load the said assembly from:

    C:\Program Files\Dynamo 0.8

That itself is insufficient, the `RevitServices.dll` resides in the same folder where `DynamoRevitDS.dll` resides, which is the following folder:

    C:\Program Files\Dynamo 0.8\Revit_2015

This pull request enhances the logic by first checking to see if `RevitServices.dll` can be found alongside `DynamoRevitDS.dll`.

@Randy-Ma, I'm going to merge this in to stabilize the build, but please have a look afterwards, thanks :)


Randy-Ma:
LGTM

PR #248, opened 2015-03-15
Version Selector support for all Dynamo installs

sharadkjaiswal:
- To support side-by-side install of Dynamo,  we need to ensure that
version selector can pick all the installed versions.
- Used Split button to show available Dynamo versions as drop down.
- UI strings and messages are extracted to resource files so that it can
be localized.
- Dynamo version on Ribbon is shown by looking at the version info of
DynamoRevit.dll
- ProductLookUp logic is implemented in a brand new DynamoInstallDetective.dll to find all installed Dynamo on the system.
- We can't add a reference to DynamoInstallDetective.dll because it is built in Dynamo core folder and Revit while loading version selector will not be able to resolve this dependency.
- Hence use reflection to load DynamoInstallDetective.dll and then call a utility method to get all installed Dynamo.
- Fixes following issues:
- http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6553
- http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6608

**Reviewers:**
- [ ] @ikeough 
- [x] @aparajit-pratap


src/AssemblySharedInfoGenerator/AssemblySharedInfo.tt
```diff
@@ -9,7 +9,7 @@ using System.Runtime.InteropServices;
 // associated with an assembly.
 [assembly: AssemblyCompany("Autodesk, Inc")]
 [assembly: AssemblyProduct("Dynamo")]
-[assembly: AssemblyCopyright("Copyright © Autodesk, Inc 2014")]
+[assembly: AssemblyCopyright("Copyright © Autodesk, Inc 2015")]
 [assembly: AssemblyTrademark("")]
```
aparajit-pratap: Since this is a merge to the Revit2014 branch I think this should remain as 2014

sharadkjaiswal: No, it's version Copyright on Dynamo dlls, it doesn't matter if it's for Revit2014 or Revit2015. It's being built in year 2015, so the copyright should represent current year always.

src/Legacy/DynamoRevitVersionSelector/ProductLookUp.cs
```diff
@@ -0,0 +1,345 @@
...
+        void LookUpAndInitProducts(IProductLookUp lookUp);
+    }
+
+    /// <summary>
+    /// Implements basic look up algorithm to get product installations using registry keys
+    /// </summary>
+    public class InstalledProductLookUp : IProductLookUp
+    {
+        const string REG_KEY64 = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\";
+        const string REG_KEY32 = @"SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\";
```
aparajit-pratap: Where do you get these from?

src/Legacy/DynamoRevitVersionSelector/ProductLookUp.cs
```diff
@@ -0,0 +1,345 @@
...
+        /// <summary>
+        /// Implements a product look up algorithm based on registry key.
+        /// </summary>
+        /// <param name="lookUpName">Product name to lookup</param>
+        /// <param name="fileLookup">file name pattern to lookup</param>
+        public InstalledProductLookUp(string lookUpName, string fileLookup)
+        {
+            this.LookUpProductName = lookUpName;
+            this.fileLookup = fileLookup;
+        }
```
aparajit-pratap: what is the difference between these properties?

src/Legacy/DynamoRevitVersionSelector/ProductLookUp.cs
```diff
@@ -0,0 +1,345 @@
...
+    /// </summary>
+    public class InstalledProductLookUp : IProductLookUp
+    {
+        const string REG_KEY64 = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\";
+        const string REG_KEY32 = @"SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\";
+
+        /// <summary>
+        /// Product name for lookup
+        /// </summary>
+        public string LookUpProductName { get; private set; }
```
aparajit-pratap: This name sounds like an action (method name). This could be renamed to `ProductLookupName`.

src/Legacy/DynamoRevitVersionSelector/ProductLookUp.cs
```diff
@@ -0,0 +1,345 @@
...
+                lookUp.GetProductNameList()
+                    .Select(lookUp.GetProductFromProductName)
+                    .Where(p => p != null).OrderBy(p => p);
+        }
+    }
+
+
+    public class DynamoProducts : InstalledProducts
+    {
+        const string PRODUCT_ID07_X = @"{6B5FA6CA-9D69-46CF-B517-1F90C64F7C0B}";
```
aparajit-pratap: Where does this come from?

src/Legacy/DynamoRevitVersionSelector/ProductLookUp.cs
sharadkjaiswal: Windows installation maked an entry to these registry key for 64bit or 32bit installation. Innosetup installer makes use of 32bit installation whereas msi makes use of 64bit installation.

src/Legacy/DynamoRevitVersionSelector/ProductLookUp.cs
sharadkjaiswal: For Dynamo lookup name is Dynamo and the file lookup name is DynamoCore.dll to validate the Dynamo installation.

src/Legacy/DynamoRevitVersionSelector/ProductLookUp.cs
sharadkjaiswal: This clsid was used as product code for 07 installer.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -1,88 +1,87 @@
...
+
 namespace Dynamo.Applications
 {
     [Transaction(TransactionMode.Automatic)]
     [Regeneration(RegenerationOption.Manual)]
     public class VersionLoader : IExternalApplication
     {
-        internal static string BasePath = @"C:\Autodesk\Dynamo\Core";
-        internal static string BetaPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+        public static List<IInstalledProduct> Products { get; private set; }
```
aparajit-pratap: This can be declared `IEnumerable`?

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -102,15 +140,25 @@ public Result OnShutdown(UIControlledApplication application)
...
 
-            Utils.WriteToFile(loadPath, commandData.Application.Application.VersionName);
+            return Result.Succeeded;
+        }
+    }
 
-            Utils.ShowRestartMessage(FileVersionInfo.GetVersionInfo(loadPath).FileVersion);
+    [Transaction(TransactionMode.Automatic)]
+    [Regeneration(RegenerationOption.Manual)]
+    public class Command1 : IExternalCommand
```
aparajit-pratap: Is not there a better way to do this? It seems like you are creating `Command`s for each product. Can we not dynamically create as many `Command` classes as there are products?

aparajit-pratap:
Apart from minor comments, I don't have any others. Please try and move the tests from https://github.com/DynamoDS/Dynamo/pull/3987 to this repo if possible in case you are not submitting that PR.

ikeough:
@sharadkjaiswal I'm going to ask a dumb question... Does a framework already exist to do what you've developed here? It seems like there must be an existing framework for getting all installed versions of an application based on product codes. Have we developed this because we have legacy products that don't use the same scheme and need special logic?

sharadkjaiswal:
@ikeough from 0.8 onwards we are going to follow a specific convention to write information about Dynamo installs into registry, we need to use that convention to detect the product installations. For older version, we either track by looking at specific folders or looking into the registry for product code used by iss installer. But to support side-by-side installation, product code needs to be changed.

I hope this answers your question, we can discuss offline if you need more info on the approach.

aparajit-pratap:
LGTM once PR: https://github.com/DynamoDS/Dynamo/pull/3987 is merged

PR #256, opened 2015-03-16
Version Selector support for all Dynamo installs

sharadkjaiswal:
Merge #248 and #255 to Revit2016 branch.


sharadkjaiswal:
@aparajit-pratap PTAL

PR #267, opened 2015-03-18
Categorize one test case to the Failure category

Randy-Ma:
There is one test case which is failing because of other component. Before the conclusion is made, I am temporarily marking this test case as Failure.

@ke-yu 
PTAL


ke-yu:
LGTM

PR #279, opened 2015-03-23
Fix MAGN-6738  Dynamo 0.8.0 throws error while launching Dynamo ...

Randy-Ma:
<h6>Summary</h6>

The error is thrown because RevitServiceUpdater is initialized twice. I am removing one because the RevitServiceUpdater has been changed in: https://github.com/DynamoDS/Dynamo/pull/3554. In this pull request, the initialization is done when the application plugin starts.

@sharadkjaiswal 
PTAL


sharadkjaiswal:
Thanks, @Randy-Ma LGTM

PR #280, opened 2015-03-23
Kill Infinite Loops

ikeough:
This PR fixes an issue described in [MAGN-6360](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6360), where certain `Revit.Element` types would cause infinite looping when they set property values on Revit elements and subsequently responded to the document modification events which those editions caused.

Here we do two things to "cut the loop":
- We introduce the `CurveUtilities.AreCurvesSimilar` method which compares two curves for similarity. For `Revit.Element` types where we set a curve on an internal Revit element, this allows us to check the curves for similarity before setting to avoid causing a document modification.
- We add checks at a number of locations in the code for curve similarity or object value to avoid setting a property on a revit element if it's not absolutely necessary.

PTAL:
- [x] @Randy-Ma (because you investigated for me)

Requires Merge:
- [ ] Revit2015
- [ ] Revit2016
- [ ] RC0.8.0_Revit2014
- [ ] RC0.8.0_Revit2015
- [ ] RC0.8.0_Revit2016

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/CurveUtils.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/CurveUtils.cs
@@ -1,11 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Windows.Media;
 
 using Autodesk.DesignScript.Runtime;
 using Autodesk.Revit.DB;
+
+using DynamoUnits;
 
 namespace Revit.GeometryConversion
 {
@@ -28,7 +28,7 @@ namespace Revit.GeometryConversion
                 XYZ norm = null;
 
                 //keep old plane computations
-                if (System.Math.Abs(p0.Z - p2.Z) < Tolerance)
+                if (Math.Abs(p0.Z - p2.Z) < Tolerance)
                 {
                     norm = XYZ.BasisZ;
                 }
@@ -69,7 +69,7 @@ namespace Revit.GeometryConversion
             return bestFitPlane.ToPlane(false);
         }
 
-        public static bool IsLineLike(Autodesk.Revit.DB.Curve crv)
+        public static bool IsLineLike(Curve crv)
         {
             if (crv is Line) return true;
             if (crv is HermiteSpline) return IsLineLikeInternal(crv as HermiteSpline);
@@ -79,14 +79,28 @@ namespace Revit.GeometryConversion
             return false;
         }
 
+        public static bool CurvesAreSimilar(Curve a, Curve b, double tolerance = 1.0e-6)
+        {
+            if (a.GetType() != b.GetType())
+                return false;
+
+            if (!CurvesHaveSameEndpoints(a, b, tolerance))
+                return false;
+
+            dynamic ad = a;
+            dynamic bd = b;
+
+            return AreSimilar(ad, bd, tolerance);
+        }
+
         #region IsLineLike Helpers
 
-        private static bool IsLineLikeInternal(Autodesk.Revit.DB.NurbSpline crv)
+        private static bool IsLineLikeInternal(NurbSpline crv)
         {
             return IsLineLikeInternal(crv.CtrlPoints);
         }
 
-        private static bool IsLineLikeInternal(Autodesk.Revit.DB.HermiteSpline crv)
+        private static bool IsLineLikeInternal(HermiteSpline crv)
         {
             var controlPoints = crv.ControlPoints;
             var controlPointsAligned = IsLineLikeInternal(controlPoints);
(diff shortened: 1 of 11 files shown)


src/Libraries/RevitNodes/GeometryConversion/CurveUtils.cs
```diff
@@ -124,6 +149,72 @@ private static bool IsLineLikeInternal(IList<XYZ> points)
...
+        private static bool AreSimilar(Line a, Line b, double tolerance)
+        {
+            return CurvesHaveSameEndpoints(a,b, tolerance);
+        }
+
+        private static bool AreSimilar(Arc a, Arc b, double tolerance)
+        {
+            return a.Center.IsAlmostEqualTo(b.Center) && 
+                a.Radius.AlmostEquals(b.Radius, 1.0e-6) && 
+                a.IsCyclic == b.IsCyclic;
```
Randy-Ma: Are we assuming the inputs are always full circles? Otherwise, I think it is better to compare the start angles, end angles and normals.

src/Libraries/RevitNodes/GeometryConversion/CurveUtils.cs
```diff
@@ -124,6 +149,72 @@ private static bool IsLineLikeInternal(IList<XYZ> points)
...
+                a.Radius.AlmostEquals(b.Radius, tolerance) &&
+                a.XVector.IsAlmostEqualTo(b.XVector) &&
+                a.YVector.IsAlmostEqualTo(b.YVector) &&
+                a.ZVector.IsAlmostEqualTo(b.ZVector);
+        }
+
+        private static bool CompareRandomParameterLocationDistances(
+            Curve a, Curve b, double tolerance)
+        {
+            for (var i = 0; i < 10; i++)
```
Randy-Ma: Ten points happen to be the same can not logically prove that the two curves are the same.

src/Libraries/RevitNodes/GeometryConversion/CurveUtils.cs
ikeough: The start point and end point are already compared in an outer check. I agree we should add a normal check.

src/Libraries/RevitNodes/GeometryConversion/CurveUtils.cs
ikeough: I agree 10 is too few. But, we need to balance this call being too expensive with being fool-proof. You could say the same about 10,000 points being the same. We have to start somewhere :)

Randy-Ma:
@ikeough 
Thanks for the fix! 
But in my opinion, if the input of some nodes has not changed, ideally the graph should not run again. Or even if we re-execute the graph, the part with no changes should not run again. Ideally we do not even need to check whether the input has changed or not when the function is called.

ikeough:
@Randy-Ma You're right. This is not the perfect fix. 

I think we might be able to use the HomeWorkspaceModel.GetExecutingNodes API that was created for the run preview work to limit the number of OnNodeModified calls that are triggered in the document modification handlers. That is the only way, that I can think of, to determine whether the node is going to execute during the next run.

src/Libraries/RevitNodes/GeometryConversion/CurveUtils.cs
Randy-Ma: @ikeough 
I really really do not like to compare curves in this way. :(

Randy-Ma: Bear with me. :)

Randy-Ma:
@ikeough 
I have just had an discussion with @junmendoza. The current logic to cut the infinite loop looks good to me. But for comparing the curves, I do hope we can improve the algorithm. Other than this, LGTM! Also thanks a lot for your patient work to cut the infinite loop for other elements!

ikeough:
@Randy-Ma @junmendoza What I proposed above, to use the `GetExecutingNodes` API won't work because after the document modification event is handled, but before the modification event is raised on the node, the node won't be part of the collection of nodes that are going to be executed, as reported by that API. So we wouldn't be able to skip raising their modified event based on their inclusion in that list.

Here's another idea...

In RevitDynamoModel we maintain a list called `LoopCandidates`. After execution we insert the Guid of every node that just executed into this list. When the document modification event is handled, as we loop over the collection of nodes whose bound elements have been modified, we check if that node's Guid is in the `LoopCandidates` list. If so, we don't raise a modification event for that node. After all nodes are modified, we clear the `LoopCandidates` list.

For the case where an execution of the node raises a document modification event, this will prevent the same node from re-executing as that modification event is handled. For the case where some interaction in Revit modifies the element, this will still allow for the node to raise its modified event.

PR #281, opened 2015-03-24
Use runtimecore to get runtime warnings

junmendoza:
@Randy-Ma


Randy-Ma:
As I can see, these are changes applied according to some changes in the engine to fix the build errors! LGTM!

PR #284, opened 2015-03-24
Move one test case out of the Failure category

Randy-Ma:
<h4>Summary</h4>
As the test case which is temporarily marked as failure has been fixed, I am moving this test case out of the Failure category.

@junmendoza 
PTAL


junmendoza:
LGTM

PR #295, opened 2015-03-25
Fix MAGN-6737 File Exit crashes Revit

Randy-Ma:
<h4>Summary</h4>

The singleton instance of RevitServicesUpdater is disposed but later it is still used by the model. I am moving the disposing method call from where the view is closed to where the model is shutdown.

@Benglin 
PTAL


Benglin:
Thanks a lot for helping to merge, @pboyer!

PR #297, opened 2015-03-25
Fix MAGN-6737 File Exit crashes Revit

Randy-Ma:
<h4>Summary</h4>

The singleton instance of RevitServicesUpdater is disposed but later it is still used by the model. As the instance is only instantiated once when the plugin is initialized and so I am removing the call to dispose the instance when the view is closed.

@Benglin 
PTAL


pboyer:
As this is the same fix I conisdered, I'm just merging this in.

Benglin:
Thanks a lot for helping to merge, @pboyer!

PR #301, opened 2015-03-25
MAGN-5850 Element Binding / Dynamo not properly deleting Revit Elements created in previous Dynamo session by same file

ikeough:
This pull request fixes [MAGN-5850](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-5850#comment=61-17451), by adding an element reconciliation step after the first evaluation.

## Why
Open a Revit model that, for example, has 5 structural framing members that were created by Dynamo. Then open the corresponding dyn file, and adjust some parameters that affect the layout of the structural framing such that you will get 2 structural framing members, and hit Run. What you get in Revit is 2 elements that are in the right place and 3 that haven't been touched. Those 3 elements should be deleted. But Dynamo doesn't know that it's supposed to delete them, because it doesn't know anything about the historical state of the model.

## How The Fix Works
After a `HomeWorkspaceModel` is added to `DynamoModel`, we check to see whether the workspace has `PreloadedTraceData`. This will be the raw trace data as it is serialized into the .dyn file. If trace data is available, we deserialize the data and build a map for the workspace of previously executed nodes and the elements ids that were associated with them. This is stored in the `historicalElementData` field. This map is built in such a way as to support multiple home workspaces in the future. 

After evaluation is completed, we check whether there is historical element data available for the workspace. If there is we proceed with element reconciliation. During element reconciliation we build another element id map, this time using the trace data from the current run. We compare the current map with the historical map and we identify any elements that were created historically, but were not created during this run. Those "orphan" elements are then deleted. At the conclusion of element reconciliation we delete the current workspace's historical id map so that this reconciliation never takes place again.

This PR requires [this core PR](https://github.com/DynamoDS/Dynamo/pull/4093).

PTAL:
- [x] @Randy-Ma (Because I think you'll get a kick out of this...and because of some small fixes to `ElementIdLifeCycleManager`)

Requires Merge:
- [ ] Revit2015
- [ ] Revit2016

DO NOT APPLY THIS CHANGE TO THE 0.8.0 RC BRANCHES!!

### DIFF
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -1,8 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Formatters.Soap;
 using System.Windows.Forms;
 
 using Autodesk.Revit.DB;
@@ -20,6 +23,8 @@ using Dynamo.Utilities;
 using DynamoServices;
 
 using Greg;
+
+using ProtoCore;
 
 using Revit.Elements;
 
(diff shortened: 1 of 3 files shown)


src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -360,12 +419,101 @@ private void OnApplicationViewActivated(object sender, ViewActivatedEventArgs e)
...
+
+            foreach (var kvp in workspaceHistory)
+            {
+                var nodeGuid = kvp.Key;
+
+                // If the current run doesn't have a key for 
+                // this guid, then all of these elements are 
+                // orphaned.
+
+                if (!currentRunData.ContainsKey(nodeGuid))
```
Randy-Ma: So this is assuming some nodes are removed from the graph?

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -360,12 +419,101 @@ private void OnApplicationViewActivated(object sender, ViewActivatedEventArgs e)
...
+                    orphanedIds.AddRange(kvp.Value);
+                    continue;
+                }
+
+                var currentRunNodeGuids = currentRunData[nodeGuid];
+
+                // If the current run didn't create an element
+                // in the current run, then add an orphan.
+
+                orphanedIds.AddRange(kvp.Value.Where(id => !currentRunNodeGuids.Contains(id)));
```
Randy-Ma: I do not quite understand the logic here. Who will delete the items in the pre-loaded trace data? Assuming the input is changed from 1..3 to 1..1, there should still be 3 items in the trace data for the call site.

Randy-Ma:
@ikeough 

This is amazing!

I temporarily have not seen any side effects to dispose the life cycle manager instance. 

The sync based on work space id looks like a good idea.

src/DynamoRevit/Models/RevitDynamoModel.cs
ikeough: This is a good question.

When we go from 1..3 to 1..1 there are now fewer items in the current run's trace data. The old ones are no longer there. That's why this works. I didn't delve into how call sites clean up, but there is definitely not trace data for the old elements hanging around after the first run. This is why this works.

src/DynamoRevit/Models/RevitDynamoModel.cs
ikeough: Yes, this would be the case that before you run, you delete a node.

src/DynamoRevit/Models/RevitDynamoModel.cs
Randy-Ma: @ikeough 
As I can understand, the old ones should still be there according to the current implementation.

ikeough: @Randy-Ma 

I would say that we can delete the out of date trace data as we do the comparison, but the data is all packed together in a SingleRunTraceData object. The serialized blob represents the whole thing. I can just delete specific elements from those SingleRunTraceData objects. If we don't delete them, then they will be left around forever and potentially saved back to the file, which is something that I didn't check.

The other strategy that I tried was to preload the trace data before the first run. Then we could go and delete specific ids off the SingleRunTraceData at the CallSites. But when I tried this, I got some null refs from the engine which scared me away. Perhaps I should try this again.

Randy-Ma: @ikeough 

I just checked the logic related to trace data and verified that the current run trace data will not include those that are left over. For example, if we change from 0..3 to 0..1, the current run trace data will only contain 2 instead of 4.

Based on this, your current logic is correct.

Randy-Ma:
@ikeough 

Thanks a lot for the fix! 

I have just verified the logic related to the trace data! Your logic is correct! LGTM!

PR #302, opened 2015-03-26
Fix MAGN-6762 Revit crashes after reopen of project after family creation

Randy-Ma:
<h4>Summary</h4>

Because now we have moved the event handlers related to opening/closing documents to the model, when the model is shutdown, the event handlers will be unregistered. As a result, before a new model is created and the event handlers are registered, the events will not be handled. Even a document opening event will not be handled and the document hold by the document manager instance is invalid.

This submission makes the change to unregister the events for some nodes so there will be no chance to access the document to avoid a crash.

This will also help avoid memory leaks.

@Benglin 
PTAL


Benglin:

PR #303, opened 2015-03-26
Fix possible memory leaks

Randy-Ma:
<h4>Summary</h4>

This is to apply the changes from:
https://github.com/DynamoDS/DynamoRevit/pull/302

There is no crash here but it is better for us to apply the changes to avoid memory leaks.

@Benglin 
PTAL


Benglin:

PR #304, opened 2015-03-26
Fix possible memory leaks

Randy-Ma:
<h4>Summary</h4>

This is to apply the following changes:
https://github.com/DynamoDS/DynamoRevit/pull/302

@Benglin 
PTAL


Benglin:

PR #307, opened 2015-03-26
Fix possible memory leaks

Randy-Ma:
This is to merge the changes in:
https://github.com/DynamoDS/DynamoRevit/pull/304

@Benglin 
PTAL


Benglin:

PR #309, opened 2015-03-27
Updated Revit related components to preload libraries

Benglin:
This is done for the built-in libraries are now removed from `DynamoModel.Preloader` (see [another pull request](https://github.com/DynamoDS/Dynamo/pull/4050) for details). Now Revit components (include unit test cases) assume the responsibility of specifying which built-in libraries to preload for the session.

Hi @ikeough, please have a look, nothing too surprising I hope (if you need any background, @pboyer can fill you in). Thanks in advance! :)

### DIFF
--- a/test/Libraries/RevitTestServices/RevitTestPathResolver.cs
+++ b/test/Libraries/RevitTestServices/RevitTestPathResolver.cs
@@ -1,18 +1,26 @@
 ﻿using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-using Dynamo.Interfaces;
+using TestServices;
 
 namespace RevitTestServices
 {
-    public class RevitTestPathResolver : IPathResolver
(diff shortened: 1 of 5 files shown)


test/Libraries/RevitTestServices/RevitTestPathResolver.cs
```diff
@@ -1,18 +1,26 @@
 ﻿using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-using Dynamo.Interfaces;
+using TestServices;
 
 namespace RevitTestServices
 {
-    public class RevitTestPathResolver : IPathResolver
+    public class RevitTestPathResolver : TestPathResolver
```
Benglin: Bulk of the code below is removed because `TestPathResolver` already implements them.

PR #318, opened 2015-04-02
Fixed regression failures

Benglin:
Added additional node directory and resolve path.

Hi @Randy-Ma please take a look, thanks so much! :)


Randy-Ma:
LGTM

PR #321, opened 2015-04-02
Build Fix for PR #4157

sharadkjaiswal:
- ArgumentLacing property has been made readonly, so it can't be set externally.
- Use UpdateValue method to change ArgumentLacing property.

@Benglin  PTAL


Benglin:
Thanks so much for helping to fix this, @sharadkjaiswal. LGTM!

PR #324, opened 2015-04-02
Fixed Revit build failures due to ArgumentLacing

Benglin:


sharadkjaiswal:
I have already submitted the fix.

PR #327, opened 2015-04-07
Family rotation

ShengxiZhang:
Add a new node to set the rotation of the in-placed family instances.


ShengxiZhang:
@LianweiHuang 
@XiaojieZhang
@aandu

ShengxiZhang:
@kiki-zhang

ShengxiZhang:
@ikeough  @Randy-Ma

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -298,5 +300,39 @@ public override string ToString()
...
+           double[] oldRotationAngles;
+           TransformUtils.ExtractEularAnglesFromTransform(oldTransform, out oldRotationAngles);
+
+           Double newRotationAngle = degree * Math.PI / 180;
+
+           if (!oldRotationAngles[0].AlmostEquals(newRotationAngle, 1.0e-6))
+           {
+              double rotateAngle = newRotationAngle - oldRotationAngles[0];
+              var axis = Line.CreateUnbound(oldTransform.Origin, oldTransform.BasisZ);
+              ElementTransformUtils.RotateElement(Document, new ElementId(Id), axis, -rotateAngle);
```
Randy-Ma: Where does this function come from?

src/Libraries/RevitNodes/Elements/InternalUtilities/TransformUtils.cs
```diff
@@ -0,0 +1,47 @@
...
+using Autodesk.Revit.DB;
+using System.Windows.Media.Media3D;
+
+namespace Revit.Elements.InternalUtilities
+{
+   [IsVisibleInDynamoLibrary(false)]
+   public class TransformUtils
+   {
+      // decomp_ZYX
+      public static void ExtractEularAnglesFromTransform(Transform transform, out double[] angles )
```
Randy-Ma: ExtractEulerAnglesFromTransform?

ShengxiZhang: Oh! Sorry for the typo! Eular=>Euler

Randy-Ma:
@ShengxiZhang 
Have you tried running a graph with the new node? I doubt there will be an infinite loop.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
ShengxiZhang: This is from RevitAPI assembly.

ShengxiZhang:
Yes, I tried the node. There is no infinite loop because here we set a fixed rotation, so from the 2nd time, the document won't be changed so the node won't be executed.

Randy-Ma:
There should be no infinite loops some time back. But now it is slightly different. Could you help add a system test case? In the test case, open and run a .dyn file instead of calling APIs.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
Randy-Ma: Thanks!
But I do not understand why we put a minus sign before rotateAngle.

ShengxiZhang: Ah, from the testing, it looks the rotation direction is different between the euler angle and RotateElement.

Randy-Ma: But then if the rotating is not along a vector, how shall we define the rotation?

ShengxiZhang: We define the rotation is always along its local Z-axis, this is why we removed the argument rotation_axis. Does it make sense?

ShengxiZhang:
@Randy-Ma 
Hi Randy, I have added the system test case, could you please review it? Thank you!

Randy-Ma:
@ShengxiZhang 
Great! Thanks a lot! LGTM!

Randy-Ma:
But do not forget to cross-merge the changes here.

PR #328, opened 2015-04-07
Fix MAGN-6710 Run CurveByPoints.byReferencePoints will crash Revit and Dynamo

Randy-Ma:
<h4>Summary</h4>

There is an infinite loop when creating an instance of CurveByPoints. The reason is because when creating the instance, the document is modified. Our document watcher will search the nodes which are causing the modification and re-run the nodes, and then the document is modified again, .... Thus there is an infinite loop.

The fix here is to avoid modifying the curve when it is found the input to set the curve does not change.

I have also added a test case for this.

@ikeough 
PTAL

### DIFF
--- a/src/Libraries/RevitNodes/Elements/CurveByPoints.cs
+++ b/src/Libraries/RevitNodes/Elements/CurveByPoints.cs
@@ -8,6 +8,8 @@ using DynamoServices;
 using RevitServices.Persistence;
 using RevitServices.Transactions;
 
+using Revit.GeometryConversion;
+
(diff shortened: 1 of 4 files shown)


Randy-Ma:
@sharadkjaiswal 
PTAL

ikeough:
@Randy-Ma Thanks! I'm sure we'll find more of these :) LGTM.

Randy-Ma:
@ikeough 
Nice catch! I have updated my code to use the standard API which also have a version to set a precision.

PR #334, opened 2015-04-09
Fixed "Resources.resx" for "RevitNodes"

Benglin:
This pull request attempts to do the following:

1. Align `*.resx` in `Revit_2014` to match with those of `Revit_2015` and `Revit_2016` branches.
2. We do not want `DynamoRevit.csproj` to generate documentation `XML` file (because the file is used for localization process, and it is not user-facing).

Hi @Randy-Ma, could you please have a look at this? Thanks!


Randy-Ma:
@Benglin 
Thanks for the fix! LGTM!

PR #335, opened 2015-04-09
Fixed "Resources.resx" for "RevitNodes"

Benglin:
This pull request attempts to do the following:

1. Align `*.resx` in all of  `Revit_2014`, `Revit_2015` and `Revit_2016` branches.
2. We do not want `DynamoRevit.csproj` to generate documentation `XML` file (because the file is used for localization process, and it is not user-facing).

Hi @Randy-Ma, could you please have a look at this? Thanks!


src/DynamoRevit/DynamoRevit.csproj
```diff
@@ -36,6 +36,8 @@
     <Prefer32Bit>false</Prefer32Bit>
     <DocumentationFile>$(OutputPath)\DynamoRevitDS.XML</DocumentationFile>
     <DocumentationFile>$(OutputPath)\$(UICulture)\DynamoRevitDS.xml</DocumentationFile>
```
Randy-Ma: As I can see from the summary, the xml files will not be generated. But it is not the case here.

src/DynamoRevit/DynamoRevit.csproj
```diff
@@ -48,6 +50,8 @@
...
```
Randy-Ma: As I can see from the summary, the xml files will not be generated. But it is not the case here.

src/Libraries/RevitNodes/Properties/Resources.en-US.resx
```diff
@@ -144,7 +144,7 @@
   <data name="CurtainPanelShouldRepresentRevitPanel" xml:space="preserve">
     <value>Curtain Panel should represent Revit panel</value>
   </data>
-  <data name="NumberOfDivisionsShouldBeGreaterThan2" xml:space="preserve">
+  <data name="NumberOfDivisionMustBeGreaterThan2" xml:space="preserve">
```
Randy-Ma: NumberOfDivisionsMustBeGreaterThan2 looks better than NumberOfDivisionMustBeGreaterThan2

src/Libraries/RevitNodes/Properties/Resources.resx
```diff
@@ -144,7 +144,7 @@
   <data name="CurtainPanelShouldRepresentRevitPanel" xml:space="preserve">
     <value>Curtain Panel should represent Revit panel</value>
   </data>
-  <data name="NumberOfDivisionsShouldBeGreaterThan2" xml:space="preserve">
+  <data name="NumberOfDivisionMustBeGreaterThan2" xml:space="preserve">
```
Randy-Ma: NumberOfDivisionsMustBeGreaterThan2 looks better than NumberOfDivisionMustBeGreaterThan2

src/DynamoRevit/DynamoRevit.csproj
```diff
@@ -36,6 +36,8 @@
...
+    <DocumentationFile>
+    </DocumentationFile>
```
Randy-Ma: The xml tag does not contain anything. we can remove it.

src/DynamoRevit/DynamoRevit.csproj
```diff
@@ -48,6 +50,8 @@
...
```
Randy-Ma: The xml tag does not contain anything. we can remove it.

src/DynamoRevit/DynamoRevit.csproj
Benglin: Argh! That's `cherry-pick` being smart on me! Okay will fix this, good catch.

src/Libraries/RevitNodes/Properties/Resources.en-US.resx
Benglin: That's a great suggestion! Updated :)

Benglin:
Hi @Randy-Ma, those are really good catches. I have updated the PR, please have a look, thanks!

Randy-Ma:
@Benglin 
Thanks for the fix and the update! LGTM!

PR #337, opened 2015-04-14
Fix MAGN-3419 Element.Parameter not returning value for Mark

Randy-Ma:
<h4>Summary</h4>

AsValueString is not working when the parameter's storage type is string. It is found another function AsString will work if the storage type is string. This submission makes the fix by making the corresponding call based on the storage type.

@ikeough 
PTAL


ikeough:
LGTM. Thanks Randy!

PR #347, opened 2015-04-17
Implement FacingOrientation for FamilyInstance

Randy-Ma:
<h4>Summary</h4>

This is to implement FacingOrientation for FamilyInstance. As FacingOrientation is only defined for FamilyInstance, so the implementation here is for FamilyInstance only. 

I have also added a test case for this.

@ikeough 
PTAL


Randy-Ma:
@ke-yu 
PTAL

ke-yu:
LGTM

ikeough:
@Randy-Ma Sorry I didn't get to this Randy. Looks good. Is this attached to a particular defect?

PR #351, opened 2015-04-29
MAGN-7268 Dynamo fails to load  within Revit if the system  Environment variable "Path" has string within curly braces e.g.{0}

ikeough:
### Purpose
This PR solves an issue raised in https://github.com/DynamoDS/Dynamo/issues/4256#issuecomment-9740693.

### Declarations
  - Without adjusting the system path for the Revit test process, we cannot test this.

### Reviewers
- [x] @ramramps

### Requires Merge
- [ ] Revit2015
- [ ] Revit2016

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -142,10 +142,9 @@ namespace Dynamo.Applications
             // Add the main exec path to the system PATH
             // This is required to pickup certain dlls.
             var path =
-                string.Format(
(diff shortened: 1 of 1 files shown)


RodRecker:
Looks good for a starting point for the template. Thanks for doing it. I think we would agree that the testing needs to be resolved before calling it complete.

ramramps:
LGTM !!

RodRecker:
@ikeough this is approved by Ram - do you want to go ahead and merge it?

PR #353, opened 2015-04-30
Test revit node 

zora-wang:
workflow test part1: wall
@monikaprabhu 
@riteshchandawar


monikaprabhu:
@zora-wang - Please change the title of the PR to Workflow Tests with Revit nodes.
Also please add verification for all the tests- Similar to Workflow_test01

PR #356, opened 2015-05-06
Fix MAGN-6862 Revit_ImportSolid sample goes in infinite loop when this sample opened after few other samples

Randy-Ma:
<h4>Summary</h4>

There are some selection nodes in the graph. Once an import instance is created and if the node is run for the second time, the element binding happening for ImportInstance will delete the old instance and create a new one. The deleting operation will trigger some document event handlers. The event handler in the selection nodes will cause the graph run again. Then the ImportInstance node will re-run and delete and create and trigger the document event handler.... Thus there is an infinite loop.

In this submission, I am cutting the infinite loop by checking whether the deleted objects contains any one in the selected objects for the selection node, and if not, the graph will not be run. 

@ikeough 
PTAL

### DIFF
--- a/src/Libraries/RevitNodesUI/Selection.cs
+++ b/src/Libraries/RevitNodesUI/Selection.cs
@@ -225,6 +225,13 @@ namespace Dynamo.Nodes
                 return;
             }
 
+            // If the deleted list doesn't include any objects in the current selection
+            // then return;
+            if (!SelectionResults.Select(x => x.Id).Any(deleted.Contains))
+            {
+                return;
+            }
+
             // We are given a set of ElementIds, but because the elements
             // have already been deleted from Revit, we can't get the 
             // corresponding GUID. Instead, we just go through the collection of


ikeough:
This is great @Randy-Ma, and @mccrone will be happy for it. Colin, this is the infinite loop that I was describing with many workspaces that have Selection nodes in them. 

Randy, I'd like a little more documentation around that condition to know exactly why it's there. I believe it's there because the `UpdateSelection` method call that follows it will call `NodeModified`. But you have to get pretty far into it to figure that out. We could also change the name of `UpdateSelection` to something like `UpdateSelectionAndNotify`.

mccrone:
+1!

Randy-Ma:
@ikeough 
Thanks for the comments! Right now UpdateSelection will set SelectionResults. It is the set method for SelectionResults which will call OnNodeModified and cause the graph run again if the graph is set to run automatically. I think it is reasonable because when you change the selected objects, the graph need to run again(for automatic run).

Randy-Ma:
@ikeough 
This will fix MAGN-7075 as well.

ikeough:
@Randy-Ma Sorry Randy! I hadn't realized you were still waiting for my response. LGTM.

PR #357, opened 2015-05-06
Fix MAGN-6862 Revit_ImportSolid sample goes in infinite loop when this sample opened after few other samples

Randy-Ma:
<h4>Summary</h4>

This is to cross merge the fix in:
https://github.com/DynamoDS/DynamoRevit/pull/356

@ikeough 
PTAL


Randy-Ma:
I am merging this based on comments in https://github.com/DynamoDS/DynamoRevit/pull/356

PR #370, opened 2015-05-17
MAGN-7262 Line created using Select Model Element, coming with default white color.

ikeough:
### Purpose

This PR addresses [MAGN-7262](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7262). `CurveElement` objects in revit were visualizing with white lines instead of the appropriate edge color. 

### Declarations

Check these iff you believe they are true

  - `DisplayTests` class added to Revit tests to ensure against regression.

### Reviewers

@ramramps 

### Merges
- [ ] Revit2015
- [ ] Revit2016


ramramps:
Is it OK to move the Extension class RenderPackageExtensions to a new file?  Other than that , LGTM !!

ramramps:
ping @ikeough

PR #373, opened 2015-05-18
categorize panels tests as failure due to lack of lunchbox packange

zora-wang:
Purpose

These two test cases are failed due to lack of Lunchbox package in testing machine.
so we choose to categorize them as failure now.

Reviewers

@monikaprabhu


monikaprabhu:
LGTM.

PR #374, opened 2015-05-18
Fix failing test cases

Randy-Ma:
### Purpose

This submission is to fix two regressions caused by changes in the selection nodes. The original logic has issues in that if SelectionResults contain some deleted elements, then trying to get Id from these elements will throw exceptions.

An attempt to fix the issue by filtering SelectionResults to check whether the element is valid. But the filtered result will never have even one element contained by the deleted elements. So in this submission, the issue is fixed by checking whether SelectionResults contains any invalid elements.

### Declarations

### Reviewers

@aparajit-pratap


aparajit-pratap:
LGTM!

PR #377, opened 2015-05-19
Added test case for defect MAGN-6862

riteshchandawar:
### Purpose
To catch regression in future.
Test Case for defect http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6862
PRs for Revit 2015 #378 and 2016 branch #379. 
### Declarations

Check these iff you believe they are true

- [x] The code base is in a better state after this PR
- [x] The level of testing this PR includes is appropriate

### Reviewers

@Randy-Ma 

### FYIs

@ikeough 

**Note** Test case is passing with recent build.
image

PR #378, opened 2015-05-19
Text for magn6862

riteshchandawar:
Merging changes from https://github.com/DynamoDS/DynamoRevit/pull/377

Reviewer:
@Randy-Ma


Randy-Ma:
@riteshchandawar 
I just noticed this. But it seems this need to be updated.

PR #380, opened 2015-05-19
Correct the filtering logic and add a test case

Randy-Ma:
### Purpose

This submission is to correct the filtering logic which is made when I am fixing failing test cases in https://github.com/DynamoDS/DynamoRevit/pull/374.

A test case has been added to track regressions.

### Declarations

### Reviewers

@aparajit-pratap

### DIFF
--- a/src/Libraries/RevitNodesUI/Selection.cs
+++ b/src/Libraries/RevitNodesUI/Selection.cs
@@ -227,7 +227,7 @@ namespace Dynamo.Nodes
 
             // If the deleting operations does not make any elements in SelectionResults
             // invalid, then there is no need to update.
-            if (SelectionResults.Select(el => !el.IsValidObject).Count() == 0)
+            if (SelectionResults.Where(el => !el.IsValidObject).Count() == 0)
             {
                 return;
             }
--- a/test/Libraries/RevitIntegrationTests/BugTests.cs
+++ b/test/Libraries/RevitIntegrationTests/BugTests.cs
@@ -19,6 +19,7 @@ using RTF.Framework;
 using RevitServices.Persistence;
 using System.Collections.Generic;
 using RevitServices.Transactions;
+using Revit.GeometryConversion;
 
 using DoubleSlider = DSCoreNodesUI.Input.DoubleSlider;
 using IntegerSlider = DSCoreNodesUI.Input.IntegerSlider;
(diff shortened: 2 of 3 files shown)


src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -227,7 +227,7 @@ protected override string GetIdentifierFromModelObject(TSelection modelObject)
 
             // If the deleting operations does not make any elements in SelectionResults
             // invalid, then there is no need to update.
-            if (SelectionResults.Select(el => !el.IsValidObject).Count() == 0)
+            if (SelectionResults.Where(el => !el.IsValidObject).Count() == 0)
```
Randy-Ma: @aparajit-pratap 
Here, Where should be used.

aparajit-pratap: Oh yes, I overlooked this. Wonder how it passed the tests?

test/Libraries/RevitIntegrationTests/BugTests.cs
```diff
@@ -621,6 +622,85 @@ public void MAGN_6710()
...
+            using (var trans = new Autodesk.Revit.DB.Transaction(document, "DeleteReferencePoint"))
+            {
+                trans.Start();
+                document.Delete(lineID);
+                trans.Commit();
+            }
+
+            RunCurrentModel();
+            //There should be no infinite loop, otherwise, there will be an error with this test case.
+        }
```
aparajit-pratap: So this test just checks if the deletion of a Revit element happens successfully?

Randy-Ma: This is to check whether an update in the graph, e.g., deleting an element will cause the graph to  run endlessly.

src/Libraries/RevitNodesUI/Selection.cs
Randy-Ma: Because it somehow avoids the call to get the element ID for invalid objects.

aparajit-pratap:
In addition you could check if there is no change in state of the graph at the end of the test case, i.e. the graph has not re-executed since that is what we need to essentially test here. Apart from this minor comment, LGTM.

aparajit-pratap:
Please give a thought to the declarations made in the PR description and whether they are true.

PR #381, opened 2015-05-19
Correct the filtering logic and add a test case

Randy-Ma:
This is to merge the changes in:
https://github.com/DynamoDS/DynamoRevit/pull/380


Randy-Ma:
Merging this based on comments in https://github.com/DynamoDS/DynamoRevit/pull/380

PR #385, opened 2015-05-25
Fix the failing test case named as Regressions

Randy-Ma:
### Purpose

This is to fix the failing test case named as Regressions. The test case is failing is because the ViewModel is still used after it is assigned null. In this fix, I am removing the redundant lines of code as in the shutdown method, the same code will run.

### Declarations

### Reviewers

@aparajit-pratap 

### FYIs

@ikeough

### DIFF
--- a/test/Libraries/RevitIntegrationTests/RegressionTests.cs
+++ b/test/Libraries/RevitIntegrationTests/RegressionTests.cs
@@ -69,9 +69,6 @@ namespace RevitSystemTests
             }
             finally
             {
-                ViewModel.Model.ShutDown(false);
-                ViewModel = null;
-                RevitServicesUpdater.DisposeInstance();
                 TearDown();
             }
 


test/Libraries/RevitIntegrationTests/RegressionTests.cs
```diff
@@ -69,9 +69,6 @@ public void Regressions(RegressionTestData testData)
             }
             finally
             {
-                ViewModel.Model.ShutDown(false);
-                ViewModel = null;
-                RevitServicesUpdater.DisposeInstance();
```
aparajit-pratap: are these called later on inside the `TearDown` method?

Randy-Ma: Yes, it will be called by ViewModel.PerformShutdownSequence

aparajit-pratap:
LGTM!

PR #389, opened 2015-05-25
Update Manager configuration with specific lookup

sharadkjaiswal:
### Purpose
Initializes UpdateManager with default configuration and DynamoRevitLookUp to start DynamoRevitModel.

- DynamoRevitLookUp implements abstract class DynamoLookUp to get Dynamo installations on the system.
- UpdateManagerConfiguration is updated with DynamoRevitLookUp instance.
- It's a follow-up of https://github.com/DynamoDS/Dynamo/pull/4556

### Declarations

### Reviewers
- [x] @aparajit-pratap

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -2,12 +2,12 @@ using Dynamo.Applications.ViewModel;
 
 using Dynamo.Applications;
 using Greg.AuthProviders;
-using RevitServices.Elements;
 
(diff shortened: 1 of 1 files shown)


aparajit-pratap:
@sharadkjaiswal test cases?

sharadkjaiswal:
We already have test cases for DynamoLookUp class and we can't write test cases for registry lookup.

aparajit-pratap:
LGTM

PR #418, opened 2015-05-28
Part 2 of fixing MAGN-7251 [Regression from 0.7.5] Select elements will cause a crash after switching Revit documents

Randy-Ma:
### Purpose

This is the second part to fix MAGN-7251. The fix here is to add a property called RevitDynamoModel in RevitSelection. The property is set when the related view customization is constructed. As this property will only be set when the view is created, there may be possibility that the property is null. So for places where this property is used, a null check will be conducted.

Refer the first part of the fix:
https://github.com/DynamoDS/Dynamo/pull/4594

### Declarations

### Reviewers

@Benglin 

### FYIs

### DIFF
--- a/src/Libraries/RevitNodesUI/Selection.cs
+++ b/src/Libraries/RevitNodesUI/Selection.cs
@@ -35,14 +35,48 @@ namespace Dynamo.Nodes
     public abstract class RevitSelection<TSelection, TResult> : SelectionBase<TSelection, TResult>
     {
         protected Document SelectionOwner { get; private set; }
-        protected RevitDynamoModel RevitDynamoModel { get; private set; }
+        private RevitDynamoModel revitDynamoModel;
 
         #region public properties
 
-        /* TODO: Now that nodes know nothing about their owners, we can't access RunEnabled.
+        public RevitDynamoModel RevitDynamoModel
+        {
+            get
+            {
+                return revitDynamoModel;
+            }
+            set
+            {
+                if (revitDynamoModel != null)
+                {
+                    var hwm = revitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
+                    hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
+                }
+
+                revitDynamoModel = value;
+
+                if (revitDynamoModel != null)
+                {
+                    var hwm = revitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
+                    hwm.RunSettings.PropertyChanged += revMod_PropertyChanged;
+                }
+            }
+        }
+
         public override bool CanSelect
         {
-            get { return base.CanSelect && RevitDynamoModel.RunEnabled; }
+            get
+            {
+                if (revitDynamoModel != null)
+                {
+                    var hwm = RevitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
+                    return base.CanSelect && hwm.RunSettings.RunEnabled;
+                }
+                else
+                {
+                    return base.CanSelect;
+                }
+            }
             set { base.CanSelect = value; }
         }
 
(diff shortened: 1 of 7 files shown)


src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -35,14 +35,48 @@ namespace Dynamo.Nodes
...
-        /* TODO: Now that nodes know nothing about their owners, we can't access RunEnabled.
+        public RevitDynamoModel RevitDynamoModel
+        {
+            get
+            {
+                return revitDynamoModel;
+            }
+            set
+            {
+                if (null != revitDynamoModel)
```
Benglin: I stopped doing this (as I have been advised few times before) quite some time back as it is no longer solving any real problem in C#, and instead I now do `revitDynamoModel != null`. So I decided to also advise you to not do this in the future. No big deal for now though :)

src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -108,6 +148,12 @@ public override void Dispose()
             RevitServicesUpdater.Instance.ElementsDeleted -= Updater_ElementsDeleted;
             RevitServicesUpdater.Instance.ElementsModified -= Updater_ElementsModified;
             DocumentManager.Instance.CurrentUIApplication.Application.DocumentOpened -= Controller_RevitDocumentChanged;
+
+            if (null != revitDynamoModel)
+            {
+                var hwm = RevitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
+                hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
+            }
```
Benglin: What about opening a new file (which will update `HomeWorkspaceModel` inside `DynamoModel`)? In that case we need to unregister from the event. If that is the case, we might want to detect property change of `CurrentSpace` on `DynamoModel`, and subscribe/unsubscribe the handler.

src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -50,9 +84,17 @@ public override string SelectionSuggestion
...
             {
-                return RevitDynamoModel.RunEnabled
-                    ? base.SelectionSuggestion
-                    : "Selection is disabled when Dynamo run is disabled.";
+                if (null != revitDynamoModel)
+                {
+                    var hwm = RevitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
+                    return hwm.RunSettings.RunEnabled
+                        ? base.SelectionSuggestion
+                        : "Selection is disabled when Dynamo run is disabled.";
```
Benglin: `SelectionSuggestion` seems to be user facing, this needs to go into a resource file. Do take note though, the WWL team localizes only resources inside `Revit_2014` and then the localized resources are copied into `Revit_2015` and `Revit_2016`. So it is mandatory that we keep these three branches in-sync, and new strings are added only to `Revit_2014`.

test/Libraries/RevitIntegrationTests/BugTests.cs
```diff
@@ -720,6 +721,28 @@ public void MAGN_7075()
             //There should be no infinite loop, otherwise, there will be an error with this test case.
         }
 
+        [Test]
+        [Category("RegressionTests")]
+        [TestModel(@".\empty.rfa")]
+        public void MAGN_7251()
```
Benglin: There are various feedback before this, is that when anyone who sees this `MAGN_7251` will have no immediate way to find out what it is about (both this test and the file names). A better way from this point onward would be to name them something like `SelectionButtonShouldBeReenabledWhenRunCompletes`. Does that make sense?

src/Libraries/RevitNodesUI/Selection.cs
Randy-Ma: In that case, the old workspace and its RunSettings will be disposed and so do the selection nodes in it. When RunSettings is disposed, I think the event will be automatically unsubscribed. Is that right?

test/Libraries/RevitIntegrationTests/BugTests.cs
Randy-Ma: Maybe SelectionButtonShouldBeReenabledWhenRunCompletes_MAGN_7251?

Benglin: Sounds just as good :)

src/Libraries/RevitNodesUI/Selection.cs
Benglin: Yah okay I got it, as long as it is disposed then we are good.

Benglin:
Thanks a lot for updating this, @Randy-Ma. Now it's crucial that we ensure all the three `*.resx` in `Revit_2014`, `Revit_2015` and `Revit_2016` branches are exactly the same, so please don't forget to ensure that's the case. Thanks again!

Randy-Ma:
@Benglin 
Thanks for the reminder! But as this pull request need https://github.com/DynamoDS/Dynamo/pull/4594 to be reviewed and merged as well, I am temporarily marking it as "DO NOT MERGE"

PR #420, opened 2015-05-29
add pythonTest in Revit2016

zora-wang:
Purpose:

add pythonTest in Revit2016

Reviewer:
@riteshchandawar @Benglin


zora-wang:
This test case is not related to any Revit so we need to add more test in this files later. and we close it now

PR #461, opened 2015-06-08
Fix crash for "All Elements in Active View" node

sharadkjaiswal:
### Purpose
Fixes http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7527
- Revit Document Event handlers were registered outside API context and hence it crashed.
- Fixed by moving the event registration thru Idle action registration.

### Declarations

### Reviewers

- [ ] @ikeough 

### FYIs

@monikaprabhu

### DIFF
--- a/src/Libraries/RevitNodesUI/Elements.cs
+++ b/src/Libraries/RevitNodesUI/Elements.cs
@@ -2,23 +2,18 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-
 using Autodesk.Revit.DB;
 using Autodesk.Revit.DB.Architecture;
 using Autodesk.Revit.DB.Electrical;
 using Autodesk.Revit.DB.Structure;
-
+using Dynamo.Applications;
 using Dynamo.Applications.Models;
 using Dynamo.Models;
-
 using ProtoCore.AST.AssociativeAST;
-
 using Revit.Elements;
 using Revit.Elements.InternalUtilities;
-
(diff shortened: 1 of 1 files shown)


sharadkjaiswal:
same as #460

src/Libraries/RevitNodesUI/Elements.cs
```diff
@@ -215,37 +210,46 @@ public ElementsInView()
...
+
+                    RevitServicesUpdater.Instance.ElementsDeleted -=
+                        RevitServicesUpdaterOnElementsDeleted;
+                    RevitServicesUpdater.Instance.ElementsModified -=
+                        RevitServicesUpdaterOnElementsModified;
+                    RevitServicesUpdater.Instance.ElementsAdded -=
+                        RevitServicesUpdaterOnElementsAdded;
+                });
+
             base.Dispose();
```
Benglin: It should be fine I guess but now these event handlers are unregistered way after `base.Dispose` method is called, if for some reasons these events are raised then it's important to ensure they don't access things that are already disposed of.

Benglin:
Generally it looks good (just have to ensure the event handlers do not do anything even if they are invoked, after the element is disposed. Then LGTM.

src/Libraries/RevitNodesUI/Elements.cs
sharadkjaiswal: Now event handlers are unregistered before the base is disposed.

PR #466, opened 2015-06-12
Revit2016 bugs

nguyen-binh-minh:
- [MAGN-7624](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7624) Category shouldn't be localized
- [MAGN-7544](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7624) Localize RevitDropDown
- Migration of Rectangle.ByWidthHeight
@sharadkjaiswal PTAL

### DIFF
--- a/src/Libraries/Migrations/RevitNodes/Rectangle.cs
+++ b/src/Libraries/Migrations/RevitNodes/Rectangle.cs
@@ -8,7 +8,7 @@ namespace Dynamo.Nodes
         [NodeMigration(from: "0.6.3.0", to: "0.7.0.0")]
         public static NodeMigrationData Migrate_0630_to_0700(NodeMigrationData data)
         {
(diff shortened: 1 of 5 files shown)


src/Libraries/Migrations/RevitNodes/Rectangle.cs
```diff
@@ -9,7 +9,7 @@ public class Rectangle : MigrationNode
         public static NodeMigrationData Migrate_0630_to_0700(NodeMigrationData data)
         {
             return MigrateToDsFunction(data, "ProtoGeometry.dll", "Rectangle.ByWidthHeight",
```
sharadkjaiswal: It would be good to update the nickname as well, because Rectangle.ByWidthHeight node can't be searched in the library.

PR #467, opened 2015-06-12
Revit2014 bugs

nguyen-binh-minh:
- [MAGN-7624](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7624) Category shouldn't be localized
- [MAGN-7544](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7624) Localize RevitDropDown
- Migration of Rectangle.ByWidthHeight
@sharadkjaiswal PTAL


Benglin:
Hi @nguyen-binh-minh, thanks for doing this. One very important thing to note is that all `*.resx` files in `Revit2014` will be localized, and the resulting `*.dll` files are duplicated for `Revit2015` and `Revit2016`. 15 and 16 are not localized separately at all. So please make sure `*.resx` are the same across all three branches (also the RC branches). Thanks!

PR #468, opened 2015-06-12
update GetIdentifierFromModelObject method in Selection.cs

zora-wang:
### Purpose

Updated `GetIdentifierFromModelObject` method to be more robust in handling *missing element* scenario (e.g. when the host document is closed). This is meant to address [an internally tracked defect](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7658).

### Declarations

### Reviewers

Mr . @Benglin 

### FYIs

@ikeough


Benglin:
Thanks a lot for helping with so many branches, @zora-wang!

PR #474, opened 2015-06-12
mark SelectionButtonSholudBeDisableAfterOpeningNewDoc as failure

zora-wang:
Purpose:

 Marking this test case as failure since it is being worked on as MAGN-7656

Reviewer:
@Benglin

### DIFF
--- a/test/Libraries/RevitIntegrationTests/BugTests.cs
+++ b/test/Libraries/RevitIntegrationTests/BugTests.cs
@@ -703,12 +703,13 @@ namespace RevitSystemTests
         }
 
 
-        [Test]
+        [Test, Category("Failure")]
         [Category("RegressionTests")]
         [TestModel(@".\empty.rfa")]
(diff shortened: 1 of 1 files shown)


test/Libraries/RevitIntegrationTests/BugTests.cs
```diff
@@ -703,12 +703,13 @@ public void MAGN_7075()
...
 
-        [Test]
+        [Test, Category("Failure")]
         [Category("RegressionTests")]
         [TestModel(@".\empty.rfa")]
         public void SelectionButtonShouldBeDisabledAfterOpeningNewDocument()
         {
-             string filePath = Path.Combine(workingDirectory, @".\Bugs\MAGN_7251.dyn");
+            // Marking this test case as failure since it is being worked on as MAGN-7656
+            string filePath = Path.Combine(workingDirectory, @".\Bugs\MAGN_7251.dyn");
```
Benglin: Can we move it to the end of `Test, Category` attribute instead of here?

PR #480, opened 2015-06-15
Fix MAGN-7656: select button incorrectly enabled

Benglin:
### Purpose

This pull request is meant to fix the issue where selection buttons incorrectly enabled on a different Revit document. Fundamentally, the problem is cause by the fact that the state of selection button is dependent solely on the value of `RunSettings.RunEnabled`, which may be unstable in various stages of the execution.

The fix introduces a new property `bool RevitDynamoModel.IsInRightDocumentContext` which indicates if the active document is the same document that was bound to the Home workspace. In addition to `RunSettings.RunEnabled`, `RevitSelection.CanSelect` is now made dependent on `RevitDynamoModel.IsInRightDocumentContext`, which means the selection button will never incorrectly be enabled in the wrong document context.

There's also the addition of `DocumentManager.ActiveDocumentHashCode` property, which is always updated to reflect the current active document in Revit. This is necessary because Revit API `UIApplication.ActiveUIDocument` is not able to correctly identify the current active document (most of the time it does not change no matter how user switches the document).

This defect is internally tracked as:

[MAGN-7656](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7656) Selection button is activated in different Revit document

### Declarations

### Reviewers

Hi @ikeough, please have a look at this one and let me know if you spot any issue. In the long run (or even in the near term), we need to make constructs around *session document* more robust and only have very few entry-points for setting them. Thanks!

### FYIs

Hi @Randy-Ma, you might be interested in this if you're online.

### DIFF
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -131,7 +131,7 @@ namespace Dynamo.Applications.Models
             }
         }
 
-        public bool IsInRightDocumentContext
+        public bool IsInMatchingDocumentContext
         {
             get
             {
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -131,6 +131,28 @@ namespace Dynamo.Applications.Models
             }
         }
 
+        public bool IsInRightDocumentContext
+        {
+            get
+            {
+                if (DocumentManager.Instance == null)
+                    return false; // Not initialized yet.
+
+                var dm = DocumentManager.Instance;
+                if (dm.CurrentDBDocument == null)
+                    return false; // There's no current document stored.
+
+                // Selection is not allowed in perspective view mode.
+                var view3D = dm.CurrentUIDocument.ActiveView as View3D;
+                if ((view3D != null) && view3D.IsPerspective)
+                    return false; // There's no view, or in perspective view.
+
+                // Is the cached document the same as the active document?
+                var currentHashCode = dm.CurrentDBDocument.GetHashCode();
+                return currentHashCode == dm.ActiveDocumentHashCode;
+            }
+        }
+
         #endregion
 
         #region Constructors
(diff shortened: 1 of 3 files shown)


src/Libraries/RevitServices/Persistence/DocumentManager.cs
```diff
@@ -106,6 +123,21 @@ public IEnumerable<Element> ElementsOfCategory(BuiltInCategory category)
         }
 
         /// <summary>
+        /// This property represents the hash code for the current active Revit
+        /// Document object. If there is no active document, this value will be 
+        /// set to zero. This property is made an 'int' because it is primarily
+        /// meant to indicate the active document for comparison purposes, no 
+        /// Document specific operations should be allowed on it. This property 
+        /// is updated when a document switch or a document closure happens. The
+        /// reason this property is needed because the following property cannot
```
ikeough: @Benglin Why is this the case? Is the Revit API broken? If so, have we reported this to the Revit team?

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -131,6 +131,28 @@ override internal string AppVersion
             }
         }
 
+        public bool IsInRightDocumentContext
```
ikeough: @Benglin Small name change suggestion. `IsInCorrectDocumentContext` or `IsInValidDocumentContext`.

ikeough:
Thanks @Benglin. I agree that we need a better design around the idea of the "session document". We've been talking for some time about designing an experience so that users can switch their active session document, and your new API will be useful in this.

The Revit API's documentation on `GetHashCode` for a document instance is unlcear:  
>The hash code is the same for document instances that represent the same document currently opened in the Revit session. The hash code is generated when a Revit file is opened or created in session. If the same Revit file is opened later (in the same session or a different session) the hash code will not be the same.

I don't know how one goes about opening two different versions of the same document. The only edge case that I can think of is that someone copied a revit file to another location an attempts to open both the original and the copy at the same time. I would assume the hash code would be different for those two according to the above.

I have one small renaming comment then LGTM unless my above statements cause any reconsideration.

src/Libraries/RevitServices/Persistence/DocumentManager.cs
Benglin: @ikeough, it appears to be, and I have reported to Tau yesterday as I was working on this (it either broke, or did not work reliably for us to function well).

src/DynamoRevit/Models/RevitDynamoModel.cs
Benglin: Thanks man, I suck at names :+1:

Benglin:
What that documentation meant was the hash code can reliably identify a given `DB.Document` object, but this value is generated during runtime (I suspect it is a hash of the underlying unmanaged pointer address) and will remain valid for as long as the `Document` stays in memory. I had a word with Revit developer Tau about this, and it has nothing to do with the `DocumentClosingEventArgs.DocumentId` value (which seems to be a redundant piece of data for `DocumentClosingEventArgs`), I think this fact combined with unclear documentation ultimately resulted in confusion. But thanks anyway for looking!

Benglin:
CI system for Revit tests is currently unstable, I'll wait for that to be clean before merging this in.

Benglin:
Thanks again @ikeough for your time/help. I'm merging this in because it needs more testing.

PR #498, opened 2015-06-18
update Revit file version that are tested in Revit2014

zora-wang:
### Purpose
update Revit file(.rvt, .rfa) that are used in Revit2014

### Reviewer
@riteshchandawar


riteshchandawar:
LGTM!

@ke-yu Can you please merge this. It is just opening revit file in Revit 2014 and saving them so that there will not be any migration while running tests.

Thanks,
Ritesh

PR #506, opened 2015-06-18
update Revit file version that are used in testing in Revit2016

zora-wang:
### Purpose

update Revit file version that are used in testing in Revit2016

### Reviewer:
@riteshchandawar


riteshchandawar:
LGTM!

Merging this in.
This will improve our Revit test time by 2 and half hr., as there will not be any migration now.

PR #511, opened 2015-06-23
Fix MAGN-7679 Loading a second Revit sample hangs Revit/Dynamo after loading and running the first sample

Randy-Ma:
### Purpose

This submission is trying to fix MAGN-7679. The issue is with the selection node. Every time a new file is opened, the old workspace will be removed from the model's workspace list and disposed and so is the selection node. The problem is that the selection node is trying to get the workspace from the model's list of workspaces when it is disposed. But unluckily the list does not contain the workspace because it has been removed.

The fix here is trying to use the CurrentWorkspace property instead of getting the first workspace from the workspace list.

### Declarations

### Reviewers

@Benglin

### DIFF
--- a/src/Libraries/RevitNodesUI/Selection.cs
+++ b/src/Libraries/RevitNodesUI/Selection.cs
@@ -49,16 +49,22 @@ namespace Dynamo.Nodes
             {
                 if (revitDynamoModel != null)
                 {
-                    var hwm = revitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
-                    hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
+                    var hwm = revitDynamoModel.CurrentWorkspace as HomeWorkspaceModel;
+                    if (hwm != null)
+                    {
+                        hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
+                    }
                 }
 
                 revitDynamoModel = value;
 
                 if (revitDynamoModel != null)
                 {
-                    var hwm = revitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
-                    hwm.RunSettings.PropertyChanged += revMod_PropertyChanged;
+                    var hwm = revitDynamoModel.CurrentWorkspace as HomeWorkspaceModel;
+                    if (hwm != null)
+                    {
+                        hwm.RunSettings.PropertyChanged += revMod_PropertyChanged;
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -49,16 +49,22 @@ public RevitDynamoModel RevitDynamoModel
             {
                 if (revitDynamoModel != null)
                 {
-                    var hwm = revitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
-                    hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
+                    var hwm = revitDynamoModel.CurrentWorkspace as HomeWorkspaceModel;
+                    if (hwm != null)
+                    {
+                        hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
```
aparajit-pratap: Just curious why this is not being de-registered at the time of removing the workspace?

src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -155,8 +175,11 @@ public override void Dispose()
 
             if (revitDynamoModel != null)
             {
-                var hwm = RevitDynamoModel.Workspaces.OfType<HomeWorkspaceModel>().ElementAt(0);
-                hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
+                var hwm = revitDynamoModel.CurrentWorkspace as HomeWorkspaceModel;
+                if (hwm != null)
+                {
+                    hwm.RunSettings.PropertyChanged -= revMod_PropertyChanged;
```
aparajit-pratap: I see that the de-registration is done here. Then why is it needed again in `RevitDynamoModel` setter?

src/Libraries/RevitNodesUI/Selection.cs
Randy-Ma: @aparajit-pratap 
Thanks for the comment!

If registrition or unregistrition happens in the workspace, that means every time when a file is opened, we need to check whether there are selection nodes, if so, register the event. This is not necessary for files which do not contain selection nodes. Also if selection nodes are added/removed (users are free to edit a file by adding or removing nodes), the event handler need to be registered/unregistered. At this time, the workspace is open and it will not be removed when adding/removing nodes happens.

src/Libraries/RevitNodesUI/Selection.cs
Randy-Ma: Here is when the selection node is disposed.

The one in the setter is to un-register the old event handler for the old model and register the event handler for the new model.

aparajit-pratap:
Thanks @Randy-Ma. LGTM!

Benglin:
Thanks @Randy-Ma and @aparajit-pratap , as discussed earlier this only makes the existing code more error-proof, so LGTM.

PR #520, opened 2015-06-30
Fix MAGN-7641 All Elements in Active View : result changes after zoom out and save zoom and save

Randy-Ma:
### Purpose

This pull request is to fix MAGN-7641. The cause of this defect is when an element is added to the active view, it will be added as an element of the "All Elements in Active View" node no matter the element is visible or not.

The fix here is to filter the element using a filter constructed by an existing function.

### Declarations

### Reviewers
@sharadkjaiswal

### DIFF
--- a/test/Libraries/RevitIntegrationTests/BugTests.cs
+++ b/test/Libraries/RevitIntegrationTests/BugTests.cs
@@ -763,6 +763,24 @@ namespace RevitSystemTests
             RunCurrentModel();
         }
 
+        [Test]
+        [Category("RegressionTests")]
+        [TestModel(@".\empty.rfa")]
+        public void AllElementsInActiveViewReturnsViewableElements()
+        {
+            string filePath = Path.Combine(workingDirectory, @".\Bugs\MAGN_7641_simplified.dyn");
+            string testPath = Path.GetFullPath(filePath);
+
+            ViewModel.OpenCommand.Execute(testPath);
+            AssertNoDummyNodes();
+            RunCurrentModel();
+            RunCurrentModel();
+
+            var elements = GetPreviewCollection("55a73e51-1021-44e4-aacd-a4222ca2ba25");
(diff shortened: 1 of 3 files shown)


test/Libraries/RevitIntegrationTests/BugTests.cs
```diff
@@ -763,6 +763,24 @@ public void CanOpenAndRunOtherFilesAfterOpeningFileWithSelectNode()
...
+        [Category("RegressionTests")]
+        [TestModel(@".\empty.rfa")]
+        public void AllElementsInActiveViewReturnsViewableElements()
+        {
+            string filePath = Path.Combine(workingDirectory, @".\Bugs\MAGN_7641_simplified.dyn");
+            string testPath = Path.GetFullPath(filePath);
+
+            ViewModel.OpenCommand.Execute(testPath);
+            AssertNoDummyNodes();
+            RunCurrentModel();
```
sharadkjaiswal: Can we add Revit view manipulation to simulate the exact user scenario here before running it the second time?

Randy-Ma: @sharadkjaiswal 
There is a limitation here!
Right now the test case is running in the background thread. In the process, we have no chance to manipulate the views.

sharadkjaiswal: In that case do we have any effect of calling RunCurrentModel() twice, without any view changes?

Randy-Ma: Yes, the actual cause of this defect is not quite related to view operations.

That is why I have simplified the original case. The first run is to create an import instance which is intentionally created at the center of the view. After the import instance is created, a lot of elements which are not visible will be created. The second run is to sure the node to be tested will not return these invisible elements.

PR #521, opened 2015-07-01
Fix the defect that AsfamilyInstance misses discription

zora-wang:
### Purpose

Fix the defect that "AsFamilyInstance" node missed its node description in Revit2016

### Reviewer
@ke-yu 

Thanks

### DIFF
--- a/src/Libraries/RevitNodes/Elements/CurtainPanel.cs
+++ b/src/Libraries/RevitNodes/Elements/CurtainPanel.cs
@@ -460,7 +460,10 @@ namespace Revit.Elements
          }
          return result.ToArray();
       }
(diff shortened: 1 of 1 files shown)


src/Libraries/RevitNodes/Elements/CurtainPanel.cs
```diff
@@ -460,7 +460,10 @@ public Mullion[] SupportingMullions()
          }
          return result.ToArray();
       }
-
+      /// <summary>
+      /// Gets family instance from curtainPanel
```
ke-yu: "curtainPanel" -> "curtain panel"

PR #524, opened 2015-07-01
fix document description when open file in revit

zora-wang:
### Purpose
Cross-merge fix document description when open file in revit #522 

### Reviewer
@ke-yu


src/DynamoRevit/ViewModel/DynamoRevitViewModel.cs
```diff
@@ -97,9 +97,9 @@ private static string GetDocumentPointerMessage()
             var docPath = DocumentManager.Instance.CurrentUIDocument.Document.PathName;
             var message = String.IsNullOrEmpty(docPath)
                 ? "a new document."
```
ke-yu: "a new document." should be localized.

src/DynamoRevit/ViewModel/DynamoRevitViewModel.cs
```diff
@@ -97,9 +97,9 @@ private static string GetDocumentPointerMessage()
...
-                : String.Format("document: {0}", docPath);
+                : String.Format( Resources.Document+ " {0}", docPath);
```
ke-yu: Why not use string concatenation?

ke-yu: Resource.Document + " : " + docPath

PR #527, opened 2015-07-03
update .dll files in extern folder

zora-wang:
### Purpose
fix the node description : the dll files need to be updated in DynamoRevit also.
the related PR is here: both of PR need to be merged https://github.com/DynamoDS/DynamoRaaS/pull/9

### Reviewer
@ke-yu


extern/DynamoRaaS/App.config
```diff
@@ -22,7 +22,7 @@
     <httpRuntime executionTimeout="3600" maxRequestLength="1024000" useFullyQualifiedRedirectUrl="false" minFreeThreads="8" minLocalRequestFreeThreads="4" appRequestQueueLimit="100"/>
   </system.web>
   <startup>
-    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5"/>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0"/>
```
ke-yu: It is intended?

zora-wang: It has changed in my machine. But i talked with Randy and He said to it is better to keep the origin version. Later i will revert version to 4.5 . Thanks for the notation

PR #530, opened 2015-07-06
localize floor type in resource file

zora-wang:
### purpose
localize term "no floor type available" into the resource file.

### Reviewer
@ke-yu


src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -302,7 +302,7 @@ protected override void DeserializeCore(XmlElement nodeElement, SaveContext cont
     [IsDesignScriptCompatible]
     public class FloorTypes : RevitDropDownBase
     {
-        private const string noFloorTypes = "No floor types available.";
+        private string noFloorTypes = Properties.Resources.NoFloorTypesAvailable;
```
ke-yu: Replace `noFloorTypes` with resource string, it is not necessary to keep this field there.

PR #534, opened 2015-07-07
Fix MAGN-5870 GetParamterValueByName node returns different value each time for Sheet Number

Randy-Ma:
### Purpose

This submission is to fix MAGN-5870. Right now the GetParamterValueByName node may return different value each time it is run. This is because an element may have multiple parameters of the same name. We are trying to only get the first one. But the first one is not fixed.

The fix here is to use a different API to get the user visible parameters so that we will not get multiple parameters of the same name.

### Declarations

### Reviewers

@aparajit-pratap 

### FYIs
@ikeough

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -310,7 +310,7 @@ namespace Revit.Elements
         {
             object result;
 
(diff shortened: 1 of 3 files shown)


aparajit-pratap:
LGTM!

PR #540, opened 2015-07-21
Add event handlers through a proxy for view or document events

Randy-Ma:
### Purpose

The main purpose of this pull request is to avoid direct reference to the subscriber of several events: DocumentOpened, DocumentClosing, DocumentClosed, ViewActivating and ViewActivated.

To achieve this, a proxy class has been implemented which will contain six events and six member functions. The methods of the proxy object will be subscribed to the application events. The original subscriber will subscribe to the events of the proxy. The implementation of the six methods in the proxy will invoke the corresponding events.

Additionally OnApplicationIdle has been implemented which will be registered to the Idling event of the application. An public function: AddIdleAction is created so that it is possible to add some tasks which need to be run when the application is in an idle state.

### Declarations

### Reviewers
@ke-yu PTAL

### FYIs
@ikeough

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -35,7 +35,6 @@ using RevitServices.Threading;
 
 using MessageBox = System.Windows.Forms.MessageBox;
 using DynUpdateManager = Dynamo.UpdateManager.UpdateManager;
-using System.Collections.Generic;
 using Microsoft.Win32;
 
 
@@ -227,8 +226,7 @@ namespace Dynamo.Applications
 
             revitDynamoModel.ShutdownStarted += (drm) =>
             {
-                var uiApplication = DocumentManager.Instance.CurrentUIApplication;
-                uiApplication.Idling += DeleteKeeperElementOnce;
+                DynamoRevitApp.AddIdleAction(DeleteKeeperElement);
             };
 
             return viewModel;
--- a/src/DynamoRevit/DynamoRevitApp.cs
+++ b/src/DynamoRevit/DynamoRevitApp.cs
@@ -2,25 +2,25 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
+using System.Linq;
 using System.IO;
 using System.Reflection;
 using System.Resources;
 using System.Windows;
 using System.Windows.Interop;
 using System.Windows.Media.Imaging;
-
 using Autodesk.Revit.ApplicationServices;
 using Autodesk.Revit.Attributes;
 using Autodesk.Revit.DB;
 using Autodesk.Revit.UI;
-
+using Autodesk.Revit.UI.Events;
 using Dynamo.Applications.Properties;
-
 using RevitServices.Elements;
 using RevitServices.Persistence;
 using RevitServices.Transactions;
-
 using MessageBox = System.Windows.Forms.MessageBox;
+using Dynamo.Models;
+using RevitServices.EventHandler;
 
 namespace Dynamo.Applications
 {
@@ -33,16 +33,21 @@ namespace Dynamo.Applications
         private static readonly string assemblyName = Assembly.GetExecutingAssembly().Location;
         private static ResourceManager res;
         public static ControlledApplication ControlledApplication;
+        public static UIControlledApplication UIControlledApplication;
         public static List<IUpdater> Updaters = new List<IUpdater>();
         internal static PushButton DynamoButton;
+        private static readonly Queue<Action> idleActionQueue = new Queue<Action>(10);
+        private static EventHandlerProxy proxy;
 
         public Result OnStartup(UIControlledApplication application)
         {
             try
             {
-                SubscribeAssemblyResolvingEvent();
-
+                UIControlledApplication = application;
                 ControlledApplication = application.ControlledApplication;
+
+                SubscribeAssemblyResolvingEvent();
+                SubscribeApplicationEvents();
 
                 TransactionManager.SetupManager(new AutomaticTransactionStrategy());
                 ElementBinder.IsEnabled = true;
@@ -89,8 +94,51 @@ namespace Dynamo.Applications
         public Result OnShutdown(UIControlledApplication application)
         {
             UnsubscribeAssemblyResolvingEvent();
+            UnsubscribeApplicationEvents();
 
             return Result.Succeeded;
+        }
+
+        private static void OnApplicationIdle(object sender, IdlingEventArgs e)
+        {
+            if (!idleActionQueue.Any())
+                return;
+
+            Action pendingAction = null;
+            lock (idleActionQueue)
+            {
+                pendingAction = idleActionQueue.Dequeue();
+            }
+
+            if (pendingAction != null)
+                pendingAction();
+        }
+
+
+        /// <summary>
+        /// Add an action to run when the application is in the idle state
+        /// </summary>
+        /// <param name="a"></param>
+        public static void AddIdleAction(Action a)
+        {
+            // If we are running in test mode, invoke 
+            // the action immediately.
+            if (DynamoModel.IsTestMode)
+            {
+                a.Invoke();
+            }
+            else
+            {
+                lock (idleActionQueue)
+                {
+                    idleActionQueue.Enqueue(a);
+                }
+            }
+        }
+
+        public static EventHandlerProxy EventHandlerProxy
+        {
+            get { return proxy; }
         }
 
         // should be handled by the ModelUpdater class. But there are some
@@ -115,6 +163,34 @@ namespace Dynamo.Applications
                 filter,
                 Element.GetChangeTypeAny());
             Updaters.Add(sunUpdater);
+        }
+
+        private void SubscribeApplicationEvents()
+        {
+            UIControlledApplication.Idling += OnApplicationIdle;
+
+            proxy = new EventHandlerProxy();
+
+            UIControlledApplication.ViewActivated += proxy.OnApplicationViewActivated;
+            UIControlledApplication.ViewActivating += proxy.OnApplicationViewActivating;
+
+            ControlledApplication.DocumentClosing += proxy.OnApplicationDocumentClosing;
+            ControlledApplication.DocumentClosed += proxy.OnApplicationDocumentClosed;
+            ControlledApplication.DocumentOpened += proxy.OnApplicationDocumentOpened;
+        }
+
+        private void UnsubscribeApplicationEvents()
+        {
+            UIControlledApplication.Idling -= OnApplicationIdle;
+
+            UIControlledApplication.ViewActivated -= proxy.OnApplicationViewActivated;
+            UIControlledApplication.ViewActivating -= proxy.OnApplicationViewActivating;
+
+            ControlledApplication.DocumentClosing -= proxy.OnApplicationDocumentClosing;
+            ControlledApplication.DocumentClosed -= proxy.OnApplicationDocumentClosed;
+            ControlledApplication.DocumentOpened -= proxy.OnApplicationDocumentOpened;
+
+            proxy = null;
         }
 
         private void SubscribeAssemblyResolvingEvent()
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -7,33 +7,25 @@ using System.Reflection;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Soap;
 using System.Windows.Forms;
-
 using Autodesk.Revit.DB;
 using Autodesk.Revit.DB.Events;
 using Autodesk.Revit.UI;
 using Autodesk.Revit.UI.Events;
-
 using DSIronPython;
-
 using Dynamo.DSEngine;
 using Dynamo.Interfaces;
 using Dynamo.Models;
 using Dynamo.UpdateManager;
 using Dynamo.Utilities;
-
 using DynamoServices;
-
 using Greg;
-
 using ProtoCore;
-
 using Revit.Elements;
-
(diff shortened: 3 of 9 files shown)


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -134,6 +137,22 @@ public Result Execute(ExternalCommandData commandData, ref string message, Eleme
             return Result.Succeeded;
         }
 
+        private static void OnApplicationIdle(object sender, IdlingEventArgs e)
```
ke-yu: Why we need `OnApplicationIdle` instead of adding a task to scheduler?

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -134,6 +137,22 @@ public Result Execute(ExternalCommandData commandData, ref string message, Eleme
...
+        {
+            if (idleActions.Count == 0)
+                return;
+
+            Action pendingAction = null;
+            lock (idleActions)
+            {
+                pendingAction = idleActions[0];
+                idleActions.RemoveAt(0);
```
ke-yu: We could use `Queue`

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -417,6 +435,26 @@ private static void DeleteKeeperElement()
...
 
         #endregion
+
+        /// <summary>
+        /// Add an action to run when the application is in the idle state
+        /// </summary>
+        public static void AddIdleAction(Action a)
+        {
+            // If we are running in test mode, invoke 
+            // the action immediately.
```
ke-yu: Code is self-explanatory, comment is unnecessary :-)

src/Libraries/RevitServices/Events/EventHandlerProxy.cs
```diff
@@ -0,0 +1,175 @@
...
+                        AddDictItem(type, m);
+                        m.Invoke(h, new[] { sender, args });
+                    }
+                }
+            }
+        }
+
+        protected void AddDictItem(EventType type, MethodInfo method)
+        {
+            if (_dict.ContainsKey(type))
```
ke-yu: This if-else is equivalent to `_dict[type] = method`.

src/DynamoRevit/DynamoRevit.cs
Randy-Ma: Adding a task to a scheduler need that the scheduler is initialized and the scheduler is not disposed when the operation happens. It has been done for fixing one defect previously:
https://github.com/DynamoDS/Dynamo/pull/3531

For your suggestion, I am thinking whether I can use the existing ExecuteOnIdleAsync for cases which happens when a valid scheduler is available. But I do not want to make my part of change depend on the RevitDynamoModel instance.

The point is that I am just subscribing/unsubscribing an event, why do I need to make it depend on more objects?

src/Libraries/RevitServices/Events/EventHandlerProxy.cs
Randy-Ma: Nice catch!

src/DynamoRevit/DynamoRevit.cs
Randy-Ma: But at least it will have no harm.

ke-yu:
@Randy-Ma , correct me if I'm wrong. 

Previously, `DynamoRevitModel` is not released because it could be referenced by `Application`'s event handler.

Now we have proxy object, and proxy object is referenced by `Application`'s event handler. But as `DynamoRevitModel` owns proxy object, if proxy object is not GCed, `DynamoRevitModel` won't be GCed as well. 

Do I miss something?

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -115,6 +116,8 @@ public Result Execute(ExternalCommandData commandData, ref string message, Eleme
 
                 TryOpenWorkspaceInCommandData(extCommandData);
 
+                extCommandData.Application.Idling += OnApplicationIdle;
```
ke-yu: Seems not unregister from `Idling`?

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -355,12 +349,16 @@ private void SubscribeApplicationEvents(ExternalCommandData commandData)
                 return;
             }
 
-            commandData.Application.ViewActivating += OnApplicationViewActivating;
-            commandData.Application.ViewActivated += OnApplicationViewActivated;
+            proxy = new EventHandlerProxy<RevitDynamoModel>(this);
```
ke-yu: Can we directly add event handler instead of adding in OnIdle()?

src/Libraries/RevitServices/Events/EventHandlerProxy.cs
```diff
@@ -0,0 +1,168 @@
...
+            if (h != null)
+            {
+                MethodInfo m;
+                if (_dict.TryGetValue(type, out m) && m != null)
+                {
+                    m.Invoke(h, new[] { sender, args });
+                }
+                else
+                {
+                    var t = h.GetType();
```
ke-yu: 1) why we need dictionary? 
2) In which case `else` branch will be hit?

src/DynamoRevit/Models/RevitDynamoModel.cs
Randy-Ma: It  has no issues if SubscribeApplicationEvents happens in the scope of the execute method of the command. But I think it is safer to put it in an idling event.

src/Libraries/RevitServices/Events/EventHandlerProxy.cs
Randy-Ma: For some cases, the proxy will only be used for 1 or 2 types of events, then I will add the delegate's methodinfo to the dictionary. For RevitDynamoModel, all 6 cases are used, so the methodinfo will be found through reflection.

Randy-Ma:
As per discussed, the proxy will be set to null after the event is unsubscribed, and the proxy is only holding a weak reference to the host. So the host will be GC'ed and my test also identifies the host is GC'ed.

ke-yu:
@Randy-Ma , as now `EventHandlerProxy` doesn't have a weak reference to `DynamoReivtModel`, the PR description probably should be updated.

@ikeough , could you take a look as well? It is to fix memory leak on Revit. Although memory leak is still there (`EventHandlerProxy` instance), but as the size of `EventHandlerProxy` is small, memory leak won't be noticeable.

Randy-Ma:
Thanks! Updated

src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -117,6 +165,34 @@ private static void RegisterAdditionalUpdaters(UIControlledApplication applicati
...
+            UIControlledApplication.Idling -= OnApplicationIdle;
+
+            UIControlledApplication.ViewActivated -= proxy.OnApplicationViewActivated;
+            UIControlledApplication.ViewActivating -= proxy.OnApplicationViewActivating;
+
+            ControlledApplication.DocumentClosing -= proxy.OnApplicationDocumentClosing;
+            ControlledApplication.DocumentClosed -= proxy.OnApplicationDocumentClosed;
+            ControlledApplication.DocumentOpened -= proxy.OnApplicationDocumentOpened;
+
+            proxy = null;
```
ke-yu: As `proxy`'s lifetime is as long as application, not need to set it to null. 

Question: why adding event handler is not necessary in onIdle()?

Randy-Ma: The proxy's lifetime is as long as the addin application, but not the Revit application.

Related to adding event handlers in the idling event handlers, it is not necessary to add event handlers for our own events in the idling event handler. I will make the change.

Randy-Ma:
@ke-yu 
Thanks for the great comments! I have made some changes.

src/Libraries/RevitServices/Events/EventHandlerProxy.cs
```diff
@@ -0,0 +1,57 @@
...
+using Autodesk.Revit.UI.Events;
+using Dynamo.Models;
+
+namespace RevitServices.EventHandler
+{
+    /// <summary>
+    /// This is a event handler proxy class to serve as a proxy between the event publisher and
+    /// the event subscriber
+    /// </summary>
+    public class EventHandlerProxy
```
ke-yu: Can we make `EventHandlerProxy` be a inner class of application? I don't think `EventHandlerProxy` is for public use, and probably should be sealed.

Randy-Ma: Well, it is more like a service which depends only on Revit APIs and can be reused. What's more, a modularized implementation is better than putting everything in one place.

If deriving from one class and will cause issues, perhaps we need to make it not inheritable. But in this case, inheriting from this class won't cause any issues. Why not make it more open?

ke-yu: IMHO, 

1) as this class is a proxy for application level event, it should have the same lifetime as the host application, otherwise if it is released somewhere, the subscriber won't be able to notice

2) it is sort of a workaround for Revit event handler memory leak. In the code wherever we use `EventHandlerProxy`, it is actually from `DynamoRevitApplication.EventHandlerProxy`. 

We don't encourage to use `EventHandlerProxy` for all cases except those Revit application events which will cause memory leak, so making it be an inner class makes more sense.

Randy-Ma: @ke-yu 
Thanks for the comments!

But a class won't have a lifetime. I assume you are referring to the class instance. But when creating and releasing the instance(by setting it to null), we have ensured that the lifetime of this instance is within the lifetime of the addin application, that is also why the instance is member of the addin application. As for your worry about that it can be released somewhere, there is not a set method for the EventHandlerProxy property and so it is impossible to set the proxy to null outside the addin application class.

The reason to put EventHandlerProxy to a separate module is for modulization so that the code is more easier to read. Anyway putting EventHandlerProxy as an inner class can not prevent it from being used, we have to make the EventHandlerProxy as public and the class to be public as well. Then the class can still be used like:
var proxy = new DynamoRevitApp.EventHandlerProxy();

ke-yu:
@Randy-Ma other classes no need to know `EventHandlerProxy`...

Can we do in this way:
```
class DynamoRevitApplication
{
    private class EventHandlerProxy
    {
        WeakReference _app;
        
        EventHandlerProxy(DynamoRevitApplication app)
        {
            _app = app;
            _app.ConotrolApplication.DocumenOpened += OnDocumentOpen;
           ...
        }

        private void OnDocumentOpened(...)
        {
          // pass it back
           _app.DocumentOpened(...);
        }
        ...
    }

    private eventHandlerProxy = new EventHandlerProxy(this);

    ...
    public event EventHandler<DocumentOpenedEventArgs> DocumentOpened;
    ...
}

...
DynamoRevitApplication.DocumentOpened += revitModel.OnDocumentOpened(...);
...
```

Randy-Ma:
@ke-yu 
Thanks for various ideas for possible improvements! Some of them are very good ideas and some of them make me think about the implementation more and actually learned a lot in the process.

For OOP, if we have some functions which are quite related to each other, why not put them in one class? Even if you make a class as a private type of another class, it still can be accessed through reflection. You can not hide DynamoRevitApplication but I think no one will ever try to create an instance of DynamoRevitApplication in the addin.

If you really do not want modulization, it will be much simpler by not even defining the EventHandlerProxy class, you can move all the members of the EventHandlerProxy class as members of the DynamoRevitApplication class.

In your suggested implementation, _app may be null and _app.DocumentOpened may be null. The way to call DocumentOpened may cause race condition.

If we wrap all related functions in one class, it will be easier for error preventing, reading and maintaining.

ikeough:
@ke-yu @Randy-Ma This is great conversation, and a tricky problem! There is one other potential solution, if I am understanding the problem correctly. Create a static member `DynamoRevitApp.Proxy`. Then use the `OnStartup` and `OnShutdown` methods of the `IExternalApplication` interface to register and unregister your handlers on your static proxy object. In `OnStartup` you have access to both the `UIApplication` and the `ControlledApplication` required to register those events. Unregister the event handlers in `OnShutdown`. The life of your proxy will start at startup and end at shutdown and you can pick how much access you want developers to have to that member.

Randy-Ma:
@ikeough 
Thanks a lot for looking into this issue!
Your suggestion is exactly what has been implemented for now:
1). There is a static member called proxy in DynamoRevitApp;
2). The handlers are registered in the OnStartup and OnShutdown methods and there are two new methods which will be called: SubscribeApplicationEvents and UnsubscribeApplicationEvents;
3). We pretty much have control of the lifetime of the proxy object.

ke-yu:
@Randy-Ma , definitely we will put everything in `EventHandlerProxy`, but we follow the least privilege principle. My point is not to expose this one and keep it be private in `DynamoRevitApplication` because no one else will use this one (and shouldn't use it) outside of `DynamoRevitApplication`. 

We don't want people to think that any new application level event handler should be added to `EventHandlerProxy` -- it is just a workaround instead of a better pattern. 

But anyway this PR LGTM.

ikeough:
@Randy-Ma Ha! That's what I get for reviewing PRs late at night :) At least we are on the same page. I agree with @ke-yu on the restriction of access to the proxy class.

Randy-Ma:
@ikeough 
If following the same logic, we might need to move the class or some functions of DocumentManager, RevitSchedulerThread, or some others out as well. IMHO, we have to make a balance to achieve modulization. The RevitProxyHandler class only depends on Revit APIs and that is why I think it is better to make it stay at a lower level. Assuming when you are reviewing some code changes in DynamoRevitApp, you will never bother to see more code in RevitProxyHandler, and when you are merging the code and you will not bother to solve possible conflictions related to it, will that be nice? I will merge this first and I think we can continue discuss on this and make another submission to address this one.

PR #543, opened 2015-07-29
Use xcopy to copy a file instead of copy

Randy-Ma:
### Purpose

This is to fix an issue in the script for the PostBuildEvent. The issue is that the copy will fail if the target directory does not exist. This fix uses xcopy instead of copy.

### Declarations

### Reviewers

@riteshchandawar PTAL


riteshchandawar:
LGTM!

Randy-Ma:
@riteshchandawar 
Thanks a lot!

PR #546, opened 2015-07-29
Fixes Version Selector UI defect MAGN-7250

sharadkjaiswal:
### Purpose

This PR fixes https://github.com/DynamoDS/Dynamo/issues/4194, where Dynamo version selector and visual programming panel at far apart.

- When a user selects the specific Dynamo version using version selector, we no longer need the version selector UI in the current session of Revit, and hence it was disabled before. 

- This PR now hides the Ribbon Panel for the version selector so that we don't show two Dynamo icon on Ribbon.

- For Dynamo 0.8.2 onwards, the Dynamo Versions panel is renamed to "Visual Programming" so we show the Dynamo button at the same location. 

- We can't do as above for the older version of Dynamo because the previous version of Dynamo always creates a new panel named with "Visual Programming". And if this panel already exists, it throws an error. 0.8.2 onwards DynamoRevitApp checks for any available panel with this name and tries to add the Dynamo button to existing panel.

#### When Revit is launched:
image

#### If Dynamo 0.8.2 is launched 
image

#### If pre 0.8.2 Dynamo is launched
image

### Declarations

### Reviewers

@Randy-Ma

### DIFF
--- a/src/DynamoRevit/DynamoRevitApp.cs
+++ b/src/DynamoRevit/DynamoRevitApp.cs
@@ -53,8 +53,10 @@ namespace Dynamo.Applications
                 ElementBinder.IsEnabled = true;
 
                 // Create new ribbon panel
-                RibbonPanel ribbonPanel =
-                    application.CreateRibbonPanel(Resources.App_Description);
+                var panels = application.GetRibbonPanels();
+                var ribbonPanel = panels.FirstOrDefault(p => p.Name.Contains(Resources.App_Description));
+                if(null == ribbonPanel)
+                    ribbonPanel = application.CreateRibbonPanel(Resources.App_Description);
 
                 var fvi = FileVersionInfo.GetVersionInfo(assemblyName);
                 var dynVersion = String.Format(Resources.App_Name, fvi.FileMajorPart, fvi.FileMinorPart);
--- a/src/Legacy/DynamoRevitVersionSelector/Properties/Resources.en-US.resx
+++ b/src/Legacy/DynamoRevitVersionSelector/Properties/Resources.en-US.resx
@@ -133,4 +133,7 @@
   <data name="DynamoVersionTooltip" xml:space="preserve">
     <value>Launches Dynamo {0}, to launch a different version of Dynamo you may need to restart Revit</value>
   </data>
+  <data name="VisualProgramming" xml:space="preserve">
+    <value>Visual Programming</value>
+  </data>
 </root>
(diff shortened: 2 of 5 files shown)


src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -53,8 +53,10 @@ public Result OnStartup(UIControlledApplication application)
                 ElementBinder.IsEnabled = true;
 
                 // Create new ribbon panel
-                RibbonPanel ribbonPanel =
-                    application.CreateRibbonPanel(Resources.App_Description);
+                var panels = application.GetRibbonPanels();
+                var ribbonPanel = panels.FirstOrDefault(p => p.Name.Contains(Resources.App_Description));
+                if(null == ribbonPanel)
+                    ribbonPanel = application.CreateRibbonPanel(Resources.App_Description);
```
Randy-Ma: Shall we make the same check for creating the panel in DynamoRevit?

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -124,24 +121,52 @@ public static bool LaunchDynamoCommand(int index, ExternalCommandData commandDat
...
+                    .Invoke(revitCmd, new object[] { commandData, message, null });
+
+                uiApplication = null; //release application, no more needed.
+                return true;
+            }
+            catch (Exception)
+            {
+                splitButton.Enabled = true; //Disable the split button, no more needed.
+                splitButton.Visible = true; //Hide it from the UI
+                ribbonPanel.Visible = true; //Hide the panel
```
Randy-Ma: It looks like the comments are not correct for the three lines.

src/Legacy/DynamoRevitVersionSelector/Properties/Resources.en-US.resx
```diff
@@ -133,4 +133,7 @@
   <data name="DynamoVersionTooltip" xml:space="preserve">
     <value>Launches Dynamo {0}, to launch a different version of Dynamo you may need to restart Revit</value>
   </data>
+  <data name="VisualProgramming" xml:space="preserve">
```
Randy-Ma: Shall we use the existing resource item named as App_Description?

sharadkjaiswal: App_Description is defined in DynamoRevit project, and that dll doesn't get loaded until DynamoRevitVersionSelector is loaded. Also, DynamoRevit shouldn't depend on version selector, so there we can't used resource defined here.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
sharadkjaiswal: good catch, let me correct it.

Randy-Ma:
LGTM

PR #549, opened 2015-08-06
Fix MAGN-7978 Element.Geometry does not find geometry for import instances

Randy-Ma:
### Purpose

This submission is to fix MAGN-7978. Originally when a solid is converted to a proto-solid, all faces will be converted to surfaces and then these surfaces will be joined together to get a solid. But unfortunately there are cases that the original solid might just contain few faces, or few edges and they can not be converted and joined to create a solid.

For the following case, the solid can not converted to one solid by joining.
image

This submission fixes the issue by converting the solid to surfaces and curves, if it fails convert it to one solid by joining.

### Declarations

### Reviewers
@pboyer


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -462,17 +462,19 @@ public object[] Geometry()
...
-                catch (Exception)
+                else
                 {
-                    // we catch all geometry conversion exceptions
+                    var solid = geometryObject as Autodesk.Revit.DB.Solid;
+                    if (solid != null)
+                    {
+                        var geomObjs = solid.ConvertToMany();
+                        converted.AddRange(geomObjs.Where(x => { return x != null; }));
+                    }
```
pboyer: I put the try catch inside of the foreach so that you'd get as much geometry out as is possible. This isn't ideal obviously, but it gives users a "best effort" solution. Do you disagree with that?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -462,17 +462,19 @@ public object[] Geometry()
 
             foreach (var geometryObject in InternalGeometry())
```
pboyer: Is it possible that `InternalGeometry()` would return a null element?

pboyer:
Just some minor comments - LGTM.

src/Libraries/RevitNodes/Elements/Element.cs
Randy-Ma: No, it will return an empty list if there are no geometries.

Randy-Ma:
@pboyer Thanks for reviewing the code!

PR #555, opened 2015-08-12
Not to clean up elements when open a new .dyn file

ke-yu:
### Purpose

This PR is to fix defect [MAGN Opening a new file while running auto deletes all Revit elements made with previous file in same Dynamo session](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7229)). 

In `Element.Dispose()` we not only check if Dynamo is shutting down, but also check if it is closing workspace or not (`DisposeLogic.IsClosingHomeworkspace`)

Test case `OpenNewFileNotCleanupOldElements` is added to test elements won't be cleaned up when open the other .dyn file or clear current workspace.

Run RevitSystemTests.BugTests and RevitSystemTests.ElementBindingTests, no regression introduced.

### Declarations

### Reviewers

@ikeough please take a look.


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -193,8 +194,9 @@ protected ElementId InternalElementId
         public virtual void Dispose()
         {
 
-            // Do not cleanup Revit elements if we are shutting down Dynamo.
-            if (DisposeLogic.IsShuttingDown)
+            // Do not cleanup Revit elements if we are shutting down Dynamo or
+            // closing homeworkspace.
+            if (DisposeLogic.IsShuttingDown || DisposeLogic.IsClosingHomeworkspace)
```
Randy-Ma: Although originally it is returning directly here, but I think it will be better if we unregister this element from the ElementIDLifecycleManager instance.

src/Libraries/RevitServices/RevitServices.csproj
```diff
@@ -78,6 +78,7 @@ limitations under the License.
     <Compile Include="$(SolutionDir)AssemblySharedInfoGenerator\AssemblySharedInfo.cs">
       <Link>Properties\AssemblySharedInfo.cs</Link>
     </Compile>
+    <Compile Include="Elements\DisposeLogic.cs" />
```
Randy-Ma: Where is this file?

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -191,6 +207,8 @@ public new static RevitDynamoModel Start(IRevitStartConfiguration configuration)
         private RevitDynamoModel(IRevitStartConfiguration configuration) :
             base(configuration)
         {
+            DisposeLogic.IsShuttingDown = false;
```
Randy-Ma: If IsClosingHomeworkspace is implemented the same way as IsShuttingDown, shall we initialize it here as well?

ke-yu: Not necessary, by default it is `false`. We need to reset `IsShuttingdown` here because it is the only place to reset `IsShuttingDown` back.

src/Libraries/RevitNodes/Elements/Element.cs
ke-yu: Will that unconnect from element binder? When open the previous document, we still need to connect elements to node.

src/Libraries/RevitServices/RevitServices.csproj
ke-yu: Thanks. Added.

src/Libraries/RevitNodes/Elements/Element.cs
Randy-Ma: No, this is not quite related to element binding.

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -136,6 +136,22 @@ private void OnInvalidRevitDocumentActivated()
             if (handler != null) handler();
         }
 
+        protected override void OnWorkspaceRemoveStarted(WorkspaceModel obj)
```
ikeough: Small nit: rename `obj` to `workspace`

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -136,6 +136,22 @@ private void OnInvalidRevitDocumentActivated()
...
+        {
+            base.OnWorkspaceRemoveStarted(obj);
+
+            if (obj is HomeWorkspaceModel)
+                DisposeLogic.IsClosingHomeworkspace = true;
+        }
+
+        protected override void OnWorkspaceRemoved(WorkspaceModel obj)
```
ikeough: Same nit as above.

src/Libraries/RevitServices/Elements/DisposeLogic.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace RevitServices.Elements
+{
+    public class DisposeLogic
```
ikeough: Nice! I like that this doesn't live in core any more.

ikeough:
Two small comments then LGTM. Thanks @ke-yu!

PR #556, opened 2015-08-12
Fix MAGN-7931 Dynamo/Revit hangs when operating on shared nested families

Randy-Ma:
### Purpose

This is one part of the fix to MAGN-7931. The fix here is to catch the possible exceptions when joining surfaces and if there are exceptions, null will be returned so that the dot operations to return solids will not be broken.

### Declarations

### Reviewers

@aparajit-pratap PTAL

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/RevitToProtoSolid.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/RevitToProtoSolid.cs
@@ -28,9 +28,24 @@ namespace Revit.GeometryConversion
             {
                 srfs.AddRange(face.ToProtoType(false));
             }
-            var converted = Solid.ByJoinedSurfaces(srfs);
-            srfs.ForEach(x => x.Dispose());
-            srfs.Clear();
+
+            Solid converted = null;
+            try
+            {
+                converted = Solid.ByJoinedSurfaces(srfs);
+            }
+            catch(Exception)
+            {
+                return null;
+            }
+            finally
+            {
+                srfs.ForEach(x => x.Dispose());
+                srfs.Clear();
+            }
+
+            if (converted == null)
+                return null;
 
             if (performHostUnitConversion)
                 UnitConverter.ConvertToDynamoUnits(ref converted);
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/GeometryConversion/RevitToProtoSolid.cs
```diff
@@ -28,10 +28,22 @@ public static class RevitToProtoSolid
...
                 srfs.AddRange(face.ToProtoType(false));
             }
-            var converted = Solid.ByJoinedSurfaces(srfs);
+
+            Solid converted = null;
+            try
+            {
+                converted = Solid.ByJoinedSurfaces(srfs);
+            }
+            catch (Exception e)
```
aparajit-pratap: @Randy-Ma as discussed could you try to catch `ApplicationException` and log the specific geometry error message and in `catch(Exception e) { }` block, log a localized message saying `Solid.ByJoinedSurfaces failed` or something to that extent.

Randy-Ma: @aparajit-pratap 
As discussed, if this function is called to convert a list of geometries, if an exception is let to be thrown for one geometry, other valid geometries won't be returned. Also for the defect that is fixed, the call is happening in an external dll, where you can not log a message to let it to be a warning message.

aparajit-pratap: In that case it would be better to just return `null` from the `catch` block. Then you wouldn't have to do the `null` check on `converted` below, you can simply return it at the end of the function.

aparajit-pratap:
@Randy-Ma just one more minor point above and then you can merge it after that change.

src/Libraries/RevitNodes/GeometryConversion/RevitToProtoSolid.cs
Randy-Ma: Just in case Solid.ByJoinedSurfaces() is returning null but does not throw an exception, the check for converted is still reserved. But I have changed to return null if there is an exception.

Randy-Ma:
I also removed my initial change to filter out the null result. So now the solids returned may contain null.
@aparajit-pratap PTAL

aparajit-pratap:
LGTM

Randy-Ma:
@aparajit-pratap 
Thanks a lot for reviewing the pull request!

PR #571, opened 2015-08-18
Migrationtest name fix, search tags error, and fixing copy error

holyjewsus:
a couple of errors in the last merge, dynamo customization was not being copied, the test dyn file had an incorrect name, and there were extra slashes in search tags.


holyjewsus:
going to merge this in to 2015 - watch the CI and apply it to 2014 and 2016 when I bring in name changes

PR #579, opened 2015-08-18
Fix the failing test cases caused by invalid runtime binding

Randy-Ma:
### Purpose

This is to fix the failing test cases introduced in:
https://github.com/DynamoDS/DynamoRevit/pull/556

The test cases are failing because which version of the function InternalConvert is called is determined at runtime. But unfortunately we are always assuming the result is a single geometry which is not correct. For those cases which are returning a collection of geometries, the assignment will fail.

### Declarations

### Reviewers

@aparajit-pratap PTAL

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/GeometryObjectConverter.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/GeometryObjectConverter.cs
@@ -28,7 +28,7 @@ namespace Revit.GeometryConversion
             if (geom == null) return null;
 
             dynamic dynGeom = geom;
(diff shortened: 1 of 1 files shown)


Randy-Ma:
@aparajit-pratap 
I have run the failing test cases based on the changes and they are passing. I will merge this one first but please take a look and I will address your comments separately.

aparajit-pratap:
@Randy-Ma LGTHM (looks good to have merged)

Randy-Ma:
@aparajit-pratap 
Thanks a lot for reviewing the changes!

PR #586, opened 2015-08-18
GregRevitAuth Update

pboyer:
### Purpose

Revit 2016 was crashing on package publish submission. This should fix that behavior due to using an outdated API to get the user's oxygen authentication. This fix corresponds to the following fix in the OAuth code and should be moved into 0.8.2.

https://github.com/DynamoDS/GRegOAuth/commit/a62b041a77c368914ba753be3c535b6d83e6e73c

### Declarations

### Reviewers

@ikeough


ikeough:
LGTM.

PR #589, opened 2015-08-19
Adjust to return a list to fix a failing test case

Randy-Ma:
### Purpose

This is to fix one failing test case: DynamoAllSelectionNodeTests_WithPreSelectedEntities. The test case fails because elements used by dot LINQ expressions are disposed before used. The fix here will try to convert the result to list .

### Declarations

### Reviewers

@aparajit-pratap PTAL

### FYIs

@ikeough


Randy-Ma:
@aparajit-pratap and @ikeough 

As this is a small fix to a failing test case, I will merge this in first. But if you have any comments, I will address them later.

aparajit-pratap:
Apart from the fact that this method will not do lazy evaluation any longer and if that's ok, LGTM

Randy-Ma:
@aparajit-pratap 
Are there any downside effects? I think it is ok.

PR #595, opened 2015-08-19
Fix failing test cases related to FamilyInstance

Randy-Ma:
### Purpose

This is to fix failing test cases which are introduced in the following pull request:
https://github.com/DynamoDS/DynamoRevit/pull/568

The pull request has deleted some lines I think by mistake. I am adding them back here.

### Declarations

### FYIs
@holyjewsus


holyjewsus:
@Randy-Ma thanks so much for catching and fixing this!

PR #602, opened 2015-08-20
Fix MAGN-7977 Select Face returns incorrectly transformed geometry for import instances

Randy-Ma:
### Purpose

This is to fix MAGN-7977. The cause here is similar to MAGN-4753. So the fix here is also similar to the fix in https://github.com/DynamoDS/Dynamo/pull/3038. Basically the fix is to check whether the element is an instance. If so, the same workaround used for family instances will be used for import instances. Revit experts are positive about the solution. My test also verifies that it does work.

You can notice that two surfaces do not duplicate before the fix:
image

But after the fix, they will be overlapped with each other:
image

### Declarations

### Reviewers

@pboyer PTAL


pboyer:
@Randy-Ma This seems like a reasonable fix to me! I've never really like my solution here (it's a hack), but I haven't been able to find a better solution, even after talking at length with Scott Conover.

Randy-Ma:
@pboyer 
Thanks for reviewing! I have narrowed the modification to exactly fix issues with ImportInstance.

pboyer:
LGTM

Randy-Ma:
@pboyer 
Thanks again!

PR #611, opened 2015-08-25
Fix MAGN-8056 FamilyInstance.SetRotation is not working correctly for the first element in a sliced list

Randy-Ma:
### Purpose

This is to fix MAGN-8056. It is found that setting level again before the transaction related to creating the family instance is commited will cause an exception. The reason why it is set again is:
1). At first, the family instance creation node will run to create the family instance. At this time, the level will be set to the family instance.
2). When the SetRotation node is run, the family instance will be updated which will trigger the family instance creation node be executed again. The level will be set.
The two steps are run in one transaction.

The fix here is to check whether the family instance's level is exactly the level to be set by comparing the level IDs. If so, we will not bother to set the level.

Here is the effect after the fix:
image

### Declarations

### Reviewers

@aparajit-pratap PTAL

### DIFF
--- a/src/Libraries/RevitNodes/Elements/FamilyInstance.cs
+++ b/src/Libraries/RevitNodes/Elements/FamilyInstance.cs
@@ -155,6 +155,9 @@ namespace Revit.Elements
 
         private void InternalSetLevel(Autodesk.Revit.DB.Level level)
         {
+            if (InternalFamilyInstance.LevelId.Compare(level.Id) == 0)
+                return;
+
             TransactionManager.Instance.EnsureInTransaction(Document);
 
             // http://thebuildingcoder.typepad.com/blog/2011/01/family-instance-missing-level-property.html
--- a/test/Libraries/RevitIntegrationTests/BugTests.cs
+++ b/test/Libraries/RevitIntegrationTests/BugTests.cs
@@ -25,6 +25,7 @@ using DoubleSlider = DSCoreNodesUI.Input.DoubleSlider;
 using IntegerSlider = DSCoreNodesUI.Input.IntegerSlider;
 using Dynamo.Applications.Models;
 using System;
+using Revit.Elements.InternalUtilities;
 
 namespace RevitSystemTests
 {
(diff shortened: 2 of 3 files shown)


src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -155,6 +155,9 @@ private void InitFamilyInstance(Autodesk.Revit.DB.FamilySymbol fs, XYZ pos)
 
         private void InternalSetLevel(Autodesk.Revit.DB.Level level)
         {
+            if (InternalFamilyInstance.LevelId.Compare(level.Id) == 0)
```
aparajit-pratap: Just curious, what is the `InternalFamilyInstance.LevelId` the first time the family instance is created? Does it have some default value before it is set for the first time or is it `null`?

test/Libraries/RevitIntegrationTests/BugTests.cs
```diff
@@ -912,6 +912,36 @@ public void SelectedFaceIsTransformedCorrectly_MAGN_7977()
...
+            point.X.ShouldBeApproximately(2.0);
+
+            string numberSliderNodeId = "971b7986-59e2-4ad1-b087-0e9e4158506a";
+            DoubleSlider slider = GetNode<DoubleSlider>(numberSliderNodeId) as DoubleSlider;
+            slider.Value = 10; // Set the new rotation
+
+            RunCurrentModel();
+            locations = GetPreviewCollection(locationNodeId);
+            point = locations[0] as Point;
+            point.X.ShouldBeApproximately(2.0);// The x coordination of the column should not change
```
aparajit-pratap: Can you also check for some value such as `Y` or `Z` coordinate that is supposed to change so that we know the rotation has happened?

Randy-Ma: There have already been two test cases to test the SetRotation nodes: SetFamilyInstanceRotation and Rotate_ZAxis

This test case is to check when the family instance rotates, the location does not move.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
Randy-Ma: No. For reasons described in the purpose section, the family instance is set to the same level twice.

aparajit-pratap: Yes, I got it but what is the level set to the first time it is created before it is set? The reason I ask is since it looks like it is set in this function so there will be this check even before it is set for the first time, isn't it? I suppose it is some non-null value.

test/Libraries/RevitIntegrationTests/BugTests.cs
aparajit-pratap: Yes, the location does not move, but some coordinates on the rotated elements would have changed due to the rotation about its axis. Can we check for anything to say for sure that the rotation is successful in this particular scenario?

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
Randy-Ma: The first time is when the family instance is created, the level will be set there directly.

test/Libraries/RevitIntegrationTests/BugTests.cs
Randy-Ma: I will add some checks to check the rotation angles.

Randy-Ma:
@aparajit-pratap 
I have updated the test case to also check the rotation angle.

aparajit-pratap:
Thanks @Randy-Ma! LGTM

Randy-Ma:
@aparajit-pratap 
Thanks for all the suggestions!

PR #612, opened 2015-08-25
MAGN-8222 Sort Element.Parameters when getting a parameter by name

pboyer:
### Purpose

In order to fix [MAGN-5870](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-5870), @Randy-Ma made a change to use `Element.GetOrderedParameters`. This was a sensible approach, but it turns out `GetOrderedParameters` does not return all of the `Parameter`s for an `Element`. This is from the API docs:

> The collection consists of only visible parameters associated to the element; it returns a different list than Element.Parameters.

So, we need to manually sort them, which is done in this PR. Here is the result:

sortd

### Declarations

 - I ran the test from created to validated MAGN-5870 as well
image

### Reviewers

@ikeough 

### FYIs

@jnealb @RodRecker

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -316,8 +316,15 @@ namespace Revit.Elements
         {
             object result;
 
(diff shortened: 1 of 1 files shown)


ikeough:
LGTM.

PR #620, opened 2015-08-27
DirectShape Nodes

mjkkirschner:
### Purpose

This PR adds 2 DirectShape constructors

```
DirectShape.ByGeometry
DirectShape.ByMesh
```

http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8012

screen shot 2015-08-26 at 7 51 28 pm

screen shot 2015-08-31 at 8 46 01 pm

This PR enables Meshes, Solids, and Surfaces to be wrapped by a DirectShape, given a name, category, and material and placed into the Revit Model. If a user uses the overload with no material argument the DynamoMaterial for Revit Preview is used (orange glass)

### Impact assessment

Addition of 2 Revit Nodes
Changes to ToRevitType For ProtoToRevitMesh() Geometry converter and the addition of a mesh to RevitMesh method.

#### References

@Racel and I have met about icons for these 4 constructors.

#### Testing strategy

Currently there are node and system tests for this functionality.

#### Note:

 there are some failing tests, 1 node test is failing because the MaterialsManager is not initialized for tests, this is not a big issue as the other tests test the same pathway with materials successfully. 

The larger issue is two failing system tests that I cannot reproduce through the application, the same procedures run without error when used in the UI.  - When running through RTF these tests throw this exception upon starting a transaction after rebinding - 
```
Cannot modify the document for either a read-only external command is being executed, or changes to the document are temporarily disabled
```
I am currently blocked in finding a solution to this issue when testing.
I think this code and functionality should be reviewed now and more robust tests created as a followup task.

### Implementation:

This node follows the same pattern as other Revit elements like family instance. I have implemented rebinding for geometry, name, and material changes. If the category is changed a new directShape will need to be generated and rebinding is not possible.

One issue with geometrical rebinding is that when a rebound directShape updates the Revit geometry objects on the real Revit element an `OnNodeModified` event is triggered on the node and the directshape constructor node is re-executed.
Currently, stopping this infinite loop requires that when the geometry is set, the constructor writes into the `Tags` dictionary of the protogeometry `Geometry or Mesh, the elementID` of the directShape element and the currently assigned `materialID`. When the constructor is modified again, it will not re assign the shape again if it detects its own elementID and materialID present in the Tags dictionary of the input protogeometry.

We also now write a syncID guid into trace and into the tags dict, this value is updated when geometry is rebound and if the ids are out of sync then we update the geometry when rebinding.

It's possible this workaround is causing the issues seen when testing, but I am not sure.

### Changes:

I have modified the ProtoToRevitMesh methods to take TessellatedShapeBuilder targets and fallback settings as well as materialIds so surfaces and solids can target higher quality geometry when being tessellated.

### Reviewers

@pboyer 
@ikeough 

### FYIs

@jnealb @Racel


mjkkirschner:
made a slight change - extracted a longer exception message to resx file

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+        /// <summary>
+        /// Sets the internalDirectShape to point to some geometry,
+        /// this method also generates tessellated geometry from the protogeometry object
+        /// and sets the material of the generated Revit faces
+        /// </summary>
+        /// <param name="shapeReference"></param>
+        /// <param name="materialId"></param>
+        private void InteralSetShape(DesignScriptEntity shapeReference, ElementId materialId)
```
pboyer: Typo

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+            //the value stored at that key is equal to the materialId we're trying to set then we don't
+            //need to regenerate the geometry
+            if((shapeReference.Tags.LookupTag(this.InternalElementId.ToString()) as ElementId) == materialId)
+            {
+                return;
+            }
+
+            TransactionManager.Instance.EnsureInTransaction(Document);
+
+            IList<GeometryObject> tessellatedShape = GenerateTessellatedGeo(shapeReference, materialId);
```
pboyer: Can you just simply use `var`?

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+                DocumentManager.Regenerate();
+                var Revitbb = this.InternalElement.get_BoundingBox(null);
+                TransactionManager.Instance.TransactionTaskDone();
+                var bb = Revitbb.ToProtoType();
+                var cube = bb.ToCuboid();
+                var point = cube.Centroid();
+                bb.Dispose();
+                cube.Dispose(); 
+              
+                return point;
```
pboyer: This would not actually calculate the centroid, simply the bounding box centroid. I don't see a strong reason to have this property anyway, let's remove it.

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+        /// <summary>
+        /// Create a Revit DirectShape given some geometry, a name for the shape, and category.
+        /// The geometry will be tessellated before being placed in the Revit model
+        /// The category of a DirectShape cannot be changed after placing a DirectShape, so
+        /// a new DirectShape will be generated if the category input is changed
+        /// </summary>
+        /// <param name="geometry">a surface that will be tessellated and placed in the Revit model as a DirectShape</param>
+        /// <param name="name">a string name for the directshape</param>
+        /// <param name="category">must be a top level built-in category</param>
+        /// <returns>a DirectShape Element</returns>
```
pboyer: Let's do the capitalization right. :)

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+            {
+                IsRevitOwned = isRevitOwned
+            };
+        }
+
+        #endregion
+
+        public override string ToString()
+        {
+            return InternalDirectShape.Name;
```
pboyer: I'm guessing this produces something sensible when the user has not supplied a name?

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+        #endregion
+
+        #region Private mutators
+        /// <summary>
+        /// SetInternalElement to a DirectShape element
+        /// </summary>
+        /// <param name="shape"></param>
+        private void InteralSetDirectShape(Autodesk.Revit.DB.DirectShape shape)
```
pboyer: Typo

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+    public class DirectShape : Element
+    {
+        
+        #region Private Properties
+
+        private static Guid DYNAMO_DIRECTSHAPE_APP_GUID
+        {
+            get
+            {
+                return Guid.Parse("03deaabe-6989-4d20-ba68-557001d2f45c");
```
pboyer: Use a static constructor for building reference types whose value never changes.

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+            {
+                tessellatedShape = (shapeReference as Autodesk.DesignScript.Geometry.Surface).ToRevitType(TessellatedShapeBuilderTarget.AnyGeometry, TessellatedShapeBuilderFallback.Mesh, MaterialId: materialId);
+            }
+
+            else if (shapeReference is Autodesk.DesignScript.Geometry.Solid)
+            {
+                tessellatedShape = (shapeReference as Autodesk.DesignScript.Geometry.Solid).ToRevitType(TessellatedShapeBuilderTarget.AnyGeometry, TessellatedShapeBuilderFallback.Mesh, MaterialId: materialId);
+            }
+
+            if (tessellatedShape == null)
```
pboyer: When you have really long lines, it's nice to break them up.
```
Foo( Fooooooooooooo fooooooooooooooo, 
         Fooooooooooooo2 fooooooooooooooo2 ){
...
}
```

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+            IList<GeometryObject> tessellatedShape = null;
+            if (shapeReference is Autodesk.DesignScript.Geometry.Mesh)
+            {
+                tessellatedShape = (shapeReference as Autodesk.DesignScript.Geometry.Mesh).ToRevitType(TessellatedShapeBuilderTarget.Mesh, TessellatedShapeBuilderFallback.Salvage, MaterialId: materialId);
+            }
+            else if (shapeReference is Autodesk.DesignScript.Geometry.Surface)
+            {
+                tessellatedShape = (shapeReference as Autodesk.DesignScript.Geometry.Surface).ToRevitType(TessellatedShapeBuilderTarget.AnyGeometry, TessellatedShapeBuilderFallback.Mesh, MaterialId: materialId);
+            }
+
```
pboyer: It's weird to have this empty line here.

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,403 @@
...
+        {
+          
+            if (name != this.InternalDirectShape.Name)
+            {
+                TransactionManager.Instance.EnsureInTransaction(Document);
+                this.InternalDirectShape.SetName(name);
+                TransactionManager.Instance.TransactionTaskDone();
+            }
+
+
```
pboyer: Are you saving this space for something else?

pboyer:
@mjkkirschner Looks good. I'd love to have some feedback from @mccrone and yourself. I have a few questions:

1. Do we really need to force the user to assign a name to any `DirectShape`? It sounds cumbersome. Maybe we could supply a default if the name is not supplied?
2. The `ByGeometry...` nodes currently tessellate the geometry. What if `DirectShape` is improved to support smooth surfaces? Will we change the behavior of the node? That's an unusual thing to do, but maybe it's okay here. 
3. I'd like to get some feedback on the descriptions and wording - it would be great to get some feedback from @mccrone, keeping in mind that it's a lot easier to get it right the first time around. :)

pboyer:
Another thing, is it possible to have fewer constructors if we supplied sensible defaults for the `Material`?

mjkkirschner:
@pboyer yes, it would, we could then get rid of the two material-less constructors - but we cannot currently (as far as I know) supply a default value other than `null` and that would look strange to a user I supposed.

pboyer:
OK, you're probably right. I'm just trying to think about how we can make those constructors a bit less daunting, but @mccrone should have the final word...

mjkkirschner:
1 . I think name can be defaulted -, multiple direct shapes with the same name does not cause issues as far as I have seen or read.

2. I'm not sure on this one, if that change is made in the API... we could migrate this node in the future with an intermediate tessellation node, or change the behavior.

src/Libraries/RevitNodes/Elements/DirectShape.cs
mjkkirschner: @pboyer good catch! I added this property because DirectShapes are not location based and thought it would be useful to get *some* point representing the direct shape without needing to do the costly operation of extracting all geometry and converting it to protogeo.

mjkkirschner:
@pboyer Actually! I did not know about [DefaultArguementAttribute] - I will try this out as a way to provide a default material and get rid of 2 constructors.

Racel:
As I mentioned in my email, I like getting rid of the separate material constructors! I'm glad there is [DefaultArgumentAttribute], because (as Mike said) 'null' would probably be confusing to the user.

mjkkirschner:
@pboyer please take another look - I have implemented the geometry synchronization by using a guid in raw trace, given the constructors defaults for name, and material.

next commit will add icons.

mjkkirschner:
screen shot 2015-08-30 at 10 20 45 pm

Yay! icons added Thanks @Racel !

Racel:
np, @mjkkirschner! looks good.

are we not shortening the name to ByGeometry and ByMesh? I think the additional text is a bit confusing. Are the long names in anticipation of including additional ByGeometry and ByMesh nodes in the future?

mjkkirschner:
@Racel I believe our naming guide enforces the entire name of all parameters in order... maybe thats not true?

Racel:
@mjkkirschner - do you mean it automatically does it on zero touch nodes or is it a strong suggestion in API documentation? I don't think 'Point.ByCoordinates' works this way.

mccrone:
@mjkkirschner you are absolutely allowed to be judicious about naming things.  Naming all inputs isn't always a good thing if there are more than two.

mjkkirschner:
Thanks @mccrone

mccrone:
I'm back in the office today and will absolutely take a look at this this afternoon!

pboyer:
@mmcrone - do you think we should name these ByGeometry and ByMesh? The
long name is painful in code block nodes, screws up search, makes for an
unnecessarily bulky node.

mccrone:
@pboyer precisely what I was thinking.  I'll review the rest of the nodes later today too.

mjkkirschner:
@mccrone @Racel @pboyer I have changed the names of the nodes and verified that the icons work.

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,471 @@
...
+            TransactionManager.Instance.TransactionTaskDone();
+        }
+
+        /// <summary>
+        /// Sets the internalDirectShape to have a new name
+        /// if this method detects the default string it generates a more specific name
+        /// </summary>
+        /// <param name="name"></param>
+        private void InteralSetName(DesignScriptEntity shapeReference,string name)
```
pboyer: Typo is still here...

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,471 @@
...
+        {
+            if (name == DEFAULT_NAME)
+            {
+                name = name.Replace("...", shapeReference.ToString());
```
pboyer: Use String.Format and follow the ProtoGeometry style, as shown here:

https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/Display/Display.cs#L144

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,471 @@
...
+        }
+    }
+
+    /// <summary>
+    /// A Revit DirectShape, which is a wrapper for some other geometrical entities 
+    /// </summary>
+    [DynamoServices.RegisterForTrace]
+    public class DirectShape : Element
+    {
+        private const string DEFAULT_NAME = "directShape from ...";
```
pboyer: Make it just `DirectShape`

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,471 @@
...
+        }
+
+        /// <summary>
+        /// Sets the internalDirectShape's shape to point to some geometry,
+        /// this method also generates tessellated geometry from the protogeometry object
+        /// and sets the material of the generated Revit faces
+        /// </summary>
+        /// <param name="shapeReference"></param>
+        /// <param name="materialId"></param>
+        private void InteralSetShape(DesignScriptEntity shapeReference, ElementId materialId,string currentSyncId)
```
pboyer: Typo

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,471 @@
...
+        /// Create a Revit DirectShape given some mesh, a name for the shape, category, and material.
+        /// The category of a DirectShape cannot be changed after placing a DirectShape, so
+        /// a new DirectShape will be generated if the category input is changed.
+        /// </summary>
+        /// <param name="mesh">a mesh that will be placed in the Revit model as a DirectShape</param>
+        /// <param name="name">a string name for the directshape</param>
+        /// <param name="category">must be a top level built-in category</param>
+        /// <param name="material">a material to apply to the faces of the DirectShape</param>
+        /// <returns>a DirectShape Element</returns>
+        public static DirectShape ByMesh(Autodesk.DesignScript.Geometry.Mesh mesh,
```
pboyer: Please correct all capitalization issues.

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -0,0 +1,471 @@
...
+        {
+            DynamoPreviewMaterial = Material.ByName(Document.GetElement(MaterialsManager.Instance.DynamoMaterialId).Name);
+            DYNAMO_DIRECTSHAPE_APP_GUID = Guid.Parse("03deaabe-6989-4d20-ba68-557001d2f45c");
+        }
+
+        #endregion
+
+        public override string ToString()
+        { 
+            return InternalDirectShape.Name;
```
pboyer: Use the style we use with the DisplayObject:

https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/Display/Display.cs#L144

test/Libraries/RevitIntegrationTests/DirectShapeTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            Assert.IsNotNull(pnt);
+
+            Assert.IsTrue(IsFuzzyEqual(pnt.X, 10.0, 1.0e-6));
+            Assert.IsTrue(IsFuzzyEqual(pnt.Y, 10.0, 1.0e-6));
+            Assert.IsTrue(IsFuzzyEqual(pnt.Z, 10.0, 1.0e-6));
+
+        }
+
+        [Test, Category("Failure")]
+        [TestModel(@".\DirectShape\singleDirectShape.rvt")]
```
pboyer: Create follow up tasks to fix.

mjkkirschner:
@pboyer one more look! I added geometry, category, and material names to the name string if it is not set by the user.

I will create a followup task for the tests

mjkkirschner:
task is here : http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8297

pboyer:
LGTM!

mccrone:
The names are good.  I made some QA notes about a crash and a failure to fail quickly at http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8012

mjkkirschner:
merging this in, PRS for 2016 and 2017 incoming

PR #623, opened 2015-08-27
Fix MAGN-7647 cannot use "Select Divided Surface Families" node to select divided surface in Revit

Randy-Ma:
### Purpose

A discussion with @monikaprabhu and @riteshchandawar concludes that we change the text in the node so that users will know that if there are no family instances in the divided surface.

### Declarations

### Reviewers

@aparajit-pratap PTAL

### FYIs

@monikaprabhu 
@riteshchandawar

### DIFF
--- a/src/Libraries/RevitNodesUI/Selection.cs
+++ b/src/Libraries/RevitNodesUI/Selection.cs
@@ -29,6 +29,7 @@ using UV = Autodesk.DesignScript.Geometry.UV;
 using RevitServices.EventHandler;
 using Autodesk.Revit.DB.Events;
 using Dynamo.Applications;
+using DSRevitNodesUI.Properties;
 
 namespace Dynamo.Nodes
 {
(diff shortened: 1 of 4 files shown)


src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -706,6 +707,16 @@ protected override IEnumerable<Element> ExtractSelectionResults(DividedSurface s
                 RevitElementSelectionHelper<DividedSurface>.GetFamilyInstancesFromDividedSurface(
                     selection);
         }
+
+        public override string ToString()
```
aparajit-pratap: Just curious how does this `ToString` message transfer to the node warning?

Randy-Ma: This will not be transferred to the warning tooltip. It is to be displayed right in the node.

aparajit-pratap: Got it. LGTM!

Randy-Ma:
@aparajit-pratap 
Thanks a ton for reviewing the changes!

PR #624, opened 2015-08-28
MAGN-8200 Hide methods and properties in DynamoCore.dll which should not be exposed via the API.

ikeough:
See https://github.com/DynamoDS/Dynamo/pull/5195 for details.

Requires merge:
-[ ] Revit2015
-[ ] Revit2014


pboyer:
LGTM!

PR #630, opened 2015-09-02
Fix MAGN-7741 Element.OverrideColorInView fails on 2d elements(curves) in Revit

Randy-Ma:
### Purpose

When overriding colors, right now only face colors will be overridden. In this fix, I am making the projection lines' colors overridden as well.

### Declarations

### Reviewers

@aparajit-pratap PTAL


aparajit-pratap:
LGTM!

PR #637, opened 2015-09-07
Changes to ExtractSelectionResults to catch exceptions

Randy-Ma:
### Purpose

It is found that GetFamilyInstancesFromDividedSurface may throw exception. The exception makes the text in the selection node not updated as expected. Here the exception is caught and an empty list is returned

This pull request is related to:
https://github.com/DynamoDS/DynamoRevit/pull/623

### Declarations

### Reviewers

@aparajit-pratap PTAL


aparajit-pratap:
LGTM!

PR #641, opened 2015-09-08
Read automation information from journal.

ikeough:
### Purpose

This PR augments the ability to run Dynamo from a journal file. Previously, you could specify a Dynamo workspace to open from a journal file. But, because `IsTestMode` was set to `false`, a workflow once opened, even in automatic mode, would not run as the Dynamo Revit application would never go into idling mode.  Now `DynamoRevit.Execute` does the following:
- Checks the journal file's `dynShowUI` key and sets a private static field to true or false.
- Checks the journal's `dynAutomation` key and sets `IsTestMode` to correspond. `IsTestMode` is set to `false` if this key is not present or the journal file data is null.
- Checks the journal's `dynPath` key. If it exists and points to a proper path, it then opens (and runs) the workspace. 

### Declarations

  - Added null checks and block formatting.
  - No new user facing strings introduced.

### Reviewers

@Randy-Ma This PR is for work not in the current sprint, but it is in support of @andydandy74's RTCEurope class "Automating Automation". I'd appreciate your quick review.

### FYIs

@andydandy74

### Cherry Picks:
- [ ] Revit2015


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -414,4 +460,17 @@ public override IEnumerable<string> GetDynamoInstallLocations()
                 (s) => regKey.OpenSubKey(s).GetValue("InstallLocation") as string);
         }
     }
+
+    internal class AutomationThread : ISchedulerThread
```
Randy-Ma: It seems this scheduler thread is temporarily not used.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -281,11 +291,47 @@ private void HandleDebug(ExternalCommandData commandData)
...
+        private static bool ShowUI(ExternalCommandData commandData)
+        {
+            if (commandData.JournalData == null)
+            {
+                return true;
+            }
+
+            if (commandData.JournalData.ContainsKey("dynShowUI"))
+            {
+                return bool.Parse(commandData.JournalData["dynShowUI"]);
```
Randy-Ma: TryParse is better so that there will be no exceptions thrown.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -281,11 +291,47 @@ private void HandleDebug(ExternalCommandData commandData)
...
+        private static bool IsAutomationMode(ExternalCommandData commandData)
+        {
+            if (commandData.JournalData == null)
+            {
+                return false;
+            }
+
+            if (commandData.JournalData.ContainsKey("dynAutomation"))
+            {
+                return bool.Parse(commandData.JournalData["dynAutomation"]);
```
Randy-Ma: Similarly TryParse is better

Randy-Ma:
This is awesome! Except for some minor suggestions, LGTM

src/DynamoRevit/DynamoRevit.cs
ikeough: Very good point @Randy-Ma. Will do.

src/DynamoRevit/DynamoRevit.cs
ikeough: I mistakenly left this class in. It's not used. I'll remove it.

PR #643, opened 2015-09-09
MAGN-8187 Level.ByElevationName hangs Revit 

mjkkirschner:
### Purpose

The issue in this task http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8187 was found to be occurring because element rebinding would cause an infinite loop in run automatic when updating the name of the revit levels.

We currently update Revit level names so they do not have the same name, i.e.. an input of level produces level, level(1) level(2) etc... when this modified name is compared to the original input name, they are different so another modification occurs... causing the loop.

instead of finding all string comparison edge cases that can occur with various user input of names which could contain "()" or other inputs which make it difficult or impossible to tell if the name is a modified name or not, I think the more robust fix is to use the raw trace mechanism.

We store the desired input name (the string that was the input argument to the node) inside trace when the name is updated. Later when rebinding occurs we compare our input name for the new run, to the original name in trace and do not update the level name, if these are the same.

Since the method for getting the raw trace data and revit element out of raw trace was almost identical to the method for DirectShape I replaced both of these with a generic method inside ElementBinder.

#### Testing... 
I have been unable to find a mechanism to test this rebinding issue with Levels or DirectShapes, and it appears that the same exception is being thrown here that prevented testing of directShape rebinding. 

As far as I can tell when `onNodeModifed` events occur on Revit Constructor nodes, and attempt to start transactions in Revit System Tests, an exception is thrown from the revit API that the document cannot be modified when attempting to start a transaction, and instead the elements are not further modified... this stops the test from running in an infinite loop so it's unclear if our tests can currently catch this defect on any revit wrapper type.

I've included one test that passes, but marked it ignore because it passes before the fix as well, due to the issue mentioned above.

@ikeough explained that the Revit Test Framework runs in a different manner than during normal application use and this may be related to this issue. I think we should create a task around hardening / investigating this issue if possible, preferably someone familiar with the difference between RTF runs and regular DynamoRevit runs.

**At this time I'm unsure how to test this form of defect.**

### Declarations

### Reviewers

@ikeough 

### FYIs
@Randy-Ma 
@mccrone

### DIFF
--- a/src/Libraries/RevitNodes/Elements/DirectShape.cs
+++ b/src/Libraries/RevitNodes/Elements/DirectShape.cs
@@ -133,7 +133,7 @@ namespace Revit.Elements
         {
             //Phase 1 - Check to see if a DirectShape exists in trace and should be rebound
             var oldShape =
-                ElementBinder.GetElementAndTraceData<Autodesk.Revit.DB.DirectShape,DirectShapeState>();
+                ElementBinder.GetElementAndTraceData<Autodesk.Revit.DB.DirectShape,DirectShapeState>(Document);
 
             //There was a oldDirectShape, rebind to that
             if (oldShape != null)
--- a/src/Libraries/RevitNodes/Elements/DirectShape.cs
+++ b/src/Libraries/RevitNodes/Elements/DirectShape.cs
@@ -133,7 +133,7 @@ namespace Revit.Elements
         {
             //Phase 1 - Check to see if a DirectShape exists in trace and should be rebound
             var oldShape =
-                GetElementAndTraceData();
+                ElementBinder.GetElementAndTraceData<Autodesk.Revit.DB.DirectShape,DirectShapeState>();
 
             //There was a oldDirectShape, rebind to that
             if (oldShape != null)
@@ -153,7 +153,7 @@ namespace Revit.Elements
                     //we also check the material, if it's different than the currently assigned material
                     //then we need to rebuild the geo so that a new material is applied
                     //
-                    this.IntenralSetShape(shapeReference,new ElementId(material.Id),oldShape.Item2);
+                    this.IntenralSetShape(shapeReference,new ElementId(material.Id),oldShape.Item2.syncId);
                     this.InternalSetName(shapeReference,shapeName,material,category);
                     return;
                 }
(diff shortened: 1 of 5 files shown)


RodRecker:
@Randy-Ma is doing some work on the test harness that may help.

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -153,7 +153,7 @@ private void InitDirectShape(Autodesk.Revit.DB.DirectShape shape)
                     //we also check the material, if it's different than the currently assigned material
                     //then we need to rebuild the geo so that a new material is applied
                     //
-                    this.IntenralSetShape(shapeReference,new ElementId(material.Id),oldShape.Item2);
+                    this.IntenralSetShape(shapeReference,new ElementId(material.Id),oldShape.Item2.syncId);
```
ikeough: Spelling fix: `this.InternalSetShape(...)`

src/Libraries/RevitServices/Persistence/ElementBinder.cs
```diff
@@ -349,9 +351,36 @@ public static ISerializable GetRawDataFromTrace()
...
         }
-        
 
         /// <summary>
+        /// Get the Element of type T and the raw traceData of type K from Thread Local Storage.
+        /// Requires that K inherits from SerializableId so the element can be retrieved from the Revit Document
+        /// </summary>
+        /// <returns></returns>
+        public static Tuple<T, K> GetElementAndTraceData<T, K>(Document document) 
+            where T : Autodesk.Revit.DB.Element
```
ikeough: Make this `TElement` and `TId`. Nobobdy will remember what `K` is.

ikeough:
@mjkkirschner A couple of small comments inline.

Are there other places that we do this automatic renaming? Your strategy is a good one. I wonder if it applies for things like grids as well.

mjkkirschner:
@ikeough made the changes, PTAL.

I can't think of an edge case...yet.

mjkkirschner:
@ikeough also I did not find any other elements that used the auto renaming methods

ikeough:
Thanks @mjkkirschner LGTM.

PR #647, opened 2015-09-10
Direct shape.byMesh does not support non-planar quads

mjkkirschner:
### Purpose
It was found that the method for converting from Dynamo Meshes to Revit Meshes using the TessellatedShapeBuilder failed when the Dynamo mesh was built with quads and those quads were non planar. The TessellatedShapeBuilder does not support non planar quads, but it does support quads.

~~To fix the issue we look for quad faces in the Dynamo mesh that are non planar by comparing the normals of two triangles defined by quad vertices. If the normals are not parallel than we skip adding the quad and instead add two triangles to the tessellated shape builder that represent this quad.~~

~~We also check for one case where the quad is planar but twisted by checking the quad for self intersections.~~

Instead we just triangulate all quads when converting from a  Proto Mesh to Revit Mesh

http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8322

screen shot 2015-09-10 at 2 28 00 pm

screen shot 2015-09-09 at 4 16 49 pm

screen shot 2015-09-10 at 2 33 11 pm

### Testing
I've added tests for the above pictured cases and a simple non planar quad with one point out of plane.

### Declarations

### Reviewers

@pboyer 

### FYIs
@jnealb 
@Randy-Ma

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
@@ -143,13 +143,13 @@ namespace Revit.GeometryConversion
                 if (f.Count > 3)
                 {
                     //test if the face is a planar...
-                    var BACNormal = Vector.ByTwoPoints(currentVerts[1], currentVerts[0]).Cross(Vector.ByTwoPoints(currentVerts[1], currentVerts[2])).Normalized();
+                    var CABnormal = Vector.ByTwoPoints(currentVerts[2], currentVerts[0]).Cross(Vector.ByTwoPoints(currentVerts[2], currentVerts[1])).Normalized();
                     var DACNormal = Vector.ByTwoPoints(currentVerts[3], currentVerts[0]).Cross(Vector.ByTwoPoints(currentVerts[3], currentVerts[2])).Normalized();
 
                     //if the two triangle normals are not parallel then we have a non planar quad
                     //and we'll skip adding this face and instead add two new faces to list to process
                     //these new faces are two triangles representing the quad
-                    if (Math.Abs(BACNormal.Dot(DACNormal) - 1) > 0.000001 ||
+                    if (Math.Abs(CABnormal.Dot(DACNormal) - 1) > 0.000001 ||
                         //or if there are any self intersections between non 
                         //contigous polygon edges we also triangulate, this finds a twisted quad
                         QuadSelfIntersects(currentVerts)
--- a/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
@@ -93,8 +93,8 @@ namespace Revit.GeometryConversion
             var result = tsb.Build(target, fallback, ElementId.InvalidElementId);
             return result.GetGeometricalObjects();
         }
-       
-        //this method checks intersections between non contigous edges of a quad for self intersection
+
+        //this method checks intersections between non contiguous edges of a quad for self intersection
         private static bool QuadSelfIntersects(List<Autodesk.DesignScript.Geometry.Point> quadVerts)
         {
             var AB = Autodesk.DesignScript.Geometry.Line.ByStartPointEndPoint(quadVerts[0], quadVerts[1]);
@@ -142,7 +142,7 @@ namespace Revit.GeometryConversion
                 }
                 if (f.Count > 3)
                 {
-                    //test if the face is a planar...
+                    //test if the face is a planar by comparing the cross products (normals) of two tri edges
                     var CABnormal = Vector.ByTwoPoints(currentVerts[2], currentVerts[0]).Cross(Vector.ByTwoPoints(currentVerts[2], currentVerts[1])).Normalized();
                     var DACNormal = Vector.ByTwoPoints(currentVerts[3], currentVerts[0]).Cross(Vector.ByTwoPoints(currentVerts[3], currentVerts[2])).Normalized();
 
@@ -151,7 +151,7 @@ namespace Revit.GeometryConversion
                     //these new faces are two triangles representing the quad
                     if (Math.Abs(CABnormal.Dot(DACNormal) - 1) > 0.000001 ||
                         //or if there are any self intersections between non 
-                        //contigous polygon edges we also triangulate, this finds a twisted quad
+                        //contiguous polygon edges we also triangulate, this finds a twisted quad which is planar
                         QuadSelfIntersects(currentVerts)
                         )
                     {
--- a/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
@@ -108,7 +108,7 @@ namespace Revit.GeometryConversion
             var tsb = new TessellatedShapeBuilder();
             tsb.OpenConnectedFaceSet(false);
 
-            for (int faceindex = 0; faceindex < indices.Count(); faceindex++)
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
```diff
@@ -121,7 +140,27 @@ public static class ProtoToRevitMesh
...
+                if (f.Count > 3)
+                {
+                    //test if the face is a planar by comparing the cross products (normals) of two tri edges
+                    var CABnormal = Vector.ByTwoPoints(currentVerts[2], currentVerts[0]).Cross(Vector.ByTwoPoints(currentVerts[2], currentVerts[1])).Normalized();
+                    var DACNormal = Vector.ByTwoPoints(currentVerts[3], currentVerts[0]).Cross(Vector.ByTwoPoints(currentVerts[3], currentVerts[2])).Normalized();
+
+                    //if the two triangle normals are not parallel then we have a non planar quad
+                    //and we'll skip adding this face and instead add two new faces to list to process
+                    //these new faces are two triangles representing the quad
+                    if (Math.Abs(CABnormal.Dot(DACNormal) - 1) > 0.000001 ||
```
ikeough: Make this "magic number" `private const readonly precision`.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
```diff
@@ -121,7 +140,27 @@ public static class ProtoToRevitMesh
...
+                        //or if there are any self intersections between non 
+                        //contiguous polygon edges we also triangulate, this finds a twisted quad which is planar
+                        QuadSelfIntersects(currentVerts)
+                        )
+                    {
+                        indicies.Add(IndexGroup.ByIndices(f.B, f.C, f.A));
```
ikeough: Spelling police: "Indices"

test/Libraries/RevitNodesTests/Elements/DirectShapeTests.cs
```diff
@@ -148,6 +156,114 @@ public void ByName_NonexistentName()
...
+        [TestModel(@".\empty.rfa")]
+        public void ByMeshNameCategoryMaterial_TriWith4PointsPlanarForms()
+        {
+
+            var p1 = Point.ByCoordinates(0.0, 0.0, 0.0);
+            var p2 = Point.ByCoordinates(2.5, 0.0, 0.0);
+            var p3 = Point.ByCoordinates(5.0, 0.0, 0.0);
+            var p4 = Point.ByCoordinates(0.0, 5.0, 0.0);
+
+            var index1 = IndexGroup.ByIndices(0, 1, 2, 3);
```
ikeough: Starting here, I see a lot of copy paste code. Refactor the code that is common amongst these test methods. Something like `CreateDirectShapeFromQuadPoints`. It will make these tests easier to read.

ikeough:
@mjkkirschner Nice fix. A couple of small comments.

mjkkirschner:
@pboyer @ikeough after speaking with pboyer about this fix we decided to instead just triangulate all incoming quads as this is less likely to break in the future as tolerances change in TessellatedShapeBuilder. I've refactored the tests and fixed the typos as well as modified the ProtoToRevitMesh to no longer look for planar quads, instead any quad is triangulated.

jnealb:
@mjkkirschner Perhaps put your work in a package?

ikeough:
@mjkkirschner Sounds good to me. Thanks! LGTM.

mjkkirschner:
@pboyer I'll wait to merge and cherry pick this until you get a chance to take a look.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
```diff
@@ -102,35 +101,42 @@ public static class ProtoToRevitMesh
...
-                for (int i = 0; i < f.Count; i++)
+                var currentIndices = new List<uint>() { f.A, f.B, f.C, f.D };
+
+                //if this is a quad face triangulate it and skip the rest of this loop
+                if (f.Count > 3)
                 {
-                    var currentindex = Convert.ToInt32(currentIndicies[i]);
+                    indices.Add(IndexGroup.ByIndices(f.B, f.C, f.A));
+                    indices.Add(IndexGroup.ByIndices(f.A, f.C, f.D));
+                    continue;
```
pboyer: Hate to be the performance guy, but is there a strong reason to add to the original list while iterating over it? Your iteration will terminate, but the code is a bit strange to read and memory intensive. Consider the case where you're dealing with a 1 million face quad mesh - now you have 3 millions index groups in this list.

pboyer: You should also keep in mind that the `IndexGroup.ByIndices` call actually has a lot of function call overhead.

mjkkirschner:
@pboyer @ikeough - I profiled this code on Pboyer's suggestion - and made some changes based on that. I removed a linq call, removed a toList call on the Mesh.indicies, and avoid creating a new list of indicies at any point, instead just adding the faces to the tessellatedShape builder immediately.

@pboyer please take another look.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
```diff
@@ -102,35 +101,34 @@ public static class ProtoToRevitMesh
...
             var verts = mesh.VertexPositions;
-            var indicies = mesh.FaceIndices;
+            var indices = mesh.FaceIndices;
 
-            var currentVerts = new List<Autodesk.DesignScript.Geometry.Point>();
             var tsb = new TessellatedShapeBuilder();
             tsb.OpenConnectedFaceSet(false);
 
-            foreach (var f in indicies)
+            for (int faceindex = 0; faceindex < indices.Count(); faceindex++)
```
pboyer: If the length of the array is not changing, you can do this to remove the cost of `indices.Count()`:

```csharp
for (var i = 0, l = indices.Count(); i < l; i++){
...
}
```

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
```diff
@@ -102,35 +101,34 @@ public static class ProtoToRevitMesh
...
+                    var tri2 = IndexGroup.ByIndices(f.A, f.C, f.D);
 
-                    currentVerts.Add(verts[currentindex]);
+                    AddProtoFaceToTSB(tsb, tri1, verts, performHostUnitConversion, MaterialId);
+                    AddProtoFaceToTSB(tsb, tri2, verts, performHostUnitConversion, MaterialId);
+                }
+                else
+                {
+                    AddProtoFaceToTSB(tsb, f, verts, performHostUnitConversion, MaterialId);
                 }
```
pboyer: Nice, this is what I would've expected.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitMesh.cs
```diff
@@ -140,5 +138,28 @@ public static class ProtoToRevitMesh
 
             return result.GetGeometricalObjects();
         }
+
+        //this method converts a proto indexGroup and verts to a tessellated face, and adds it
+        //to the tessellated shape builder that is passed in.
+        //the verts array should be the entire vert array extracted from a proto mesh.
+        private static void AddProtoFaceToTSB(TessellatedShapeBuilder tsb,
```
pboyer: Use XML documentation with three forward slashes.

pboyer:
@mjkkirschner This definitely addresses my concerns. Please take a look at the above comments (both minor) and feel free to merge.

PR #652, opened 2015-09-15
disable raw trace writing when the element binder is turned off(in RevitNodeTests)

mjkkirschner:
### Purpose

This PR fixes two failing tests, these tests fail because they use levels that now use ``SetRawTrace`` instead of ``SetElementForTrace``.

``SetRawTrace`` did not check if the element binder was turned off before writing into trace so rebinding was occurring during testing which was causing the issue with these tests as they both try to create multiple levels but the level was being mutated instead.

@lukechurch @Randy-Ma is there a reason that setRawTrace did not have this check for the element Binder being enabled? Is it known if this was intentional?

### Declarations

### Reviewers

@lukechurch -

### FYIs

@Randy-Ma

### DIFF
--- a/src/Libraries/RevitServices/Persistence/ElementBinder.cs
+++ b/src/Libraries/RevitServices/Persistence/ElementBinder.cs
@@ -340,6 +340,7 @@ namespace RevitServices.Persistence
         /// <param name="data"></param>
         public static void SetRawDataForTrace(ISerializable data)
         {
+            if (!IsEnabled) return;
             TraceUtils.SetTraceData(REVIT_TRACE_ID, data);
         }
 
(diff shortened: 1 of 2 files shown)


Randy-Ma:
My opinion is that if a test case need element binding, then this test case need to be moved to the system tests. But the fix here to check whether element binder is turned off or not is sensible to me. @lukechurch can correct me if anything is wrong.

mjkkirschner:
@Randy-Ma thanks, I agree.
 The issue was that these tests depended on elementBinder being turned off, and when I switched levels to use RawTrace binding was still on because SetRawTracedata() did not check the state of the enabled flag in elementBinder.

Randy-Ma:
It appears a bit clearer to me that element binding should not work for such test cases. For example, if the same function is called twice, e.g.:
Level.ByElevationAndName(10, "level0");
Level.ByElevationAndName(20, "level1");
The second call will try to bind with the level created in the first call, which is not usually how element binding is supposed to work. Element binding is working in the context that two calls should not interact with each other.

So as I have said, your fix to check whether element binder is on or off looks ok to me.

mjkkirschner:
I've run all the RevitNodetests and there was one regression, which I also fixed in this PR - to get the build back on track I will merge this in and cross merge to 2016 and 2017 today.

PR #669, opened 2015-09-29
Add a test case when an updated graph is not run again

Randy-Ma:
### Purpose

This is to add a test case for MAGN-8367 when an updated graph is not run again.

### Declarations

### Reviewers

@junmendoza


junmendoza:
@Randy-Ma  thanks for adding the test

PR #673, opened 2015-10-01
Fix list duplication in node Revit>Selection>Views

dimven:
### Purpose

Currently, every time you select the drop-down list the views are added over and over to it. So after a few times of that, you end up with an incredibly long list of duplicate values and can't easily find the view you're looking for.

### Declarations

### FYIs

I don't believe we should add any new tests because no new functionality was exposed.


RodRecker:
@dimven Thank you!
@mjkkirschner Can you take a look or reassign to someone more appropriate?

mjkkirschner:
@dimven thanks for this PR, can you also create PRs for 2016 and 2017? Then I will merge them all at once.

dimven:
@mjkkirschner Sorry about the long delay. This completely slipped my mind after my holiday.

I was not able to add all three branch changes under a single PR and had to create two separate PRs. Is that the general practice, or have I fumbled something?

mjkkirschner:
3 PRs is the way, I'll take a look asap.

PR #680, opened 2015-10-05
Fix MAGN-8407 Element.GetParameterValueByName throws error if trying to access some of the parameters

Randy-Ma:
### Purpose

This is to Fix MAGN-8407. Currently in the implementation of GetParameterValueByName, different types of value are returned based on the storage type of the first found parameter. If the storage type is element ID, the wrapper element will be returned. However not all element IDs here can be converted to elements, specifically when element IDs are negative integers.

The fix here is based on suggestions from an Revit expert. When the value of an ID is negative, the parameter's ID will be checked to see whether it is for category, if so the negative ID will be converted to a proper category.

The limitation here is that not all possible parameter ID has been checked and handled. For other cases, AsValueString will be called to return a string as a minimal implementation and we can still adjust it in the future.

### Declarations

### Reviewers

@pboyer 

### FYIs

@monikaprabhu


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -327,14 +327,31 @@ public object GetParameterValueByName(string parameterName)
...
-                    // When the element is obtained here, to convert it to our element wrapper, it
-                    // need to be figured out whether this element is created by us. Here the existing
-                    // element wrappers will be checked. If there is one, its property to specify
-                    // whether it is created by us will be followed. If there is none, it means the
-                    // element is not created by us.
-                    var ele = ElementIDLifecycleManager<int>.GetInstance().GetFirstWrapper(id) as Element;
-                    result = ElementSelector.ByElementId(id, ele == null ? true : ele.IsRevitOwned);
+                    int paramId = param.Id.IntegerValue;
+                    int valueId = param.AsElementId().IntegerValue;
+                    if (valueId > 0)
```
sharadkjaiswal: I think this check shouldn't be based on -ve value of elementId, rather we should check for parameterId, and if the parameter belongs to Category, then we should return specific category.

Randy-Ma: But this will mean for any cases we will take several more CPU cycles to check the parameter ID.

pboyer:
lgtm!

PR #681, opened 2015-10-05
MAGN-8735 Re-enable Visualization tests and move into shared assembly. Update job on CI to target new visualization assembly. 2015

ikeough:
This is the Revit counterpart to the original PR, https://github.com/DynamoDS/Dynamo/pull/5431

Cherry Picks:

- [ ] Revit2016
- [ ] Revit2017


src/DynamoRevit/ViewModel/DynamoRevitViewModel.cs
```diff
@@ -46,6 +45,14 @@ public static DynamoRevitViewModel Start(StartConfiguration startConfiguration)
                     throw new Exception("An instance of RevitDynamoViewModel is required to construct a DynamoRevitViewModel.");
```
mjkkirschner: I know this is unrleated- but isn't this exception incorrect should it say ```("An instance of RevitDynamoModel is required to construct a DynamoRevitViewModel.");```

?

PR #683, opened 2015-10-06
MAGN-8492 Small curve causes Dynamo to crash when showing background 3D preview in Revit. 2016

ikeough:
### Purpose

This PR addresses [MAGN-8492](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8492). Revit cannot draw curves smaller than 1/64". We add a check whether the curve to be drawn is shorter than `Application.ShortCurveTolerance`. If so, we skip it.

### Declarations

### Reviewers

@mjkkirschner 

### Cherry Picks:
- [ ] Revit2015
- [ ] Revit2017

### DIFF
--- a/src/DynamoRevit/ViewModel/DynamoRevitViewModel.cs
+++ b/src/DynamoRevit/ViewModel/DynamoRevitViewModel.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-
 using Autodesk.Revit.DB;
 
 using Dynamo.Applications.Models;
@@ -28,8 +27,10 @@ namespace Dynamo.Applications.ViewModel
 
             if (RevitWatch3DViewModel.GetTransientDisplayMethod() != null)
             {
-                var watch3DParams = new Watch3DViewModelStartupParams(model, this, "Revit Background Preview");
-                watch3DParams.RenderPackageFactory = new DefaultRenderPackageFactory();
+                var watch3DParams = new Watch3DViewModelStartupParams(model, this, "Revit Background Preview")
+                {
+                    RenderPackageFactory = new DefaultRenderPackageFactory()
+                };
                 Watch3DViewModels.Add(RevitWatch3DViewModel.Start(watch3DParams));
             }
         }
--- a/src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
+++ b/src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
@@ -104,6 +104,8 @@ namespace Dynamo.Applications.ViewModel
 
         private void Draw()
         {
+            if (!Active) return;
+
             var graphicItems = model.CurrentWorkspace.Nodes
                 .Where(n => n.IsVisible)
                 .SelectMany(n => n.GeneratedGraphicItems(engineManager.EngineController));
@@ -210,20 +212,18 @@ namespace Dynamo.Applications.ViewModel
             curve.Tessellate(pkg, renderPackageFactory.TessellationParameters);
 
             // get necessary info to enumerate and convert the lines
-            var lineCount = pkg.LineVertexCount * 3 - 3;
+            //var lineCount = pkg.LineVertexCount * 3 - 3;
             var verts = pkg.LineStripVertices.ToList();
 
             // we scale the tesselation rather than the curve
             var conv = UnitConverter.DynamoToHostFactor(UnitType.UT_Length);
 
-            // add the revit Lines to geometry collection
-            for (var i = 0; i < lineCount; i += 3)
+            var scaledXYZs = new List<XYZ>();
+            for (var i = 0; i < verts.Count; i+= 3)
             {
-                var xyz0 = new XYZ(verts[i] * conv, verts[i + 1] * conv, verts[i + 2] * conv);
-                var xyz1 = new XYZ(verts[i + 3] * conv, verts[i + 4] * conv, verts[i + 5] * conv);
-
-                result.Add(Line.CreateBound(xyz0, xyz1));
+                scaledXYZs.Add(new XYZ(verts[i] * conv, verts[i + 1] * conv, verts[i + 2] * conv));
             }
+            result.Add(PolyLine.Create(scaledXYZs));
 
             return result;
         }
--- a/test/Libraries/RevitIntegrationTests/RevitWatch3DViewModelTests.cs
+++ b/test/Libraries/RevitIntegrationTests/RevitWatch3DViewModelTests.cs
@@ -1,0 +1,35 @@
+﻿using System.IO;
+using System.Linq;
+using Dynamo.Applications.ViewModel;
+using NUnit.Framework;
+using RevitTestServices;
+using RTF.Framework;
+
+namespace RevitSystemTests
+{
+    [TestFixture]
+    public class RevitWatch3DViewModelTests : RevitSystemTestBase
+    {
+        [SetUp]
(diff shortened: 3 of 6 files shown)


test/Libraries/RevitIntegrationTests/RevitWatch3DViewModelTests.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace RevitSystemTests
+{
+    [TestFixture]
+    public class RevitWatch3DViewModelTests : RevitSystemTestBase
+    {
+        [SetUp]
+        public void Setup()
+        {
+            var vm3D = ViewModel.Watch3DViewModels.FirstOrDefault(vm => vm is RevitWatch3DViewModel);
+            vm3D.Active = false;
```
Randy-Ma: I am curious why this is set again as it has been set in StartDynamo.

test/Libraries/RevitIntegrationTests/RevitWatch3DViewModelTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        [Test, TestModel(@".\empty.rfa")]
+        public void SmallCurve_Draw_DoesNotThrowException()
+        {
+            var vm3D = ViewModel.Watch3DViewModels.FirstOrDefault(vm => vm is RevitWatch3DViewModel);
+            Assert.NotNull(vm3D);
```
Randy-Ma: Shall we set vm3D.Active = true for this test case?

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
```diff
@@ -221,7 +223,10 @@ private IEnumerable<GeometryObject> Tessellate(Curve curve)
             {
                 var xyz0 = new XYZ(verts[i] * conv, verts[i + 1] * conv, verts[i + 2] * conv);
                 var xyz1 = new XYZ(verts[i + 3] * conv, verts[i + 4] * conv, verts[i + 5] * conv);
-
+                if (xyz0.DistanceTo(xyz1) < DocumentManager.Instance.CurrentUIApplication.Application.ShortCurveTolerance)
+                {
+                    continue;
+                }
```
Randy-Ma: If a curve of 1' in length is tessellated into 1000 segments, then the curve won't be visible.

Shall we check the length of the curve, if it is too small then it won't even be faceted? If we do facet it, shall we use Polyline.Create() to create a polyline based on the facted points?

Randy-Ma:
@ikeough 
I am seeing you are addressing an issue which I am also trying to fix. So I have put few comments.

test/Libraries/RevitIntegrationTests/RevitWatch3DViewModelTests.cs
ikeough: You are correct. This was mistakenly set to false.

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
ikeough: Yes, you are correct. The line will disappear. I considered this and figured that this was an acceptable loss. We are talking about lines that are shorter than 0.0013 units in Revit. `Polyline.Create` is a good idea, but I wonder if it is subject to the same limitations. That is, what if someone passes in a tiny little circle?

mjkkirschner: If Revit Polylines can have segments shorter than the tolerance, then discarding initial lines below that tolerance should avoid that problem... correct?

Randy-Ma: Yes, @mjkkirschner has got what I have suggested.

Just to make it more clear:
1). Check the curve's length at the very beginning of the function, if it is less than the tolerance, return from there;
2). Facet the curve;
3). Create one polyline from the faceted points;

src/DynamoRevit/ViewModel/DynamoRevitViewModel.cs
```diff
@@ -1,5 +1,5 @@
 ﻿using System;
-
+using System.Drawing;
```
mjkkirschner: do we require system drawing for some reason here?

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -221,12 +222,18 @@ protected override void StartDynamo(TestSessionConfiguration testConfig)
 
                 Model = DynamoRevit.RevitDynamoModel;
 
-                this.ViewModel = DynamoViewModel.Start(
+                this.ViewModel = DynamoRevitViewModel.Start(
```
mjkkirschner: was this a previous mistake somewhere or has something changed fundamentally?

mjkkirschner:
@ikeough I will leave the decision up to you if we should create a more robust, but complex algorithm for short curve rejection in this pr, otherwise I just have one small comment and a question.

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
```diff
@@ -210,20 +212,18 @@ private IEnumerable<GeometryObject> Tessellate(Curve curve)
...
+            var scaledXYZs = new List<XYZ>();
+            for (var i = 0; i < verts.Count; i+= 3)
             {
-                var xyz0 = new XYZ(verts[i] * conv, verts[i + 1] * conv, verts[i + 2] * conv);
-                var xyz1 = new XYZ(verts[i + 3] * conv, verts[i + 4] * conv, verts[i + 5] * conv);
-
-                result.Add(Line.CreateBound(xyz0, xyz1));
+                scaledXYZs.Add(new XYZ(verts[i] * conv, verts[i + 1] * conv, verts[i + 2] * conv));
             }
+            result.Add(PolyLine.Create(scaledXYZs));
```
ikeough: @Randy-Ma @mjkkirschner As it turns out, `Polyline.Create` is incredibly tolerant of small lines. It does not seem subject, at least when created lines for the background preview, to the short curve tolerance normally applied when drawing other types of curves in Revit. For this reason, I've switched this entire method of drawing curves, including straight curves, over to using that method.

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
ikeough: That's a good question. I am not sure why this was done originally, but it seemed that we should be testing the way that a user would be running, using `DynamoRevitViewModel`. It is possible that when we run this through testing, the reason for this change will become obvious.

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
```diff
@@ -104,6 +104,8 @@ protected override void OnNodePropertyChanged(object sender, PropertyChangedEven
 
         private void Draw()
         {
+            if (!Active) return;
```
mjkkirschner: nice

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
mjkkirschner: @ikeough cool - so there is no need to check the length of the polyline?

mjkkirschner:
@ikeough this looks good to merge pending one question

PR #688, opened 2015-10-08
Implement MAGN-8462 Implement a framework to support scenarios involving asynchronous tasks

Randy-Ma:
### Purpose

This is to add a test fixure which will run test cases in a way the tasks are scheduled asyncronously.

AsyncRunningTests is the new fixture which can run test cases in a way the tasks are scheduled asyncronously. Originally when scheduling tasks, the task will run immediately if it is in test mode. This is useful in most cases. However, if there are some tasks which will trigger a document modified event, which in turn will trigger a new transaction, there will be issues because in some event handler, a new transaction can not be started.

As such, this fixture is implemented which is more similar to the real situation where the tasks are scheduled to run asyncronously.

### Declarations

### Reviewers

@Benglin

### DIFF
--- a/test/Libraries/RevitIntegrationTests/AsynchronousModeTests.cs
+++ b/test/Libraries/RevitIntegrationTests/AsynchronousModeTests.cs
@@ -1,0 +1,65 @@
+﻿using System.IO;
+using System.Linq;
+using Autodesk.Revit.DB;
+using Dynamo.Core.Threading;
+using DSCoreNodesUI.Input;
+using NUnit.Framework;
+using Revit.Elements;
+using RevitServices.Persistence;
+using RevitTestServices;
+using RTF.Framework;
+using Dynamo.Nodes;
+
+namespace RevitSystemTests
+{
+    /// <summary>
+    /// Test cases in this fixture will be supposed to run in an asynchronous mode
+    /// </summary>
+    [TestFixture]
+    public class AsynchronousModeTests : RevitSystemTestBase
+    {
+        /// <summary>
+        /// Override the implementation from the base class so that the scheduler
+        /// will run the tasks asyncoronously
+        /// </summary>
+        protected override TaskProcessMode RevitTaskProcessMode
+        {
+            get
+            {
+                return TaskProcessMode.Asynchronous;
+            }
+        }
+
+        /// <summary>
+        /// Run the model but will wait for the asynchronous tasks to be completed
+        /// </summary>
+        private void AsyncRunCurrentModel()
+        {
+            RunCurrentModel();
+
+            // Run till all asyncronous tasks are completed
+            var scheduler = ViewModel.Model.Scheduler;
+            while (scheduler.ProcessNextTask(false)) { }
+        }
+
+        [Test]
+        [Category("RegressionTests")]
+        [TestModel(@".\empty.rvt")]
+        public void MAGN_8187()
+        {
+            string samplePath = Path.Combine(workingDirectory, @".\Async\MAGN_8187.dyn");
+            string testPath = Path.GetFullPath(samplePath);
+
+            ViewModel.OpenCommand.Execute(testPath);
+
+            AssertNoDummyNodes();
+
+            AsyncRunCurrentModel();
+
+            //ensure that the level count is 3
+            var levelColl = new FilteredElementCollector(DocumentManager.Instance.CurrentUIDocument.Document);
+            levelColl.OfClass(typeof(Autodesk.Revit.DB.Level));
+            Assert.AreEqual(levelColl.ToElements().Count(), 3);
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


src/DynamoRevit/ViewModel/DynamoRevitViewModel.cs
```diff
@@ -12,110 +12,110 @@
...
-            model.RevitContextAvailable -= model_RevitContextAvailable;
-            model.RevitContextUnavailable -= model_RevitContextUnavailable;
-            model.RevitDocumentLost -= model_RevitDocumentLost;
-            model.RevitViewChanged -= model_RevitViewChanged;
-            model.InvalidRevitDocumentActivated -= model_InvalidRevitDocumentActivated;
-
-            base.UnsubscribeAllEvents();
-        }
-    }
+	public class DynamoRevitViewModel : DynamoViewModel
```
ikeough: This is some crazy indentation. Is this on purpose?

Randy-Ma: No, it is not on purpose. Let me check what happened.

Randy-Ma: I do not need to change this file at all. I will exclude this one.

test/Libraries/RevitIntegrationTests/AsyncRunningTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using RTF.Framework;
+using Dynamo.Nodes;
+
+namespace RevitSystemTests
+{
+    /// <summary>
+    /// Test cases in this fixture will be supposed to run in an asyncronous mode
+    /// </summary>
+    [TestFixture]
+    public class AsyncRunningTests : RevitSystemTestBase
```
Benglin: A slightly better name would be `AsynchronousTests` :)

test/Libraries/RevitIntegrationTests/AsyncRunningTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using Revit.Elements;
+using RevitServices.Persistence;
+using RevitTestServices;
+using RTF.Framework;
+using Dynamo.Nodes;
+
+namespace RevitSystemTests
+{
+    /// <summary>
+    /// Test cases in this fixture will be supposed to run in an asyncronous mode
```
Benglin: Perhaps rephrase as: `Test cases in this fixture will run asynchronously`?

test/Libraries/RevitIntegrationTests/AsyncRunningTests.cs
```diff
@@ -0,0 +1,65 @@
...
+            get
+            {
+                return TaskProcessMode.Asynchronous;
+            }
+        }
+
+        /// <summary>
+        /// Run the model but will wait for the asyncronous tasks to be completed
+        /// </summary>
+        public void AsyncRunCurrentModel()
```
Benglin: Shouldn't this be a `private` method?

test/Libraries/RevitIntegrationTests/AsyncRunningTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            RunCurrentModel();
+
+            // Run till all asyncronous tasks are completed
+            var scheduler = ViewModel.Model.Scheduler;
+            while (scheduler.ProcessNextTask(false)) { }
```
Benglin: So this is more like *process tasks when `ProcessNextTask` is called*, rather than *running tests asynchronously*, right? The `XxxAsyncTask` basically executes on the test thread instead of *asynchronously* on a separate thread. I see how this will resolve document-event interlock issue, but maybe it is not in the real sense of *asynchronous execution* :)

test/Libraries/RevitIntegrationTests/AsyncRunningTests.cs
Randy-Ma: It is the scheduled tasks that are going to run asynchronously. That is the reason I am saying "run in an asyncronous mode".

test/Libraries/RevitIntegrationTests/AsyncRunningTests.cs
Randy-Ma: How about AsynchronousModeTests? :) The tests are still running synchronously.

Benglin:
Hi @Randy-Ma, thanks for fixing this. Other than few renaming requests, this basically looks good. Feel free to merge it in after renaming them.

test/Libraries/RevitIntegrationTests/AsyncRunningTests.cs
Randy-Ma: Yes

Randy-Ma:
@Benglin 
Thanks a bunch! I have addressed your comments and am merging this now.

PR #710, opened 2015-10-13
Write Version selector data to Registry

sharadkjaiswal:
### Purpose
- Fixes MAGN-8747: Default Dynamo version should 0.9.0 once it is installed.

### Problem
 Earlier the version selector data was saved in %appdata%\DynamoDllForLoad_2014.txt file. So when the newer version of Dynamo is installed, the version selector finds this file and reads the last version and shows up in the split button. We either need to delete this file at the time of installation or keep track of first run after the installation.

### Solution
  Instead of using local file to keep version specific data, I moved this data to registry of the current Dynamo version. Version selector can read this information from registry when the version selector app is loaded in Revit and update the Default Dynamo version in split button accordingly.
  Defined a new class VersionSelectorData to keep the selected version data to be stored in registry under Dynamo product information key. When product gets uninstalled, the registry data will automatically be cleaned.

### Declarations

### Reviewers

@Randy-Ma

### DIFF
--- a/src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
+++ b/src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
@@ -14,6 +14,7 @@ using Autodesk.Revit.DB;
 using Autodesk.Revit.UI;
 
 using DynamoRevitVersionSelector.Properties;
+using Microsoft.Win32;
 
 namespace Dynamo.Applications
 {
@@ -54,18 +55,9 @@ namespace Dynamo.Applications
             uiApplication = application;
             // now we have a default path, but let's look at
             // the load path file to see what was last selected
-            var cachedPath = String.Empty;
             var revitVersion = application.ControlledApplication.VersionNumber;
-            var fileLoc = Utils.GetVersionSaveFileLocation(revitVersion);
-
-            if (File.Exists(fileLoc))
-            {
-                using (var sr = new StreamReader(fileLoc))
-                {
-                    cachedPath = sr.ReadToEnd().TrimEnd('\r', '\n');
-                }
-            }
-
+            var selectorData = VersionSelectorData.ReadFromRegistry(revitVersion);
+            
             var revitFolder =
                 new DirectoryInfo(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location));
 
@@ -74,17 +66,21 @@ namespace Dynamo.Applications
 
             Products = new List<DynamoProduct>();
             int index = -1;
+            var selectedVersion = selectorData.SelectedVersion.ToString(2);
             foreach (var p in dynamoProducts)
             {
                 var path = VersionLoader.GetDynamoRevitPath(p, revitVersion);
                 if (!File.Exists(path))
                     continue;
 
-                if (path.Equals(cachedPath))
+                if (p.VersionInfo.ToString(2) == selectedVersion)
                     index = Products.Count;
 
                 Products.Add(p);
             }
+
+            if (index == -1)
+                index = Products.Count - 1;
 
             // If there are multiple versions installed, then create
             // a couple of push buttons in a panel to allow selection of a version.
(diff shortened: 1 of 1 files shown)


src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -54,18 +55,9 @@ public Result OnStartup(UIControlledApplication application)
             uiApplication = application;
             // now we have a default path, but let's look at
             // the load path file to see what was last selected
```
Randy-Ma: Minor, but the comments need to be updated.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -358,45 +353,72 @@ public Result Execute(ExternalCommandData commandData, ref string message, Eleme
...
-            if (File.Exists(path))
-            {
-                File.Delete(path);
-            }
-
-            using (var tw = new StreamWriter(path))
-            {
-                tw.WriteLine(loadPath);
-            }
+        const string RegKey64 = @"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\";
```
Randy-Ma: I think we need a different location.

Randy-Ma: I see, the data stored here are going to be cleaned after uninstallation.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -74,18 +66,22 @@ public Result OnStartup(UIControlledApplication application)
...
             int index = -1;
+            var selectedVersion = selectorData.SelectedVersion.ToString(2);
             foreach (var p in dynamoProducts)
             {
                 var path = VersionLoader.GetDynamoRevitPath(p, revitVersion);
                 if (!File.Exists(path))
                     continue;
 
-                if (path.Equals(cachedPath))
+                if (p.VersionInfo.ToString(2) == selectedVersion)
```
Randy-Ma: It is better to make it consistent to take 3 fields like what is done in the AddSplitButtonGroup function. Or is this done for a reason?

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -358,45 +353,72 @@ public Result Execute(ExternalCommandData commandData, ref string message, Eleme
...
         {
-            try
-            {
-                string appDataFolder = System.Environment.GetFolderPath(
-                    System.Environment.SpecialFolder.ApplicationData);
+            var data = new VersionSelectorData() { RevitVersion = revitVersion };
+            var regKey = RegistryKey.OpenBaseKey(RegistryHive.LocalMachine, RegistryView.Registry64);
+            regKey = regKey.OpenSubKey(RegKey64);
+            
+            var key = string.Format(@"Dynamo {0}.{1}", data.ThisDynamoVersion.Major, data.ThisDynamoVersion.Minor);
```
Randy-Ma: What is the benefit for a structure like:
```
Dynamo {major version}.{minor version}
|----{RevitVersion}
|----{RevitVersion}
```
There might be one entry every time a different version of Dynamo is installed.

Randy-Ma:
This is great! The version selector is more flexible than before. One thing to check is after a new version is installed, we'd better remove the obsoleted entries in the registry. Apart from this and few comments, LGTM

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
sharadkjaiswal: Currently, we can't have 0.9.0 and 0.9.1 installed side-by-side, hence no point comparing the 3rd field.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
sharadkjaiswal: Agree, there will be a different entry for every version of Dynamo installed, but these will be removed when Dynamo gets uninstalled.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
Randy-Ma: Thanks! This makes sense!

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
Randy-Ma: Yes, thanks

PR #718, opened 2015-10-14
Remove usage of obsolete API for Revit NurbSpline curve creation

sharadkjaiswal:
### Purpose

Autodesk.Revit.DB.NurbSpline.Create has been marked as obsolete and also call of this API is throwing exception in Revit2017 tests.

The call of NurbSpline.Create method is now replaced with a new utility method CreateNurbSpline which calls correct API based on whether the curve is rational or not.

### Declarations

### Reviewer

- [ ] @Randy-Ma

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitCurve.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitCurve.cs
@@ -100,7 +100,7 @@ namespace Revit.GeometryConversion
             {
                 using (var converted = NurbsUtils.ElevateBezierDegree(crv, 3))
                 {
-                    return Autodesk.Revit.DB.NurbSpline.Create(converted.ControlPoints().ToXyzs(false),
+                    return CreateNurbSpline(converted.ControlPoints().ToXyzs(false),
                             converted.Weights(),
                             converted.Knots(),
                             converted.Degree,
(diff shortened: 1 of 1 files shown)


src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitCurve.cs
```diff
@@ -141,6 +141,13 @@ private static Autodesk.Revit.DB.Curve Convert(Autodesk.DesignScript.Geometry.Nu
                 crv.IsRational);
 
         }
+        private static Autodesk.Revit.DB.Curve CreateNurbSpline(IList<XYZ> controlPoints, IList<double> weights, IList<double> knots, int degree, bool closed, bool rational)
```
Randy-Ma: It looks like the closed parameter is not used, maybe because it can be induced from knots, weights and degree.

Randy-Ma:
Apart from one comment which I am not quite sure, LGTM

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitCurve.cs
sharadkjaiswal: Here is what the Obsolete message says:

`[Obsolete("This function will be obsolete in Revit 2017. Use NurbSpline.CreateCurve(degree, knots, controPoints, weights) instead")]
        public static NurbSpline Create(IList<XYZ> controlPoints, IList<double> weights, IList<double> knots, int degree, bool closed, bool rational);`

PR #719, opened 2015-10-14
Fix MAGN-8852 Label is missing on the toggle option "Revit background preview on"

Randy-Ma:
### Purpose

This submission is to fix an issue that the menu item to toggle one background preview has not text.

The reason is that in the en-US resources, the string resource is missing. I am adding it here in this pull request.

### Declarations

### Reviewers

@aparajit-pratap

### DIFF
--- a/src/DynamoRevit/Properties/Resources.en-US.resx
+++ b/src/DynamoRevit/Properties/Resources.en-US.resx
@@ -151,4 +151,7 @@
   <data name="NewDocument" xml:space="preserve">
     <value>a new document.</value>
   </data>
+  <data name="BackgroundPreviewName" xml:space="preserve">
+    <value>Revit Background Preview</value>
+  </data>
 </root>


aparajit-pratap:
@Randy-Ma I think you need to change the `Resources.resx` file as well?

src/DynamoRevit/Properties/Resources.en-US.resx
```diff
@@ -151,4 +151,7 @@
   <data name="NewDocument" xml:space="preserve">
     <value>a new document.</value>
   </data>
+  <data name="BackgroundPreviewName" xml:space="preserve">
+    <value>Revit Background Preview</value>
+  </data>
```
aparajit-pratap: Maybe you could also add a comment like: `<comment>The name of the 3D background preview.</comment>`

Randy-Ma:
@aparajit-pratap Thanks for the reminder! But for that one, it is already there.

PR #725, opened 2015-10-14
Remove dependency on ExternalCommandData

sharadkjaiswal:
### Purpose
- For Revit2017 integration we need to pass command data differently.
- Hence replace the dependency based on new class DynamoRevitCommandData
modeled after ExternalCommandData.

### Reviewer

- [ ] @Randy-Ma


Randy-Ma:
LGTM

PR #730, opened 2015-10-15
Encapsulate DynamoButton field

sharadkjaiswal:
### Purpose
- Define DynamoButtonEnabled property on DynamoRevitApp class to encapsulate DynamoButton field.
- Align the code refactoring expected for Revit2017 #732 

### Reviewer
@Randy-Ma

### DIFF
--- a/src/DynamoRevit/DynamoRevitApp.cs
+++ b/src/DynamoRevit/DynamoRevitApp.cs
@@ -36,7 +36,7 @@ namespace Dynamo.Applications
         public static ControlledApplication ControlledApplication;
         public static UIControlledApplication UIControlledApplication;
         public static List<IUpdater> Updaters = new List<IUpdater>();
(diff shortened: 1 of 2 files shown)


src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -250,5 +250,11 @@ public static Assembly ResolveAssembly(object sender, ResolveEventArgs args)
                 throw new Exception(string.Format("The location of the assembly, {0} could not be resolved for loading.", assemblyPath), ex);
             }
         }
+
+        public static bool DynamoButtonEnabled
```
Randy-Ma: If this property is not supposed to be used by other modules, s\public\internal.

Randy-Ma:
One minor comment. But in general, LGTM

PR #732, opened 2015-10-15
Use command binding for Dynamo command

sharadkjaiswal:
### Purpose

This submission makes use of command binding to register Dynamo, so that Dynamo button will appear in manage tab in Revit 2017. 

If there are multiple version of Dynamo installed on the system then a version selection dialog will be shown as follows for users to choose a specific version. Once a version is selected, the VersionLoader extension application unregisters the command binding and DynamoRevitApp will do the command binding for the actual DynamoRevit command.

image

This also includes the refactoring done in #730 

### Declarations

### Reviewer
- [ ] @ikeough 
- [ ] @Randy-Ma  

### UX Reviewer

Please review the Task Dialog content as well as UX.
- [ ] @Racel 
- [ ] @lillismith 

### FYI
- [ ] @riteshchandawar 
- [ ] @kronz


Randy-Ma:
Are test cases still passing although it looks like that they will work?
Also we might need some changes to the addin which will not point to the version selector module.

Randy-Ma:
Other than these, it looks good.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -109,104 +95,125 @@ public Result OnStartup(UIControlledApplication application)
...
 
-            BitmapSource bitmapSource =
-                Imaging.CreateBitmapSourceFromHBitmap(
-                    dynamoIcon.GetHbitmap(),
-                    IntPtr.Zero,
-                    Int32Rect.Empty,
-                    BitmapSizeOptions.FromEmptyOptions());
+        void canExecute(object sender, CanExecuteEventArgs e)
+        {
+            e.CanExecute = e.ActiveDocument != null;
```
Randy-Ma: When will e.ActiveDocument == null? Is it if there are no documents opened, then this addin will not run? But then when will this addin get a chance to run later?

Randy-Ma: Okay, I assume there are no documents opened.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -109,104 +95,125 @@ public Result OnStartup(UIControlledApplication application)
...
+            // task dialog will show a Close button by default
+            mainDialog.CommonButtons = TaskDialogCommonButtons.Close;
+            mainDialog.DefaultButton = defaultResult;
+
+            TaskDialogResult tResult = mainDialog.Show();
+            if (tResult == TaskDialogResult.Close)
+                return null;
+
+            var index = (int)tResult - (int)TaskDialogCommandLinkId.CommandLink1;
+            return Products[index];
```
Randy-Ma: s\Products\products\

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
Randy-Ma: So I agree that the logic here may not be the same with SetRunEnabledBasedOnContext depending on the workflow.

Randy-Ma:
Thanks! It looks good.

PR #735, opened 2015-10-21
Fix Revit test failures

sharadkjaiswal:
### Purpose

After removing the obsolete method for NurbsCurve conversion in #729, couple of tests started failing because the new method called for NurbsCurve creation tries to simplify the geometry and hence for curve created with colinear points, it returns Line. This PR updates the tests accordingly.

### Reviewer
- [ ] @Randy-Ma


Randy-Ma:
Thanks! LGTM

PR #736, opened 2015-10-21
Resource strings for version selection

sharadkjaiswal:
### Purpose

Cross merge the string resource changes at #732 from Revit 2017 to 2015 branch.

### Reviewer
- [ ] @Randy-Ma


Randy-Ma:
Are these resource strings required?

sharadkjaiswal:
Not required in Revit2015 and Revit2016 branches, but the current localization process localizes strings from one branch and uses the same for all. So we need to always keep the resource dlls in sync across different branches.

Randy-Ma:
I see. Thanks! It looks good.

PR #758, opened 2015-11-03
Implement MAGN-3427 Update Adaptive Component Placement to use new Batch creation method

Randy-Ma:
### Purpose

The purpose of this pull request is to modify three existing nodes for creating adaptive components in batch. To create a number of adaptive components, originally they have to be created one by one by various lacing strategies. Now NewFamilyInstanceCreationData can be used for adaptive components. A number of creation data can be prepared first and the adaptive components can be created in batch by NewFamilyInstances2.

Here is one example using AdaptiveComponent.ByPoints to explain this. If 1000 adaptive components are going to be created, originally AdaptiveComponent.ByPoints will be called 1000 times, but now it will only be called one time and all the 1000 components will be created in one function call.

Except for the new override for NewFamilyInstanceCreationData mentioned earlier which make it possible to prepare creation data, the function signature of AdaptiveComponent.ByPoints is also changed from:
```
AdaptiveComponent ByPoints(Point[] points, FamilyType familyType)
```
to:
```
AdaptiveComponent[] ByPoints(Point[][] points, FamilyType familyType)
```
Thus it is possible to pass points for a number of adaptive components and use them in one function call.

Similarly, the function signature of the old two other nodes are changed respectively from:
```
AdaptiveComponent ByParametersOnFace(UV[] uvs, Surface surface, FamilyType familyType)
AdaptiveComponent ByParametersOnCurveReference(double[] parameters, Curve curve, FamilyType familyType)
```
to:
```
AdaptiveComponent[] ByParametersOnFace(UV[][] uvs, Surface surface, FamilyType familyType)
AdaptiveComponent[] ByParametersOnCurveReference(double[][] parameters, Curve curve, FamilyType familyType)
```

The following image illustrates the new nodes. The parts with red circles identify the modified parts: the input parameter is modified to have one more dimension and the output is also modified to array.
image

##### One Sample
For example to place the following adaptive component around a semi-sphere: (This is a pretty simple adaptive component family which is composed of four bars and has 4 adaptive points)
image

The used node is the new ByParametersOnFace node. A semi-sphere surface is divided into a number of parts according to u and v equally. Each part has four points. These four points will be used to place one adaptive component. The result is:
image

##### Element Binding
The element binding is implemented in a way to reuse the old elements as much as possible. If there are not enough elements, then new elements will be created based on the remaining data in batch. The reused elements will be updated to the new points one by one.

##### Migration of Old Graphs
Right now the old nodes will be migrated automatically respectively based on their signatures from:
```
AdaptiveComponent ByPoints(Point[] points, FamilyType familyType)
AdaptiveComponent ByParametersOnFace(UV[] uvs, Surface surface, FamilyType familyType)
AdaptiveComponent ByParametersOnCurveReference(double[] parameters, Curve curve, FamilyType familyType)
```
to:
```
AdaptiveComponent[] ByPoints(Point[][] points, FamilyType familyType)
AdaptiveComponent[] ByParametersOnFace(UV[][] uvs, Surface surface, FamilyType familyType)
AdaptiveComponent[] ByParametersOnCurveReference(double[][] parameters, Curve curve, FamilyType familyType)
```
If the old nodes have lacing stragies used, after migration, the node may not work or may not work as expected. This will be further described in the "Known Issues" part.

##### Performance Improvement
The performance improvement is not apparent if only a few number of adaptive components are created. But in a case where a large number of adaptive components are created, the observed time taken is 18 min and 10 min for the old nodes and new nodes. The new nodes do have performance advantage over the old ones.

##### Compatibility Consideration
Originally there are some hidden overrides for each of the three nodes to make it more compatible and easy to use. Refer https://github.com/DynamoDS/Dynamo/pull/2414 for details. These overrides are reserved but some similar modification to parameters and return value as the three visible nodes are implemented.

##### Usability Consideration
If only one adaptive component is created, the element will be returned, instead of an array of one element.

##### Known Issues
After the old nodes are migrated, if there are lacing strategies used, the graph may not work or work as expected. Right now two cases have been found:
- the old nodes have a longest or cross lacing strategy, the nodes may not work or work as expected after migration. For this reason, one of the test case called AdaptiveComponent is temporarily marked as Failure.
- the old nodes have a shortest lacing strategy, the nodes may return different result. For this reason, one of the test case called AdaptiveComponentByFace is modified so that the return data is a one dimensional array.

### Declarations

### Reviewers

@sharadkjaiswal 

### FYIs

@kronz @mccrone @ikeough @mjkkirschner


kronz:
@Randy-Ma Having 3 additional nodes, each with slightly different rules for batch versus regular creation will be very difficult for users to reason about.  Is it a requirement of the batch processing method that the user declare the number of placement points?  This difference alone will make it very difficult for users to understand what they need to do.  We need to keep the same input for batch or regular creation if at all possible, and I would like to keep the options for Adaptive Component creation to just 3 nodes.  I had suggested in MAGN-3427 that we keep the same original 3 nodes and make a boolean toggle to engage the batch method.  Is this possible?

Another question.  Usually with Revit family placements, CREATION is expensive, but MODIFICATION is relatively cheap.  Can you do a comparison in something like 1000 placements to see how delete/create-with-batch method compares with a modify-with-regular method? It MAY be that using batch to create, and regular to modify is the most performative.

Randy-Ma:
@kronz 
Thanks for the suggestions and quick response! If we remove or modify the existing nodes, we need to handle migration carefully because it may break exsting graphs which are still using the old nodes. If we want to modify the original three nodes, except for adding a boolean toggle, we need  to change the input to support a list of adaptive components and then we can handle these in batch. 
If you agree with that we can modify the input points, uvs or parameters from one dimensional array to two dimensional array, then I will keep only one set of nodes with an additional input to toggle between them.

I am in the process of adding support for element binding and will update it soon.

andydandy74:
From a user's perspective, I'm 100% with @kronz on this. I would rather *not* see a second set of nodes. 

Also, why is it not possible to deduce the required two dimensional list structure from the way that lists of points/UVs/params are currently fed into the AC nodes? I think the way it works now is pretty intuitive and it would be a shame to give it up when it is even remotely possible that the list structure can be transformed *within* the node if needed.

As a user I would expect a very prominent warning in the node description (think CAPITAL LETTERS ALL THE WAY) If the node does *not* update elements but rather creates new elements in their place. 

Final question: Why would there need to be a toggle between batch creation and normal creation at all? If batch creation is faster with larger numbers of elements and performance is more or less the same for smaller number of elements, why not just get rid of the old creation method *entirely*?

kronz:
Sorry, the conversationis happening now in 2 places, but, yes @andydandy74  we are pursuing this like you recommend (and to anwer your question about why consider a toggle:  sometimes a batching process can take longer for small numbers of elements or be more complicated to set up than a regular process.  In the end we are not pursuing that avenue anyway).  @Randy-Ma let's keep it here, as getting feeback like this from Andreas is really helpful.  Copy/pasted the side conversation below.

Zhaoyang Ma  04 Nov 2015, 18:12
Right now, the part of work to use only one set of nodes are done:
1). AdaptiveComponent.ByPoints, AdaptiveComponent.ByParametersOnFace and AdaptiveComponents.ByParametersOnCurveReference are modified to take one more dimension for uv, point or parameter. 
The return value is changed to an array.
2). Element binding is implemented to reuse old elements created in the last run as much as possible. But the old elements are updated one by one.
3). There are some invisible overloads which are not visible in the library and I have updated them as well.
(Refer https://github.com/DynamoDS/Dynamo/pull/2414 for details)
4). I discussed with Sharad Jaiswal. He has a good suggestion about migration: for each of the node, add an AllowRankReduction attribute, so that if it returns one element before, now it won't return an array which contains only one element. Also as the engine will do rank promotion, it is very likely that we do not need to spend time on migration.
Zhaoyang Ma  04 Nov 2015, 18:18
Zach Kron 
But if you really want to add an additional parameter to toggle to use batch method, I will make the change tomorrow.

Zach Kron  04 Nov 2015, 20:18
Zhaoyang Ma Nope, that is fine to leave out the additional parameter.  That was only another suggestion to use if there was going to be a limitation around element binding, but with modify using the non-batch method, then we can keep the nodes as they are.  I think at this point the best thing to do is get a build out to me and QA to test it for performance improvements.  The questions I have right now are
- does migration happen seamlessly
- can users still use the same structure for passing in data as before
- with batch creation and regular modification, does modify now take LONGER than creation?  If so, we will need to revisit the implementation.

Thanks, sorry this is turning out to be more complicated than we thought.
Z

kronz:
@Randy-Ma looks like your code does not use the scaling tools in the old node?  That is, when referencing the geometry in a metric (mm) Revit file, the Revit preview geometry shows where the geometry SHOULD go, but the adaptive components are scaled improperly.  You can see this with the Revit sample file and the Adaptive component sample dyn file.

Randy-Ma:
Thanks! I have updated the code correspondingly to fix the issue.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -225,100 +54,24 @@ private void InitAdaptiveComponent(Autodesk.Revit.DB.FamilyInstance familyInstan
...
-            IList<ElementId> placePointIds = AdaptiveComponentInstanceUtils.GetInstancePlacementPointElementRefIds(InternalFamilyInstance);
+            IList<ElementId> placePointIds = AdaptiveComponentInstanceUtils.GetInstancePlacementPointElementRefIds(fi);
 
-            if (placePointIds.Count() != points.Count())
+            if (placePointIds.Count() != pnts.Count())
                 throw new Exception(Properties.Resources.InputPointParamsMismatch);
 
-            // Set the position of each placement point
             int i = 0;
             foreach (var id in placePointIds)
```
sharadkjaiswal: It's better to run a loop over i=0 to i < pnts.Count() so that you don't have to think about increment of i inside the loop.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -412,11 +184,28 @@ public static AdaptiveComponent ByParametersOnFace(Autodesk.DesignScript.Geometr
...
-            return new AdaptiveComponent(uvs.Select(x => new[] { x.U, x.V }).ToArray(), ElementFaceReference.TryGetFaceReference(surface), familyType);
+            var result = GeometryObjectSelector.ByReferenceStableRepresentation(
+                        faceReference.InternalReference.ConvertToStableRepresentation(Document));
+            var surface = result as Surface;
+            if (surface == null)
+            {
+                var surfaces = result as List<Surface>;
+                if (surfaces != null && surfaces.Count() > 0)
+                {
+                    surface = surfaces[0];
```
sharadkjaiswal: Will there be a situation that a face has multiple surfaces? Should we throw exception to catch such a scenario?

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -475,7 +273,23 @@ public static AdaptiveComponent ByParametersOnCurveReference(double[] parameters
...
-            return new AdaptiveComponent(parameters, ElementCurveReference.TryGetCurveReference(revitCurve).InternalReference, familyType);
+            var result = GeometryObjectSelector.ByReferenceStableRepresentation(
+                        revitCurve.InternalReference.ConvertToStableRepresentation(Document));
+            var curve = result as Autodesk.DesignScript.Geometry.Curve;
+            if (curve == null)
+            {
+                var curves = result as List<Autodesk.DesignScript.Geometry.Curve>;
+                if (curves != null && curves.Count() > 0)
+                {
+                    curve = curves[0];
```
sharadkjaiswal: What if selected curve reference is exported as multiple curves? Can we throw exception?

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -554,11 +372,136 @@ public static AdaptiveComponent ByParametersOnCurveReference( double[] parameter
...
+        /// </summary>
+        /// <param name="points">a two-dimensional array of points</param>
+        /// <param name="familyType">a family type to use to create the adaptive components</param>
+        /// <returns></returns>
+        private static AdaptiveComponent[] InternalByPoints(Point[][] points, FamilyType familyType)
+        {
+            var oldInstances = ElementBinder.GetElementsFromTrace<Autodesk.Revit.DB.FamilyInstance>(Document);
+            int countToBeCreated = points.Count();
+            int countOfOldInstances = 0;
+            if (oldInstances != null) oldInstances.Count();
```
sharadkjaiswal: the return value of oldInstances.Count() is not set to any variable.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -554,11 +372,136 @@ public static AdaptiveComponent ByParametersOnCurveReference( double[] parameter
...
+            int reusableCount = countToBeCreated > countOfOldInstances ? countOfOldInstances : countToBeCreated;
```
sharadkjaiswal: Do you want to use Min(countToBeCreated, countOfOldInstances) ?

sharadkjaiswal:
Great PR @Randy-Ma, I liked the details you have put in your PR. Thanks for being so meticulous. Apart from some minor comments it's LGTM.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
Randy-Ma: Good question! This is implemented as such because ByReferenceStableRepresentation is returning an object which can be a single element or a list. Right now most probably it is a surface list which contains only one surface.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
Randy-Ma: Nice catch!

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
Randy-Ma: Yeah, I think throwing an exception is a good choice in this case.

Randy-Ma:
@sharadkjaiswal Thanks for reviewing the code! I have addressed your comments and also run tests in some fixtures which I think are most related to adaptive components. The test cases are passing.

PR #767, opened 2015-11-10
Use correct property in property change handler.

ikeough:
### Purpose

This PR fixes https://github.com/DynamoDS/Dynamo/issues/5684. The property changed handler was incorrectly referencing `IsUpdated`, which no longer exists on `NodeModel`.

In addition to fixing the item population error, this PR adds a call to `PopulateItems` in the property change handler, and sets the selected item index to 0 if the control did not previously have a selected item. These two additions together ensure that the control does not look blank, even when the `Items` collection has values.

### Declarations

  - New documentation is added to `PopulateItems` to explain what it does and its side effects.
- [ ] User facing strings, if any, are extracted into `*.resx` files <b>n/a</b>
- [ ] All tests pass using the self-service CI. <b>n/a</b> (Revit only)

### Reviewers

@mjkkirschner 

### Cherry-picks:
- [x] Revit2016
- [x] Revit2017

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitDropDown.cs
+++ b/src/Libraries/RevitNodesUI/RevitDropDown.cs
@@ -134,13 +134,15 @@ namespace DSRevitNodesUI
 
         void OnPropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
         {
-            if (e.PropertyName != "IsUpdated")
+            if (e.PropertyName != "CachedValue")
                 return;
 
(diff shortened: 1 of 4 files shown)


ikeough:
@mjkkirschner The test has been added.

mjkkirschner:
@ikeough should this PR be merged?

ikeough:
@mjkkirschner <slaps forehead> yes. That could be why @jnealb thought this was still broken :smiley_cat:

PR #768, opened 2015-11-12
MAGN-9069

ikeough:
### Purpose

[MAGN-9069](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9069)

### Declarations

- [ ] All tests pass using the self-service CI. <b>n/a</b> Revit only.

### Reviewers

@mjkkirschner 

NOTE: This does not require cherry-picking to other Revit branches as it is an error specific to the Revit API in 2017. We should fix ONLY the test until we can get feedback from the Revit team.

### DIFF
--- a/test/Libraries/RevitIntegrationTests/SelectionTests.cs
+++ b/test/Libraries/RevitIntegrationTests/SelectionTests.cs
@@ -383,8 +383,11 @@ namespace RevitSystemTests
             RunCurrentModel();
 
             // ElementsAtLevel
+            // See http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9069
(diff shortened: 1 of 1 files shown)


mjkkirschner:
@ikeough I took a look at this test as well, this fix looks fine to me, but I'm wondering if the larger issue is that the API was returning rooms incorrectly in 2016. Do the rooms exist in the model in 2016 or 2017 - so that we can see them in Revit UI?

also LGTM

ikeough:
@mjkkirschner The Rooms exist in both Revit 2016 and 2017. In 2016, the Rooms correctly return a level for their `Level` parameter. In 2017, the rooms are still there, but their `Level` parameter returns `HasValue=false`.

PR #769, opened 2015-11-12
MAGN-9070

ikeough:
### Purpose

[MAGN-9070](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9070)

### Declarations

- [ ] All tests pass using the self-service CI. <b>n/a</b> Revit only.

### Reviewers

@mjkkirschner 
  
#### Cherry Picks:  
- [x] Revit2016
- [ ] Revit2015


mjkkirschner:
@ikeough looks good to me, my only question is: will the exception contain the elements that could not be deleted and should we include that info in the message.

ikeough:
@mjkkirschner It's a good question. The problem is, there's not much the user can do with those element ids at that point. I suppose they could use the ids to look at which elements in the model are causing the problem. There also might be A LOT of them. So writing them to the log might not be great.

PR #778, opened 2015-11-26
Fix MAGN-9132 Graph adaptive components node placement does not reexecute after reopen the file and run

Randy-Ma:
### Purpose

This is to fix MAGN-9132 Graph adaptive components node placement does not reexecute after reopen the file and run.

In RevitDynamoModel, when doing reconciliation, it is assuming all ISerializable objects are SerializableId objects and will cast them. But for the batch processing adaptive component nodes, we are using MultipleSerializableId objects. As a result, the cast will fail and throws an exception and causes issues.

The fix here will collect object IDs from SerializableId objects and MultipleSerializableId objects respectively.

### Declarations

### Reviewers

@sharadkjaiswal


src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -206,11 +206,16 @@ public new static RevitDynamoModel Start(IRevitStartConfiguration configuration)
...
                 orphanedSerializables
-                .SelectMany(kvp=>kvp.Value)
-                .Cast<SerializableId>()
-                .Select(sid => sid.IntID).ToList();
+                .SelectMany(kvp => kvp.Value)
+                .Where(x => x is ISerializable);
+
+            orphanedIds.AddRange(serializables.Where(x => x is SerializableId).
+                Cast<SerializableId>().Select(sid => sid.IntID));
+            orphanedIds.AddRange(serializables.Where(x => x is MultipleSerializableId).
```
sharadkjaiswal: Should MultipleSerializableId be derived from SerializableId?? What if we introduce a new ISerializable class in future that are not derived from either SerializableId and MultipleSerializableId??

Randy-Ma: Thanks a bunch for the suggestions! ISerializable only have one member function which can not be used directly. In my humble opinion, other classes can either derive from SerializableId or MultipleSerializableId.

PR #780, opened 2015-12-13
MAGN-9085 Revit Background Preview ON interferes with ui interaction in graph

pbidenko:
### Purpose

When we ungroup n nodes `RaisePropertyChanged("IsVisible")` triggers to each node in group. It caused costly and unnecessary tessellation for each node for n times. So the complexity was `O(n^2)` and it took lot of time for complex graphs. I've managed to reduce time of execution to `O(n)`. 

### Declarations

### Reviewers
@pboyer

### DIFF
--- a/src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
+++ b/src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
@@ -70,7 +70,7 @@ namespace Dynamo.Applications.ViewModel
 
         protected override void OnEvaluationCompleted(object sender, EvaluationCompletedEventArgs e)
         {
-            Draw();
+            if (e.EvaluationTookPlace) Draw();
         }
 
         protected override void OnNodePropertyChanged(object sender, PropertyChangedEventArgs e)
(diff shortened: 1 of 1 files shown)


src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
```diff
@@ -84,20 +84,28 @@ protected override void OnNodePropertyChanged(object sender, PropertyChangedEven
...
         #region private methods
 
-        private void Draw()
+        private void Draw(NodeModel node = null)
         {
             if (!Active) return;
+            IEnumerable<IGraphicItem> graphicItems;
 
-            var graphicItems = model.CurrentWorkspace.Nodes
+            if (node != null)
```
mjkkirschner: should we check that isVisible is true?

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
```diff
@@ -70,7 +70,7 @@ protected override void OnActiveStateChanged()
 
         protected override void OnEvaluationCompleted(object sender, EvaluationCompletedEventArgs e)
         {
-            Draw();
+            if (e.EvaluationTookPlace) Draw();
```
mjkkirschner: does this handle all cases? i.e. if I delete a node will the geometry in Revit preview be deleted?

mjkkirschner:
@pbidenko this looks good, just a few questions and then LGTM

pbidenko:
@mjkkirschner please take another look

mjkkirschner:
LGTM @pbidenko thanks!

PR #783, opened 2015-12-17
MAGN-9218 Freezeing a node deletes revit elements on re-exectution

ramramps:
## Purpose
This PR fixes the issue with Revit elements getting deleted when the graph re-executes. 

This PR refers to the below tasks 
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9218

 ## Declaration

- [] The level of testing this PR includes is appropriate
- [] User facing strings, if any, are extracted into `*.resx` files

## Reviewers
 @mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -187,18 +187,23 @@ namespace Revit.Elements
         protected string InternalUniqueId;
 
         /// <summary>
+        /// Set the element's freeze state. If the node is set to freeze
+        /// all the elements for that node will be set to freeze.
+        /// </summary>
+        public bool IsFrozen = false;
+
+        /// <summary>
         /// Default implementation of dispose that removes the element from the
         /// document
         /// </summary>
         [IsVisibleInDynamoLibrary(false)]
         public virtual void Dispose()
         {
-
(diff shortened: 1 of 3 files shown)


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -187,18 +187,23 @@ protected ElementId InternalElementId
...
         /// </summary>
         [IsVisibleInDynamoLibrary(false)]
         public virtual void Dispose()
         {
-
             // Do not cleanup Revit elements if we are shutting down Dynamo or
-            // closing homeworkspace.
-            if (DisposeLogic.IsShuttingDown || DisposeLogic.IsClosingHomeworkspace)
+            // closing homeworkspace or the element itself is frozen.
+            if (DisposeLogic.IsShuttingDown || DisposeLogic.IsClosingHomeworkspace || IsFrozen)
```
ramramps: If the element is in frozen state then do not delete.

mjkkirschner:
@ramramps this looks good to me, but can you add a test? A revit system test might be needed.

ramramps:
@mjkkirschner - this is mostly the dispose logic. wondering how to write a test for this.

mjkkirschner:
@ramramps what about a system test that freezes a revit constructor node, modifies and upstream input, and then just makes sure that the element exists.

PR #795, opened 2016-01-04
Fix MAGN-9229 Crash after running a sample

Randy-Ma:
### Purpose

This submission is to fix MAGN-9229 that a crash happens when a sample is run.

Actually this is a more general issue that when the material is not created successfully, later a tessellated face can not created successfully with a material id of null and an exception is thrown.

The reason the material is not created successfully is that in some cases, there are no materials named "Glass". This may happen if the string has been localized or if the material has been deleted. 

The fix here is to try to find the material named "Glass". If it is found, a new materail will be created based on it. This is the same as the old implementation. But if the material is not found, a new material will be created.

### Declarations

### Reviewers

@mjkkirschner


mjkkirschner:
@Randy-Ma implementation looks good to me, can you cherry pick this to 2015 and 2017?

PR #813, opened 2016-01-22
Adaptive Components Deleted on Re-run

mjkkirschner:
### Purpose

http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9341

This PR overrides equals and getHashCode on `SerializableID` and `MultiSerializableID`, this is necessary because when we do post Trace reconciliation and attempt to check if we have recreated trace data for an element on a new run and compare it to historical trace data we will delete any elements that we have not re-created... thus if we can't compare the serialized entities correctly we'll delete elements for no reason.

This is a problem because equals uses reference equality for reference types and this is a newly serialized object so it returns false. - specifically when searching for OrphanedTraceData from a callsite ` IEnumerable.Contains()` uses `Equals()` to check for equality.

equals now uses sequence equals to compare the string and int lists of the `MultiSerializableID`, and we use the hashcode of concatenated string as the hash.

This code should be merged to 2015 and 2017 as well.

### Declarations

- [x] merge to .91 2015
- [x] merge to .91 2016
- [x] merge to .91 2017

- [x] merge to master 2015
- [x] merge to master 2017

### Reviewers

@Benglin 

### FYIs
@riteshchandawar 
@mccrone


mjkkirschner:
@ikeough would also appreciate your thoughts if you have time to take a look.

riteshchandawar:
@Randy-Ma PTAL,.
@mjkkirschner we need Randy's comments as well on this. He is the one who fixed most of the stuff related to Adaptive Components and Element Binding. He has recently implemented Batch Processing of Adaptive Components as well.

Thanks,
Ritesh

src/Libraries/RevitServices/Persistence/ElementBinder.cs
```diff
@@ -47,6 +47,34 @@ public SerializableId(SerializationInfo info, StreamingContext context)
...
+            if (sID == null)
+            {
+                return false;
+            }
+
+            if (this.IntID.Equals(sID.IntID) && this.StringID.Equals(sID.StringID))
+            {
+                return true;
+            }
+            return false;
```
Benglin: Just thought this would be shorter:

```csharp
return (IntID.Equals(sID.IntID) && StringID.Equals(sID.StringID));
```

src/Libraries/RevitServices/Persistence/ElementBinder.cs
```diff
@@ -47,6 +47,34 @@ public SerializableId(SerializationInfo info, StreamingContext context)
...
+
+        public override int GetHashCode()
+        {
+            unchecked // Overflow is fine, just wrap
+            {
+                int hash = 17;
+
+                hash = hash * 23 + StringID.GetHashCode();
+                hash = hash * 23 + IntID.ToString().GetHashCode();
+                return hash;
```
Benglin: I'm not too experienced with hash code generation, so I'd typically do this (not saying yours is wrong, just wondering how that is derived):

```csharp
return (StringID.GetHashCode() ^ IntID.GetHashCode());
```

Benglin:
Thanks a ton for helping with this defect, @mjkkirschner! Appreciate it, from the way it looks, I don't have too much problem with it. Let's see what @ikeough says.

src/Libraries/RevitServices/Persistence/ElementBinder.cs
mjkkirschner: @Benglin I usually refer to this stack overflow answer that references a Java book, theres an explanation of the prime multiplication vs xoring.
http://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-an-overridden-system-object-gethashcode

src/Libraries/RevitServices/Persistence/ElementBinder.cs
mjkkirschner: thanks, will change it.

src/Libraries/RevitServices/Persistence/ElementBinder.cs
```diff
@@ -47,6 +47,34 @@ public SerializableId(SerializationInfo info, StreamingContext context)
...
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked // Overflow is fine, just wrap
+            {
+                int hash = 17;
+
+                hash = hash * 23 + StringID.GetHashCode();
+                hash = hash * 23 + IntID.ToString().GetHashCode();
```
Randy-Ma: Just a bit concerned this hash might not be unique.

For hash=(17*23+x)*23+y, given a hash value, (x, y) might not be unique so it might result in hash conflicts.

Randy-Ma: But I guess in this case mostly equals is the one that works.

Randy-Ma:
@mjkkirschner 
Thanks a ton for the fix! I have an impression element binding has been tested before. In case the same issue happens again, it will be very helpful if a test case can be added. 

@Benglin @riteshchandawar 
Except for few comments, I have no problem with the fix.

src/Libraries/RevitServices/Persistence/ElementBinder.cs
mjkkirschner: @Randy-Ma yes in this case equals is the method that is used, I've overridden hashcode as well though for completeness sake. If most are more familiar and comfortable with the stand xor hash function, of properties I will use that instead. I can't say one way or the other which will provide less collisions.

I will change the hash, and add a test.

mjkkirschner:
@Randy-Ma or @Benglin  please take another look, I've changed the hash function (by the way here is a very nice article on hash codes including a similar one to the one I used originally, I guess it was proposed by Donald Knuth and apparently has some issues http://papa.bretmulvey.com/post/124027987928/hash-functions ) 

I've also added 3 tests, two that check equals on `serializableIds` and `MultserializableIds` and one system test.

mjkkirschner:
added null checks to hashcode, will return 0 for properties that are null.

Benglin:
@ke-yu says he's interested in hash codes.

mjkkirschner:
merging all cherry picks and merging this in

riteshchandawar:
@mjkkirschner Can you merge this changes to RC 0.9.1 REvit 2016 banch as well?

mjkkirschner:
@riteshchandawar this PR (referenced above) should have merge the changes:
https://github.com/DynamoDS/DynamoRevit/pull/815

did it not?

riteshchandawar:
On latest build of RC 0.9.1 I am still getting the same behavior with Revit 2016.

mjkkirschner:
@riteshchandawar what behavior are you seeing? The adaptive components being deleted? Are you on an english version of Revit?

PR #819, opened 2016-02-01
MAGN 9341 followup 

mjkkirschner:
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9341

I found that when a .dyn file contained data to rebind to, but a Revit file did not contain the elements, these elements might be deleted immediately after they had been created(on first run) if the elements shared the same elementIds as those there were in trace, this is **not** uncommon, as elementIds are not unique. The result is that we delete the elements and the graph is in a null state and appears to have done nothing and will soon crash.

I have modified PostTraceReconciliation to use uniqueIds instead of elementsIDs, so that in the case that the elementsIds in trace are the same as the ones for new elements being created we do not delete the new elements.

### updates:

I have also made some changes to handle some cases that @riteshchandawar found where elements where still being deleted on rerun of the graph if the number of inputs to the `AdaptiveComponent.ByPoints` node changed.

I've modified `PostTraceReconciliation` for the cases where `MultiSerialzableID` is found as an orphan, but  either is totally contained by the elements which are about to be created on the latest graph run Or cases where the elements which are about to be created are contained in an orphaned Id, in either of these cases the orphaned Ids will be deleted and thus the newly created elements will be null...

Instead we remove these ids which intersect between current Ids and orphaned Ids.

I have done this in PostTraceReconciliation opposed to the `GetOrphanedSerializables()` method to keep the Revit implementation details out of core
https://github.com/DynamoDS/Dynamo/blob/master/src/Engine/ProtoCore/Lang/CallSite.cs#L621

last I have added null checks to `abstractFamily` class `toString` since the new AdaptiveComponent for batching constructor sometimes deletes elements from the Revit Document there are cases where a wrapper is left in the elementLifeCycleManager and calling toString on the empty wrapper causes the graph to spit out all nulls.  Imagine the case where an element collector gathers ACs from the document, but the AC constructor destroys them, but this is just a stop gap, we should look into that case more specifically in a future task as the behavior is not well defined.

### Declarations

### Reviewers

@Randy-Ma or @Benglin 

### FYIs

@riteshchandawar 
@kronz

### DIFF
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -233,16 +233,16 @@ namespace Dynamo.Applications.Models
 
         public override void PostTraceReconciliation(Dictionary<Guid, List<ISerializable>> orphanedSerializables)
         {
-            var orphanedIds = new List<int>();
+            var orphanedIds = new List<string>();
             var serializables =
                 orphanedSerializables
                 .SelectMany(kvp => kvp.Value)
                 .Where(x => x is ISerializable);
 
             orphanedIds.AddRange(serializables.Where(x => x is SerializableId).
-                Cast<SerializableId>().Select(sid => sid.IntID));
+                Cast<SerializableId>().Select(sid => sid.StringID));
             orphanedIds.AddRange(serializables.Where(x => x is MultipleSerializableId).
-                Cast<MultipleSerializableId>().SelectMany(sid => sid.IntIDs));
+                Cast<MultipleSerializableId>().SelectMany(sid => sid.StringIDs));
 
             if (!orphanedIds.Any())
                 return;
@@ -262,14 +262,14 @@ namespace Dynamo.Applications.Models
             }
         }
 
-        private static void DeleteOrphanedElements(IEnumerable<int> orphanedIds, ILogger logger)
+        private static void DeleteOrphanedElements(IEnumerable<string> orphanedIds, ILogger logger)
         {
             var toDelete = new List<ElementId>();
             foreach (var id in orphanedIds)
             {
                 // Check whether the element is valid before attempting to delete.
                 Element el;
-                if (DocumentManager.Instance.CurrentDBDocument.TryGetElement(new ElementId(id), out el))
+                if (DocumentManager.Instance.CurrentDBDocument.TryGetElement(id, out el))
                     toDelete.Add(el.Id);
             }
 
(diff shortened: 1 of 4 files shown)


mjkkirschner:
@Randy-Ma PTAL

src/Libraries/RevitServices/Persistence/ElementBinder.cs
```diff
@@ -152,16 +152,25 @@ public override int GetHashCode()
...
 
-         }
         }
 
+        /// <summary>
+        /// this method tests if this multiSerializableId is contained in another
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns></returns>
+        public virtual bool isSubset(MultipleSerializableId other)
```
Randy-Ma: It looks like the logic is checking whether there are intersections.

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -231,22 +233,79 @@ public new static RevitDynamoModel Start(IRevitStartConfiguration configuration)
...
+            //list of UUIDs we will remove from the list of orphans
+            var toRemove = new List<String>();
+
+            //foreach orphaned ID check 2 cases:
+            //1. an orphaned MultID is totally subset in one of the latest MultIDs
+            //2. the latest IDS are subset in a MultID in the orphan list
+            //in either of these cases we must remove the IDs from the orphan list so they are not deleted by accident
+            foreach (var orphan in orphanedSerializables)
+            {
+                // Selecting all nodes that are either a DSFunction,
```
Randy-Ma: we can continue here is the orphaned serializable is not an multiserializableID.

Randy-Ma:
@mjkkirschner 
Great! Thanks for the fix! I think I understand the cause now. After the slider is updated and the graph is run, a multiSerializableID with 9 elements is created and so the old one with 5 elements are all taken as orphaned and deleted. Other than few minor comments, LGTM

src/Libraries/RevitServices/Persistence/ElementBinder.cs
mjkkirschner: This method should check that the entire set of ids is a subset of the `other`, by checking if the set difference is empty this only returns true in the case that the entire set of Ids is a subset in the other, not only if there is an intersection.

mjkkirschner:
I am going to merge this to all revit branches including RC91 branches

PR #825, opened 2016-02-05
2015: UTC offset

aosyatnik:
### Purpose

Link to YouTrack:
[MAGN-8993](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8993) Correct UTC offset for Solar Radiation workflow from Revit.

If I understood the issue correct... This is an old bug, that was first mentioned here in 2014 year: http://dynamobim.com/forums/topic/sunsettings-current/

SunSettings.CurrentDateTime as well as SunSettings.StartDateTime and SunSettings.EndDateTime return wrong datetime value. This value has some strange offset.

image

For me this offset is 3 hours. But it depends on user's time zone. That's why MOHAMMAD RAHMANI (user from link above) had the same wrong offset, but for 5 hours.

This is bug of Revit API. It returns custom time - user time zone offset.
I found old API(2014) here: http://revitapisearch.com/html/e50880d2-cab4-3cb5-56c8-6a7faaa0725b.htm

It's said, that: "The date and time. The output value will be in Coordinated Universal Time (UTC), but input may be in local time as well." But I couldn't find any way to return local time.
So, I made some workaround:
I get users local timezone offset and add it to datetime, that provide Revit API.

Results:
image
image

NB! Sometimes Sunsettings are null or wrong... For now problem is not found. And I can't reproduce it consistently.
image

### Declarations

### Reviewers

Could you try this fix in your time zone?
@mjkkirschner 
@ramramps 

### FYIs

@ikeough , I believe Ian knows more about it, then anyone else.

### DIFF
--- a/src/Libraries/RevitNodes/Elements/SunSettings.cs
+++ b/src/Libraries/RevitNodes/Elements/SunSettings.cs
@@ -124,6 +124,11 @@ namespace Revit.Elements
             // Get user local hours offset.
             var offset = TimeZoneInfo.Local.GetUtcOffset(DateTime.UtcNow).Hours;
 
+            if (TimeZoneInfo.Local.IsDaylightSavingTime(DateTime.UtcNow))
+            {
+                offset--;
+            }
+
             // Remove user local offset. Just leave pure revit datetime.
             return utc.AddHours(offset);
         }
--- a/src/Libraries/RevitNodes/Elements/SunSettings.cs
+++ b/src/Libraries/RevitNodes/Elements/SunSettings.cs
@@ -89,7 +89,7 @@ namespace Revit.Elements
         /// </summary>
         public DateTime StartDateTime
         {
-            get { return InternalSunAndShadowSettings.StartDateAndTime; }
+            get { return TranslateTime(InternalSunAndShadowSettings.StartDateAndTime); }
         }
         
         /// <summary>
@@ -97,7 +97,7 @@ namespace Revit.Elements
         /// </summary>
         public DateTime EndDateTime
         {
-            get { return InternalSunAndShadowSettings.EndDateAndTime; }
+            get { return TranslateTime(InternalSunAndShadowSettings.EndDateAndTime); }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/Elements/SunSettings.cs
```diff
@@ -89,7 +89,7 @@ public double Azimuth
         /// </summary>
         public DateTime StartDateTime
```
mjkkirschner: can you modify the summaries to specify exactly what timezone the user can expect this time in, I.E -  "the start time of the solar study given in the local time of the solar study location" for all of these properties?

src/Libraries/RevitNodes/Elements/SunSettings.cs
```diff
@@ -105,8 +105,27 @@ public DateTime EndDateTime
...
+        /// But it returns customized datetime + user timezone offset, which is incorrect.
+        /// Althought in documentation it's said, 
+        /// that "The output value will be in Coordinated Universal Time (UTC), 
+        /// but input may be in local time as well."
+        /// There is no means to return local time.
+        /// </summary>
+        internal static DateTime TranslateTime(DateTime utc)
+        {
+            // Get user local hours offset.
+            var offset = TimeZoneInfo.Local.GetUtcOffset(DateTime.UtcNow).Hours;
```
mjkkirschner: this makes sense to me, what happens with daylight savings time? I don't think UTC has it, does the TimeZoneInfo take care of it?

mjkkirschner:
@aosyatnik a question, and one small change to the XML docs. Thanks for this fix :)

mjkkirschner:
also, I tested this on my machine on 2016 and it works for me.

aosyatnik:
Hi, @mjkkirschner , nice catch about day light saving...
I moved my clock to March 1. (no day light saving)

Then moved to March 27. (day light saving) - I had 1 additional hour.
So, I added check for Daylight Saving Time. Great, that `TimeZoneInfo` can check it with `IsDaylightSavingTime`!

Results:
image
image

pbidenko:
@aosyatnik, this looks pretty good.

PR #834, opened 2016-02-16
Update test cases because of the change of replication behavior

ke-yu:
### Purpose

This PR updates two test cases `RevitSystemTests.LevelTests.Level` and `RevitSystemTests.AdaptiveComponentTests.AdaptiveComponentByCurve` because of the change of replication behavior in https://github.com/DynamoDS/Dynamo/pull/6110.

In test case `RevitSystemTests.LevelTests.Level`, the second argument of `Level.ByElevationAndName` was a two-dimensional list while the first argument is one-dimensional list. As this node uses longest lacing, because of the change in https://github.com/DynamoDS/Dynamo/pull/6110, zip replication applies. This PR flattens the second argument.

In test case `RevitSystemTests.AdaptiveComponentTests.AdaptiveComponentByCurve`, the first argument of `AdaptiveComponent.ByParametersOnCurveReference` was a one-dimensional list. As the rank of first parameter is `double[][]` and the node uses longest lacing, zip replication applies so that the first argument won't be promoted to two-dimensional list before replication. This PR updates the first argument to two-dimensional list.

### FYIs
@riteshchandawar


mjkkirschner:
@ke-yu thanks!

PR #844, opened 2016-02-22
Fix MAGN-9229 Crash after running Core_AttractorPoint.dyn sample

Randy-Ma:
### Purpose

This fix is to address some crashes when the Revit background preview is turned on. In the case when the solids/surfaces are very small, in the process of building geoemtries from the facets, there will be exceptions thrown. For the case in MAGN-9229, the solids are really small and exceptions are thrown when geometries are attempted to be created out of the facets.

The fix here strengthens the tessellation process for geometries involved in Revit background preview by providing fallback mechanisms if exceptions are thrown in the process. 

The following logic is implemented for tessellating solids:
1). Try to build geometries out of solids;
2). If there are exceptions, then try to build geometries out of surface geometries of the solids;
3). If there are exceptions, then try to build geometries out of curve geometries of the solids' edges;

The followin logic is implemented for tessellating surfaces:
1). Try to build geometries out of surfaces;
2). If there are exceptions, then try to build geometries out of curve geometries of the surfaces' edges;

The followin logic is implemented for tessellating curves:
1). Try to build polylines/lines from tessellated points of curves;
2). If there are exceptions, try to create a line from the start point and end point;
3). If there are exceptions, try to create a point from the start point;
4). If there are exceptions, do not create any geometries;

As you can see, there are some loss of geometry information for tiny geometries and geometries in some edge cases(e.g. points with infinite X, Y or Z). But the stability will be improved a lot.

For geometries which are not drawn completely, a light red bounding box which is bigger than the original bounding box of the geometry is drawn around it to identify that something wrong has happened. (But this does not include point with infinite X, Y or Z value).

The following graph is trying to draw some cuboids and some tiny spheres:
image

The spheres can not be faceted correctly and are not drawn, instead a light red bounding box is drawn around it:
image

The fix will also fix the crashes in MAGN-6024 and MAGN-9486 which are related to Revit background preview.

### Declarations

### Reviewers

@sharadkjaiswal 

### FYIs

@kronz @mccrone @Racel @riteshchandawar @monikaprabhu

### DIFF
--- a/src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
+++ b/src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


sharadkjaiswal:
Thanks for working on it, it seems to me a bit of overengineering and we are fixing the symptoms rather than the real issue. Ideally, we should be able to control the tessellation to avoid any small edges. Can we put a debug assert for the scenario where geometry is big, but it fails to tessellate due to any small edge?

Randy-Ma:
Well, I have tried to control the tessellation by recursively decreasing the number of maximum divisions, for the case in MAGN-9229, even we set the maximum divisions is reduced to 10, there will still be exceptions. This is because the bodies involved are very small and geometries can not be built out of the facets.

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
```diff
@@ -192,40 +193,135 @@ private void RevitGeometryObjectFromGraphicItem(IGraphicItem item, ref List<Geom
...
             }
         }
 
+        /// <summary>
+        /// Tessellate the curve:
+        /// 1). If there are more than 2 points, create a polyline out of the points;
+        /// 2). If there are exactly 2 points, create a line;
+        /// 3). If there are exceptions thrown during the tessellation process, firstly a line will
+        /// will be attempted to be created from the start and end points. If this fails, a point will
+        /// be created instead.
```
Benglin:
```csharp
/// 3). If there's exception thrown during the tessellation process, attempt to create 
/// a line from start and end points. If that fails, a point will be created instead.
```

Randy-Ma:
@kronz @Racel 
For geometries that can not be previewed completely in Revit background preview, I am using a fallback mechanism in this pull request and also for these geometries, a light red bounding box which is slightly bigger than the geometry's bounding box will be drawn to identify that these geometries are not completely drawn.

Randy-Ma:
@sharadkjaiswal 
As discussed offline, I am merging this now.

PR #845, opened 2016-02-22
Revit UTC

aosyatnik:
### Purpose

Link to YouTrack:
[MAGN-8993](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8993) Correct UTC offset for Solar Radiation workflow from Revit.

After some investigation I came up to the solution, that the problem is that it's used daylight saving time.
Look at those images:

image

image

But then I faced with the problem. Revit API returns just datetime. But there is no means to detect if this is the summer time. Moreover, you event don't know time zone!
Revit API returns next things: datetime, latitude, longitude and useless name of the city/region, like "Boston, MA" (have no idea, how we can use it... especially if it is not famous place).

I found one project, that can translate  latitude and longitude to timezone. 
Here is link: https://github.com/mj1856/GeoTimeZone
But it translates coordinates into Olson time zone. So I had to create one more additional method `OlsonTimeZoneToTimeZoneInfo`, where I convert Olson time zone into TimeZoneInfo.

For now I won't create 2015, 2017 branches. Maybe, it's wrong direction. And I chose bad solution.

## Results:

Now they are the same.

image

image

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@jnealb


QilongTang:
Happen to stop by here in this PR. After reading http://web.cs.ucla.edu/~eggert/tz/tz-link.htm. Looks like if we want to do it locally, the absolute right way is to integrate the timezone database (http://www.iana.org/time-zones). Not sure how much effort will it be though. There are tons of webAPI for converting (e.g. Google), but I guess they are off the table. You can find lots of info in this article.

src/Libraries/RevitNodes/RevitNodes.csproj
```diff
@@ -45,6 +45,9 @@
       <HintPath>$(DYNAMOAPI)\DynamoServices.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="GeoTimeZone">
+      <HintPath>..\..\packages\GeoTimeZone.1.1.0\lib\portable-net40+sl50+win+wp80\GeoTimeZone.dll</HintPath>
```
mjkkirschner: if we are using this package, we should add license information to the repo:
https://raw.githubusercontent.com/mj1856/GeoTimeZone/master/LICENSE

mjkkirschner:
another potential library to investigate: https://github.com/nodatime/nodatime

src/Libraries/RevitNodes/RevitNodes.csproj
mjkkirschner: is this all we need to do to get nuget to download the package?

src/Libraries/RevitNodes/Elements/SunSettings.cs
```diff
@@ -143,8 +170,165 @@ internal static DateTime TranslateTime(DateTime utc)
...
+        { "Indian/Mauritius", "Mauritius Standard Time" },
+        { "Pacific/Apia", "Samoa Standard Time" },
+        { "Pacific/Auckland", "New Zealand Standard Time" },
+        { "Pacific/Fiji", "Fiji Standard Time" },
+        { "Pacific/Guadalcanal", "Central Pacific Standard Time" },
+        { "Pacific/Guam", "West Pacific Standard Time" },
+        { "Pacific/Honolulu", "Hawaiian Standard Time" },
+        { "Pacific/Pago_Pago", "UTC-11" },
+        { "Pacific/Port_Moresby", "West Pacific Standard Time" },
+        { "Pacific/Tongatapu", "Tonga Standard Time" }
```
mjkkirschner: seems like maybe we are missing some mappings...http://unicode.org/repos/cldr-tmp/trunk/diff/supplemental/zone_tzid.html

src/Libraries/RevitNodes/Elements/SunSettings.cs
```diff
@@ -143,8 +170,165 @@ internal static DateTime TranslateTime(DateTime utc)
...
+        {
+            var windowsTimeZoneId = default(string);
+            var windowsTimeZone = default(TimeZoneInfo);
+            if (olsonWindowsTimes.TryGetValue(olsonTimeZoneId, out windowsTimeZoneId))
+            {
+                try { windowsTimeZone = TimeZoneInfo.FindSystemTimeZoneById(windowsTimeZoneId); }
+                catch (TimeZoneNotFoundException) { }
+                catch (InvalidTimeZoneException) { }
+            }
+            return windowsTimeZone;
```
mjkkirschner: if we can't find the id what is the result of returning null?

mjkkirschner:
@aosyatnik I have a few questions on this solution - thanks for investigating these issues and coming up with this solution, like @QilongTang says, I think you are on the right track - 

in addition to this work, can we add a few tests?

src/Libraries/RevitNodes/RevitNodes.csproj
aosyatnik: I guess we need `packages.config` file. It looks like it's all, that we need...

PR #848, opened 2016-02-24
Creating a separate MSI for DynamoRevit

sh4nnongoh:
### Purpose

To create a DynamoRevit MSI that is independent on DynamoCore files.

* Added a new configuration (Install) for DynamoRevit2013.sln and its projects.
With this Install configuration, the bin directory will be generated in the local DynamoRevit directory instead of the Dynamo/bin directory.

* With the new configuration, the configuration files in src/Config/* was edited accordingly.
(Specifically removing the dependency on "user_local.prop".) 

* A new solution called DynamoAddInGenerator was created in the tools directory.
This project was referenced from a similar project in Dynamo, but has been edited heavily to suit the new requirements.
This project uses DynamoInstallDetective.dll & RevitAddIn.dll to generate an executable that is used to generate/delete Dynamo Revit Addins.
The referenced dlls are stored in the tools/install/Extra directory.

* The DynamoRevitInstall project has been edited to suit the new purpose.
The new MSI will install Dynamo Revit in a separate folder from Dynamo Core.
It will also update the registry of the new installed location.
The MSI will also use the previously generated DynamoAddInGenerator.exe to create/delete the addins.

### Reviewers

@sharadkjaiswal 

### FYIs

@Benglin


src/DynamoRevitInstall/DynamoRevitInstall.wixproj
```diff
@@ -30,24 +30,48 @@
...
+      <Name>WixUtilExtension</Name>
+    </WixExtension>
     <WixExtension Include="WixUIExtension">
       <HintPath>$(WixExtDir)\WixUIExtension.dll</HintPath>
       <Name>WixUIExtension</Name>
     </WixExtension>
   </ItemGroup>
   <Import Project="$(WixTargetsPath)" />
+  <Target Name="BeforeBuild"> 
+    <GetAssemblyIdentity AssemblyFiles="$(DYNAMOAPI)\DynamoCore.dll">
```
sharadkjaiswal: The version for MSI should be based on DynamoRevitDS.dll

PR #879, opened 2016-02-27
MAGN-8993 Correct UTC offset for Solar Radiation workflow from Revit (for 2015)

sm6srw:
### Purpose

Bug fix: Get the local offset for the given time, not the current time

### Declarations

### Reviewers

@aosyatnik 

### FYIs

@jnealb


aosyatnik:
Hi, @sm6srw , I see, so we need to apply offset not from user local time, but from set time. Interesting why they did this... Well, in any way, thank you. Could you, please, delete these descriptions:
`///     Some regions may have a Daylight Savings (-1 hour) Time inaccuracy.`

aosyatnik:
After removing comments looks good to me.

sm6srw:
Thanks, comments removed.

PR #903, opened 2016-03-08
MAGN-9371 Refactor and pull BREP spike code into Revit 2017 branch to turn on Brep Builder API for Surfaces and Solids

sm6srw:
### Purpose

Use BREP builder for direct shape whenever possible. This uses recently exposed LibG methods for loops/co-edges on Faces. All surfaces are converted to nurbs surfaces before being passed on to the BREP. That works for most simple cases. The next step would be to add support for more Revit surface types and only use nurbs surfaces as the last resort.

Tessellated direct shape are still used if the BREP builder for some reason fails.

This is only for Revit 2017.

### Declarations

### Reviewers

@mjkkirschner 
@aparajit-pratap


mjkkirschner:
@sm6srw I know that the state of RTF and 2017 is not great, but I think some unit tests should be stubbed out for simple geometric cases - IE create a cube and create a Brep from it, check the resulting geometry etc. It will be good to make sure future changes don't regress the cases that you know are working currently and help inform us of what currently works and what doesn't.

src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -220,17 +220,27 @@ private static IList<GeometryObject> GenerateTessellatedGeo(DesignScriptEntity s
...
-                    TessellatedShapeBuilderFallback.Mesh,
-                    MaterialId: materialId);
+                var geometry = ProtoToRevitSolid.ToRevitType(shapeReference as Autodesk.DesignScript.Geometry.Surface);
+                if(geometry != null)
+                {
+                    tessellatedShape = new List<GeometryObject>() { geometry };
+                }
+                else
+                {
+                    tessellatedShape = (shapeReference as Autodesk.DesignScript.Geometry.Surface).ToRevitType(TessellatedShapeBuilderTarget.AnyGeometry, TessellatedShapeBuilderFallback.Mesh, MaterialId: materialId);
```
mjkkirschner: can you add a comment to these cases stating that if the BRep is null, we fallback to the tessellated shape builder?

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+using Edge = Autodesk.DesignScript.Geometry.Edge;
+using Face = Autodesk.DesignScript.Geometry.Face;
+using ProtoPt = Autodesk.DesignScript.Geometry.Point;
+using Solid = Autodesk.DesignScript.Geometry.Solid;
+using Surface = Autodesk.DesignScript.Geometry.Surface;
+
+namespace Revit.GeometryConversion
+{
+   [SupressImportIntoVM]
+   public static class ProtoToRevitSolid
```
mjkkirschner: I don't have a better suggestion - but is this the right name for this class?

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+
+               // Check if the nurbs surface has flipped compared to the original surface
+               if (geom.NormalAtParameter(.5, .5).Dot(ngeom.NormalAtParameter(.5, .5)) < 0)
+               {
+                   flipped = true;
+               }
+
+               // Create Revit nurbs surface
+               var bbface = BRepBuilderSurfaceGeometry.CreateNURBSSurface(ngeom.DegreeU, ngeom.DegreeV,
+                   ngeom.UKnots(), ngeom.VKnots(), ngeom.ControlPoints().SelectMany(x => x.Select(y => y.ToXyz())).ToList(),
```
mjkkirschner: hmm, we may need to use InHostUnits() on these xyzs we're building here from the control points depending on the performHostUnitConversion bool.... this would be a great place for a test :)

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+                   ngeom.Weights().SelectMany(x => x).ToList(),
+                   false,
+                   new BoundingBoxUV());
```
mjkkirschner: I remember not being sure about using a default boundingBox here, do we have any more information on what this should be?

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+               // add loops and connected edges
+               foreach (var loop in protoFace.Loops)
+               {
+                   var loopId = brb.AddLoop(faceId);
+
+                   foreach (var coedge in loop.CoEdges)
+                   {
+                       var edge = coedge.Edge;
+                       BRepBuilderGeometryId edgeId;
+                       if (edgeDict.ContainsKey(edge))
```
mjkkirschner: I'm a little concerned here, if it this dictionary lookup seems to be working thats great! But how is hashcode implemented for edges? I'm more curious than anything because I know this didn't used to work :smile:

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+           edgeDict.ToList().ForEach(x => x.Key.Dispose());
+           faces.ForEach(x => x.Dispose());
+
+           // Get result
+           var outcome = brb.Finish();
+           var converted = brb.GetResult();
+
+           if (converted == null)
+           {
+               throw new Exception("An unexpected failure occurred when attempting to convert the solid");
```
mjkkirschner: I think if this is going to be bubbled up the user it needs to be localized - we should add it to the resources dictionary for this project so the localization team can localize them.

https://github.com/DynamoDS/DynamoRevit/blob/Revit2015/src/Libraries/RevitNodes/Properties/Resources.resx

it needs to be added to both en-us.resx and .resx

mjkkirschner:
@sm6srw thanks for this work! looks really good to me, a few questions, and requests.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
sm6srw: ProtoToRevitBRep?

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
sm6srw: It has worked very well so far. Revit would have complained otherwise. Is there a safer way?

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
sm6srw: I guess it could for a node developer as it is right now. I will add resources for it.

src/Libraries/RevitNodes/Elements/DirectShape.cs
sm6srw: Will do.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
mjkkirschner: I like it, since this sometimes produces a surface or shell

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
mjkkirschner: you could implement your own equalityComparator for dictionary and use that so it's explicit.... but lets not do that - if it's working lets just have a test make sure it does not regress when the LibG version is changed.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
aparajit-pratap: Can we not name this `DynamoToRevitBRep` or something on those lines since we are essentially converting Dynamo geometry into Revit. I think the term `Proto` is borrowed from `ProtoGeometry` and is coming from the DesignScript days but doesn't seem to make much sense anymore especially to new developers.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+       public static GeometryObject ToRevitType(Solid sol,
+          bool performHostUnitConversion = true)
+       {
+           try
+           {
+               return ToRevitType(sol, performHostUnitConversion, BRepType.Solid);
+           }
+           catch
+           {
+               return null;
```
aparajit-pratap: Wondering if it would be nice to log this as a failure  in the Dynamo console with a message like "failed to convert geometry into Revit BRep" or something on those lines.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+           //foreach face in solid
+           foreach (Face protoFace in faces)
+           {
+               var geom = protoFace.SurfaceGeometry();
+
+               var ngeom = geom.ToNurbsSurface();
+               bool flipped = false;
+
+               // Check if the nurbs surface has flipped compared to the original surface
+               if (geom.NormalAtParameter(.5, .5).Dot(ngeom.NormalAtParameter(.5, .5)) < 0)
```
aparajit-pratap: I think we should fix this in LibG so that the NURBS surface does not flip ever. It just means that it's a bug in `ToNurbsSurface` when it happens.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+               {
+                   flipped = true;
+               }
+
+               // Create Revit nurbs surface
+               var bbface = BRepBuilderSurfaceGeometry.CreateNURBSSurface(ngeom.DegreeU, ngeom.DegreeV,
+                   ngeom.UKnots(), ngeom.VKnots(), ngeom.ControlPoints().SelectMany(x => x.Select(y => y.ToXyz())).ToList(),
+                   ngeom.Weights().SelectMany(x => x).ToList(),
```
aparajit-pratap: I think there are some issues with the `Weights` property in `NurbsSurface` in LibG. It does not always give reliable results. For example, try converting a circle to a `NurbsCurve` in LibG. You'll see incorrect weight values. If so this property needs to be fixed as well. I think there are bugs tracked against these.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
```diff
@@ -0,0 +1,154 @@
...
+                   false,
+                   new BoundingBoxUV());
+
+               // Dispose geometries
+               geom.Dispose();
```
aparajit-pratap: It would be cleaner if we used `using` statements to dispose these intermediate variables.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
aparajit-pratap: `GetHashCode` has been implemented for `Edge`s in the context of topology comparisons and not using the default reference equality. For example, you can compare an edge of a body with the same topological edge from a face of that same body even though they might be 2 different objects.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
sm6srw: Yes, I agree that `DynamoToRevitBRep` is better. We already have a couple of classes/files named `ProtoToRevit*`. Is it OK to rename those too while we at it?

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
sm6srw: Will do...

src/Libraries/RevitNodes/Elements/DirectShape.cs
sm6srw: I will move try/catch over here with proper comments.

src/Libraries/RevitNodes/GeometryConversion/ProtoToRevitFace.cs
sm6srw: I will move try/catch over to DirectShape.cs and then make sure we have proper throws here for node developers that may use these directly.

src/Libraries/RevitNodes/GeometryConversion/DynamoToRevitBRep.cs
```diff
@@ -0,0 +1,150 @@
...
+                       {
+                           flipped = true;
+                       }
+
+                       // Create Revit nurbs surface
+                       var bbface = BRepBuilderSurfaceGeometry.CreateNURBSSurface(ngeom.DegreeU, ngeom.DegreeV,
+                           ngeom.UKnots(), ngeom.VKnots(), ngeom.ControlPoints().SelectMany(x => x.Select(y => y.ToXyz())).ToList(),
+                           ngeom.Weights().SelectMany(x => x).ToList(),
+                           false,
+                           null);
```
sm6srw: Let Revit figure out the BB.

src/Libraries/RevitNodes/GeometryConversion/DynamoToRevitBRep.cs
```diff
@@ -0,0 +1,150 @@
...
+                       if (geom.NormalAtParameter(.5, .5).Dot(ngeom.NormalAtParameter(.5, .5)) < 0)
+                       {
+                           flipped = true;
+                       }
+
+                       // Create Revit nurbs surface
+                       var bbface = BRepBuilderSurfaceGeometry.CreateNURBSSurface(ngeom.DegreeU, ngeom.DegreeV,
+                           ngeom.UKnots(), ngeom.VKnots(), ngeom.ControlPoints().SelectMany(x => x.Select(y => y.ToXyz())).ToList(),
+                           ngeom.Weights().SelectMany(x => x).ToList(),
+                           false,
```
sm6srw: Let Revit figure out the BB.

src/Libraries/RevitNodes/GeometryConversion/DynamoToRevitBRep.cs
```diff
@@ -0,0 +1,150 @@
...
+                       bool flipped = false;
+                       // Check if the nurbs surface has flipped compared to the original surface
+                       if (geom.NormalAtParameter(.5, .5).Dot(ngeom.NormalAtParameter(.5, .5)) < 0)
+                       {
+                           flipped = true;
+                       }
+
+                       // Create Revit nurbs surface
+                       var bbface = BRepBuilderSurfaceGeometry.CreateNURBSSurface(ngeom.DegreeU, ngeom.DegreeV,
+                           ngeom.UKnots(), ngeom.VKnots(), ngeom.ControlPoints().SelectMany(x => x.Select(y => y.ToXyz())).ToList(),
```
mjkkirschner: @sm6srw I'm still a bit worried about the XYZs heres, have you thought about if we need to convert them to the host units?

mjkkirschner:
@sm6srw this looks good to me :) - I have one remaining question though.

src/Libraries/RevitNodes/GeometryConversion/DynamoToRevitBRep.cs
sm6srw: @mjkkirschner You where right to be worried. The whole thing falls apart when using metric projects. I need to to do some more testing.

src/Libraries/RevitNodes/GeometryConversion/DynamoToRevitBRep.cs
```diff
@@ -91,7 +78,7 @@ public static class DynamoToRevitBRep
 
                        // Create Revit nurbs surface
                        var bbface = BRepBuilderSurfaceGeometry.CreateNURBSSurface(ngeom.DegreeU, ngeom.DegreeV,
-                           ngeom.UKnots(), ngeom.VKnots(), ngeom.ControlPoints().SelectMany(x => x.Select(y => y.ToXyz())).ToList(),
+                           ngeom.UKnots(), ngeom.VKnots(), ngeom.ControlPoints().SelectMany(x => x.Select(y => y.ToXyz(performHostUnitConversion))).ToList(),
```
sm6srw: @mjkkirschner Only convert here when asked for.

src/Libraries/RevitNodes/GeometryConversion/DynamoToRevitBRep.cs
```diff
@@ -122,7 +109,7 @@ public static class DynamoToRevitBRep
                                    // But there are cases when edges ends up slightly outside one of the surfaces and Revit fails.
                                    //
                                    // This is something that we can be improve going forward.
-                                   edgeId = brb.AddEdge(BRepBuilderEdgeGeometry.Create(curve.ToRevitType()));
+                                   edgeId = brb.AddEdge(BRepBuilderEdgeGeometry.Create(curve.ToRevitType(performHostUnitConversion)));
```
sm6srw: @mjkkirschner And here...

mjkkirschner:
@sm6srw I think this looks really good, I think you can merge this, but if you have time, it might make sense to add a metric unit test since we missed that initially.

sm6srw:
@mjkkirschner Yep, that is a great idea. I will make 2 metric versions of the ones I added.

sm6srw:
@mjkkirschner Added some metric tests. I will run some tests and then merge if you don't have anymore objections.

PR #910, opened 2016-03-11
MAGN-9543 revit model updater throws exception when modifying sun settings (for Revit 2017 0.9.2) 

sm6srw:
### Purpose

Dynamo is watching for changes in the Revit model using an updater. The updater is activated when Dynamo loads but the relationship with the Revit model is established when Dynamo loads a document.

This PR is simply adding a null document check to the updater preventing the updater to do anything bad before Dynamo is ready.

This is for Revit 2017, 0.9.2.

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@jnealb

### DIFF
--- a/src/Libraries/RevitServices/Elements/ModelUpdater.cs
+++ b/src/Libraries/RevitServices/Elements/ModelUpdater.cs
@@ -167,6 +167,12 @@ namespace RevitServices.Elements
         void RevitServicesUpdater_Updated(object sender, UpdaterArgs args)
         {
             var doc = DocumentManager.Instance.CurrentDBDocument;
+
+            // Are we loaded yet?
+            if (doc == null)
+                // No
+                return;
+
             var added = args.Added.Select(x => doc.GetElement(x).UniqueId);
             var addedIds = args.Added;
             var modified = args.Modified.Select(x => doc.GetElement(x).UniqueId).ToList();


mjkkirschner:
@sm6srw all branches look okay to merge for me. I would normally say this would be good to test with a system test to make sure it does not regress, but in the case of 2017 we can't yet and since that is the only place we're hitting this we need to wait until we deal with RTF.

Can you instead add a task to add a test here in the future.

LGTM on all branches.

sm6srw:
@mjkkirschner Thanks! One question: Is it even possible to create a test for this? This will never happen as soon as Dynamo is initialized and it looks like Dynamo is initialized as soon as the test framework is up and running.  I will do a quick check.

I was thinking about adding a test on the Revit side until I realized that Dynamo isn't installed on internal builds. This is probably something we would like to change in the future.

mjkkirschner:
@sm6srw maybe I'm misunderstanding the fix here, I thought that the null check was needed because the update is getting fired when there is no document, but Dynamo is loaded, just not attached to a current revit doc.

If thats the case, there *might* be a way to get into that state in RTF, it might not be worth modifying RTF to enable that behavior, instead maybe something simple like a unit test that returns correctly if this method is raised with a null document.

sm6srw:
@mjkkirschner I had no success creating a test for this condition. I will create a task.

mjkkirschner:
@sm6srw it may not be worth it - lets create the task and reevaluate it later.

PR #916, opened 2016-03-15
ModelCurve.ByCurve fix

aosyatnik:
### Purpose

Link to YouTrack:
[MAGN-7162](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7162) if grid is an input to ModelCurve.ByCurve hangs/Crash dynamo in Run automatic mode

# Reason of infinite loop:
I created next graph with several revit elements created from grid curve: Beam, Wall and Model Curve.
image

I want to pay your attention, that every time, when I click  "Run" button Wall and Beam update old Revit element ( ID is the same). But ModelCurve creates new element (ID every time is different):
imageimage

That means Dynamo instead of updating existing item creates new one.

I hardly understand how ModelCurve should be created and updated, but I took a look at different places as walls, and beams... And tried to create something similar, fix infinite loop and do not break tests :)

Well, I hope, that I did it in the right way. Tests passed at least, so I hope, that I didn't break functionality further.
image

Unfortunately, I couldn't try it in Revit 2015, since I have next exception and couldn't load Dynamo at all: 
image

But in Revit 2016 fix works fine.

### Declarations

- [x] All tests pass using the self-service CI.

### Reviewers

I ask Revit-guru for review :)
@mjkkirschner 
@sm6srw 

### FYIs

@jnealb

### DIFF
--- a/src/Libraries/RevitNodes/Elements/ModelCurve.cs
+++ b/src/Libraries/RevitNodes/Elements/ModelCurve.cs
@@ -79,11 +79,8 @@ namespace Revit.Elements
             if (mc != null)
             {
                 InternalSetCurveElement(mc);
-                if (!InternalSetSketchPlaneFromCurve(crv))
-                {
-                    InternalSetCurve(crv);
-                    return;
-                }
+                InternalSetSketchPlaneFromCurve(crv);
+                return;
             }
 
             ElementId oldId = (mc != null) ? mc.Id : ElementId.InvalidElementId;
@@ -134,20 +131,35 @@ namespace Revit.Elements
         #region Private mutators
 
         /// <summary>
-        /// Set the curve internally.  Returns false if this method failed to set the curve
+        /// Set the plane and the curve internally.
         /// </summary>
-        /// <param name="c"></param>
-        /// <returns></returns>
-        private bool InternalSetSketchPlaneFromCurve(Autodesk.Revit.DB.Curve c)
+        private void InternalSetSketchPlaneFromCurve(Curve newCurve)
         {
             TransactionManager.Instance.EnsureInTransaction(Document);
 
-            // Infer the sketch plane
-            Autodesk.Revit.DB.Plane plane = CurveUtils.GetPlaneFromCurve(c, false);
+            Plane newPlane = CurveUtils.GetPlaneFromCurve(newCurve, false);
+            Plane oldPlane = InternalCurveElement.SketchPlane.GetPlane();
 
-            // attempt to change the sketch plane
-            bool needsRemake = false;
-            string idSpUnused = resetSketchPlaneMethod(this.InternalCurveElement, c, plane, out needsRemake);
+            var angleBetweenPlanes = newPlane.Normal.AngleTo(oldPlane.Normal);
+            var distanceBetweenOrigins = newPlane.Origin.DistanceTo(oldPlane.Origin);
+
+            Autodesk.Revit.DB.SketchPlane sp = null;
+
+            // Planes are different.
+            if (Math.Abs(angleBetweenPlanes) > CurveUtils.Tolerance || distanceBetweenOrigins > CurveUtils.Tolerance)
+            {
+                sp = GetSketchPlaneFromCurve(newCurve);
+                InternalCurveElement.SetSketchPlaneAndCurve(sp, newCurve);
+            }
+            // Planes are the same.
+            else
+            {
+                InternalSetCurve(newCurve);
+            }
+
+            string idSpUnused = String.Empty;
+            if (sp != null && InternalCurveElement.SketchPlane.Id != sp.Id)
+                idSpUnused = sp.UniqueId;
 
             // if we got a valid id, delete the old sketch plane
             if (idSpUnused != String.Empty)
@@ -156,8 +168,6 @@ namespace Revit.Elements
             }
 
             TransactionManager.Instance.TransactionTaskDone();
-
-            return !needsRemake;
         }
 
         #endregion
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/Elements/ModelCurve.cs
```diff
@@ -269,41 +262,8 @@ private static string resetSketchPlaneMethod(Autodesk.Revit.DB.CurveElement mc,
...
-                System.String nameOfMethodSetCurveAlt = "SetSketchPlaneAndCurve";
-
-                foreach (MethodInfo m in curveElementMethods)
-                {
-                    if (m.Name == nameOfMethodSetCurve || m.Name == nameOfMethodSetCurveAlt)
-                    {
-                        object[] argsM = new object[2];
-                        sp = GetSketchPlaneFromCurve(c);
-                        argsM[0] = sp;
-                        argsM[1] = null;
```
aosyatnik: Here is the most strange place. It always fails with exception curve can't be null. I believe here should be not null, but something meaningful.

mjkkirschner:
@aosyatnik thank you for your analysis and hard work here. I think we can take a simpler approach to fixing this bug.

fundamentally this issue is because of the `All Elements of Category` Node, and the `ModelCurve.ByCurve`node, and I think some bugs in this code :).

so! When a model curve is created it triggers a document modification event - I believe this triggers the `All Elements of Category` Node to fire again as well, as it needs to inspect the entire Revit Model for new items.

This in turn fires the `ModelCurve.ByCurve` node again.  We should stop the infinite loop here. But this node is not working right.

in the very start of the static constructor for this node we do a few things:
1. detect that there is old element stored in trace
2. check that we can set the plane and that the plane of the new curve is similar to the old one.
3. I think at this point the logic is incorrect, if the plane does not need to be rebuilt, then we should continue(so, I believe this bool is flipped!!)
4. If we made it this far we need to check if the curve we're about to set is similar to the curve that is stored on the old element. IF they are similar we should bail out of here because if we update the curve again we'll cause a new document modification and the whole graph will run again.

I think the changes can be this simple:

screen shot 2016-03-15 at 9 13 13 pm

can you give me your thoughts, and try this out?

mjkkirschner:
I think we also need to modify the `InternalSetCurve` method - it seems to do the exact opposite of what we want, which is not set the curve if it is different. I think the bool is again just flipped.

screen shot 2016-03-15 at 9 27 06 pm

and we could remove the outer check in the model curve constructor since it uses this `internalSetCurve` method . You can see very similar logic to this on Structural Framing, and Wall constructors.

https://github.com/DynamoDS/DynamoRevit/blob/2758fc2f42fbb0ad43ea3cad461c3a5e334c2f69/src/Libraries/RevitNodes/Elements/StucturalFraming.cs#L236

tested this fix (not extensively) on my machine and seems to work nicely

aosyatnik:
@mjkkirschner , I agree. My first thought was: that bool result is flipped. 
I tried and it works fine. One thing, that you didn't see: 
image

This code calls `ResetSketchPlaneAndCurve` with arguments `sp, null` and always fails with exception, that curve can't be null. I changed `null` into `c`, now it works.

src/Libraries/RevitNodes/Elements/ModelCurve.cs
```diff
@@ -79,11 +79,13 @@ private void InitModelCurve(Autodesk.Revit.DB.Curve crv, bool makeReferenceCurve
             if (mc != null)
             {
                 InternalSetCurveElement(mc);
-                if (!InternalSetSketchPlaneFromCurve(crv))
```
mjkkirschner: I think you need to only flip this in one place. here or on line 162

src/Libraries/RevitNodes/Elements/ModelCurve.cs
```diff
@@ -79,11 +79,13 @@ private void InitModelCurve(Autodesk.Revit.DB.Curve crv, bool makeReferenceCurve
...
+                if (InternalSetSketchPlaneFromCurve(crv))
                 {
                     InternalSetCurve(crv);
-                    return;
```
mjkkirschner: I believe we need the return statement here - we only return if can successfully set the plane and potentially the curve.

src/Libraries/RevitNodes/Elements/ModelCurve.cs
```diff
@@ -157,9 +159,10 @@ private bool InternalSetSketchPlaneFromCurve(Autodesk.Revit.DB.Curve c)
 
             TransactionManager.Instance.TransactionTaskDone();
 
-            return !needsRemake;
```
mjkkirschner: I think we should not invert this bool, since you already inverted the bool on line 82

mjkkirschner:
@aosyatnik a few changes needed I think - but besides that the others changes you made look good :)

aosyatnik:
Hi, @mjkkirschner , I believe I did right. I found Revit API:
image

As it's said `SetSketchPlaneAndCurve` sets plane and curve. So, `InternalCurveElement.GeometryCurve` will be set in `SetSketchPlaneAndCurve`. We need to call `InternalSetCurve` just in case if `SetSketchPlaneAndCurve` failed, i.e. when `needsRemake` is `false`.

src/Libraries/RevitNodes/Elements/ModelCurve.cs
mjkkirschner: hmmm, I think I see what you are saying here - atleast we should change the comment on this `InternalSetSketchPlaneFromCurve` method.

src/Libraries/RevitNodes/Elements/ModelCurve.cs
mjkkirschner: I still don't understand how the logic for case 2 here works, can you explain it to me?

the cases I see:

1. we identify that an old element existed, and we attempt to rebind, if we successfully set the curve and the plane, we do this and we return - I believe this will trigger a document modification event, and call the constructor again, we'll go into this method again, and determine we don't need to set the plane, - so we shouldn't modify the curve, and I think the loop will end here. Seems okay.

2. we identify that an old element existed, and attempt to rebind, we don't need to reset the plane - so we return false, and now... we never set the curve.... now we just return from the constructor and nothing has changed. Can you verify that this isn't an issue. I am thinking about the return statement on line 272, where `needsSketchPlaneReset` will still be false, and we won't have reset the plane, we won't set the curve in any case that I can see.

3. We identify an old element existed, and attempt to rebind, we can't find a reset method and need to reset the plane, so we return true set the curve on the model curve... will this work or throw an exception?

I guess I am just concerned because it doesn't seem there is a way for this code to hit the initial branch that creates a new model curve in the case we can't rebind. Can you please test this code out on some different curves in Dynamo, and move adjust the inputs to force rebinding if you have not already. (not just lines from grids)

PR #922, opened 2016-03-18
ModelCurve fix.

aosyatnik:
### Purpose

Link to YouTrack:
[MAGN-7162](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-7162) if grid is an input to ModelCurve.ByCurve hangs/Crash dynamo in Run automatic mode

Fix:
Rewrote `InternalSetSketchPlaneFromCurve`. Now, if planes are different old plane is updated with new one. If planes are similar, curve is updated with the new one.

Dynamo doesn't create new Revit element every time, when Run button is hit.
Dynamo updates old revit element, if it is presented.

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodes/Elements/ModelCurve.cs
+++ b/src/Libraries/RevitNodes/Elements/ModelCurve.cs
@@ -128,7 +128,7 @@ namespace Revit.Elements
 
         #endregion
 
-        #region Private mutators
+        private readonly double tolerance = 0.01;
 
         /// <summary>
         /// Set the plane and the curve internally.
@@ -146,7 +146,7 @@ namespace Revit.Elements
             Autodesk.Revit.DB.SketchPlane sp = null;
 
             // Planes are different.
-            if (Math.Abs(angleBetweenPlanes) > CurveUtils.Tolerance || distanceBetweenOrigins > CurveUtils.Tolerance)
+            if (angleBetweenPlanes > tolerance || distanceBetweenOrigins > tolerance)
             {
                 sp = GetSketchPlaneFromCurve(newCurve);
                 InternalCurveElement.SetSketchPlaneAndCurve(sp, newCurve);
@@ -169,8 +169,6 @@ namespace Revit.Elements
 
             TransactionManager.Instance.TransactionTaskDone();
         }
-
-        #endregion
 
         #region Public constructor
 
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/Elements/ModelCurve.cs
```diff
@@ -134,20 +131,35 @@ private void InitModelCurve(Autodesk.Revit.DB.Curve crv, bool makeReferenceCurve
...
+        private void InternalSetSketchPlaneFromCurve(Curve newCurve)
         {
             TransactionManager.Instance.EnsureInTransaction(Document);
 
-            // Infer the sketch plane
-            Autodesk.Revit.DB.Plane plane = CurveUtils.GetPlaneFromCurve(c, false);
+            Plane newPlane = CurveUtils.GetPlaneFromCurve(newCurve, false);
+            Plane oldPlane = InternalCurveElement.SketchPlane.GetPlane();
+
+            var angleBetweenPlanes = newPlane.Normal.AngleTo(oldPlane.Normal);
```
mjkkirschner: it is worth checking if the value returned from this AngleTo function is in degrees or radians, and then if the tolerance you use later makes sense for that return value.

aosyatnik: API says it's radians.
How can we check it?
I mean... Should I check whether it's radians or not?
image

mjkkirschner: @aosyatnik I believe it's radians if the docs say so :) - my question is does your check against `CurveUtils.Tolerance` make sense is that context? or should we use a different tolerance.

aosyatnik: @mjkkirschner , ha-ha... I see)))) actually, I don't know. It depends how small angle we want have between planes. Before we used  System.Math.PI / 1800.0 * System.Math.PI / 1800.0. Which I guess is 0.01 of radian, if I'm correct.

mjkkirschner: lets use the same tolerance as was used before if possible so we don't change any behavior here.

mjkkirschner:
@aosyatnik I think this looks very good. It should be tested thoroughly as this is a very common use case and one of the first things new Dynamo users try out is creating model curves in Revit.

so please make sure to run the ModelCurve and Rebinding Tests, and after changing the tolerance lets merge it.

aosyatnik:
@mjkkirschner , run tests, ModelCurve passed as they used to pass. Rebinding Tests passed as well.
image

PR #928, opened 2016-03-22
Magn 9748 Selection of item from dropdown list in "FamilyTypes" node is always reset to the first item.

rk-lakshmanan:
### Purpose

References: [MAGN-9748](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9748)
This PR is to fix the issue of drop down items reseting to the first item if the user does not select
any of the item. 
We also update the PopulateItem API such that the derived classes will be able to implement PopulateItemCore insteam of PopulateItem. Derived classes can handle the drop down item 
selection or leave it to the base class.

### Declarations

### Reviewers

@Benglin


Benglin:
Thanks @rk-lakshmanan for helping with this. Please make sure [this other PR](https://github.com/DynamoDS/Dynamo/pull/6288) gets merged first before this, and then cross-merge this to the other two branches.

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -41,11 +41,6 @@ protected RevitDropDownBase(string value) : base(value)
         void Controller_RevitDocumentChanged(object sender, EventArgs e)
         {
             PopulateItems();
-
```
rk-lakshmanan: I believe [this issue](https://github.com/DynamoDS/Dynamo/issues/6312) was caused by this if statement.

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -41,11 +41,6 @@ protected RevitDropDownBase(string value) : base(value)
...
-            if (Items.Any())
-            {
-                SelectedIndex = 0;
-            }
         }
```
rk-lakshmanan: I believe [this issue](https://github.com/DynamoDS/Dynamo/issues/6312) was caused by this if statement.

PR #932, opened 2016-03-27
Cones are tricky

mjkkirschner:
### Purpose
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9564

This PR fixes an issue exposed by a particular revit file where conical faces in Revit were not being converted to ASM geometry correctly.

There were three issues in the conical face conversion code... I'm not sure it worked previously.

* We used Cos(half angle) * height  to compute the radius - this is incorrect it should be Tan(halfAngle) * height. It would be Cos(half angle) * slanted height.

* The constructor for Cones using a coordinateSystem does not flip the cone if the Z is 0,0,-1... it just creates a cone in world space and then sets the CS Z axis to 0,0,-1, this means the conical face is upside down from the expected result.

* We were returning the flat cone bottom, there is a comment in the code that the second face returned from a cone solid in ASM is the flat face, and we were returning this... thats a planar face, not the conical one.

this pr fixes these issues:

screen shot 2016-03-27 at 4 39 51 pm
screen shot 2016-03-27 at 4 39 24 pm

### Declarations

### Reviewers

@sm6srw 
@pboyer  
### FYIs

@jnealb 
@pbidenko

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/SurfaceExtractor.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/SurfaceExtractor.cs
@@ -108,7 +108,7 @@ namespace Revit.GeometryConversion
 
             // Construct the radius
             var rad = Math.Tan(ang) * height;
-            var cone = Cone.ByCoordinateSystemHeightRadius(baseCS, height, rad);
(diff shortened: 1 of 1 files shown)


sm6srw:
@mjkkirschner LGTM

PR #933, opened 2016-03-28
Magn 6820 revit2017 Ellipse start angle changes depending on which radius is bigger.

rk-lakshmanan:
[MAGN-6820] (http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-6820)

This PR request is for the fix of start angle changing for ellipse arc when y-radius is greater.
The ProtoToRevit and RevitToProto are affected due to change in start angle. 
We solve this by removing the offset for start angle in RevitToProto. 
We set the axis of ellipseArc in revit to the plane's axis instead of the major/minor axis for ProtoToRevit.

Reviewer: @aparajit-pratap


rk-lakshmanan:
I have removed the additional commit for TestServices.dll.config file. The RTF test case passed as well.

PR #944, opened 2016-04-01
MAGN-8916:  Identical node description for "Wall Type" and "Floor Type" nodes

rk-lakshmanan:
References: [MAGN-8916] (http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-8916)

##Purpose
This PR is to modification correct the string description for Wall Types.

##Declarations

 The code base is in a better state after this PR
 Is documented according to the standards
 The level of testing this PR includes is appropriate
 User facing strings, if any, are extracted into *.resx files
 All tests pass using the self-service CI.
 Snapshot of UI changes, if any.
##Reviewers

@aparajit-pratap


aparajit-pratap:
@rk-lakshmanan thanks for making this change. Please keep in mind you need to cross-merge to the other Revit branches as well?

PR #950, opened 2016-04-04
Change file open to be executed immediate.

QilongTang:
### Purpose

MAGN-9824 First File, Open fails silently with Dynamo4Revit
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-9824

### Declarations

### Reviewers
@sm6srw 

### FYIs

@mjkkirschner @ramramps 

Change file open to be executed immediate.
Instead of putting the file open action into the idleActionQueue, execute it immediately so that exception handling will be consistent with DynamoStudio. Since file open right now does not call Revit API, so execution should be fairly safe.

Making the change in R2017 branch first, if everything is good, submissions to other branches will follow.

### DIFF
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -853,7 +853,9 @@ namespace Dynamo.Applications.Models
 
         protected override void OpenFileImpl(OpenFileCommand command)
         {
-            DynamoRevitApp.AddIdleAction(() => base.OpenFileImpl(command));
+            // MAGN-9824 changed file open to be executed right away instead of queue up.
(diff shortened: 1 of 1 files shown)


ramramps:
I am OK with this  change. But as Neal mentioned, can you please ensure the other task is good too? #
MAGN-5768

ramramps:
Also, I am not sure whether there are test cases coupled with this behavior. 
@Randy-Ma - I would recommend you to take a look at this.

sm6srw:
LGTM

Randy-Ma:
As I can remember, this won't impact existing test cases. :)

QilongTang:
Merging this one and testing same change in R2016, R2015.

PR #958, opened 2016-04-07
Fixed element types node error

rk-lakshmanan:
Purpose

This PR is for the fix for the [following issue](https://github.com/DynamoDS/Dynamo/issues/6380).
The problem was because the case where the SelectedIndex is -1 is not taken care of in the BuildOutputAst method for element types node.
The solution was to include a check to prevent the error.

Reviewers
@ke-yu 

FYIs

@Benglin

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitDropDown.cs
+++ b/src/Libraries/RevitNodesUI/RevitDropDown.cs
@@ -684,10 +684,22 @@ namespace DSRevitNodesUI
     {
         public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
         {
-            var typeName = AstFactory.BuildStringNode(Items[SelectedIndex].Name);
-            var assemblyName = AstFactory.BuildStringNode("RevitAPI");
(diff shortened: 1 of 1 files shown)


Benglin:
Hi @ke-yu, can I get your help in reviewing this instead? Thanks in advance for your help!

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -684,10 +684,22 @@ public class ElementTypes : AllChildrenOfType<Element>
     {
         public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
         {
-            var typeName = AstFactory.BuildStringNode(Items[SelectedIndex].Name);
-            var assemblyName = AstFactory.BuildStringNode("RevitAPI");
-            var functionCall = AstFactory.BuildFunctionCall(new Func<string,string,object>(Types.FindTypeByNameInAssembly) , new List<AssociativeNode>(){typeName, assemblyName});
-            return new []{AstFactory.BuildAssignment(GetAstIdentifierForOutputIndex(0), functionCall)};
+            AssociativeNode node;
+            if (SelectedIndex == -1)
```
ke-yu:
```
if (SelectedIndex < 0 || SelectedIndex >= Items.Count) 
{
    ...
}
else
{
    ...
}
```

PR #959, opened 2016-04-08
Fix the test case failure in CanOpenAndRunOtherFilesAfterOpeningFileWithSelectNode

rk-lakshmanan:
Purpose
This PR is to fix the test case failure in CanOpenAndRunOtherFilesAfterOpeningFileWithSelectNode under BugTest.cs. The problem was due to the wrong file name being used for the test case. 
Fixed the issue by changing the file name to Sample/Revit_color.dyn.

Reviewers:
@Randy-Ma 

FYI:
@sharadkjaiswal


Randy-Ma:
@rk-lakshmanan 
Great! Thanks a ton! LGTM
Could you also help merge this to other branches?

PR #979, opened 2016-04-26
Added new Family instance methods to place them ByFace

aparajit-pratap:
### Purpose

This PR adds 2 new nodes to place `FamilyInstance`s on a face reference. These methods are:
```
/// <summary>
        /// Place a Revit FamilyInstance given the FamilyType (also known as the FamilySymbol in the Revit API) 
        /// on a reference of another element using a line on that reference for its position
        /// </summary>
        /// <param name="familyType"></param>
        /// <param name="surface">A surface geometry reference</param>
        /// <param name="line">A line on the reference defining where the symbol is to be placed</param>
FamilyInstance ByFace(FamilyType familyType, Surface surface, Autodesk.DesignScript.Geometry.Line line)

and

/// <summary>
        /// Place a Revit FamilyInstance given the FamilyType
        /// using a location, reference direction
        /// </summary>
        /// <param name="familyType"></param>
        /// <param name="surface">A surface/face geometry reference object</param>
        /// <param name="location">Point on the reference where the instance is to be placed</param>
        /// <param name="referenceDirection">A vector that defines the direction of the family instance. Note that this direction
        /// defines the rotation of the instance on the reference, and thus cannot be parallel to the reference normal</param>
        /// <returns></returns>
       FamilyInstance ByFace(FamilyType familyType, Surface surface, Point location, 
            Vector referenceDirection)
```

### Reviewers
@sharadkjaiswal Reviewer 1
@Randy-Ma  Reviewer 2  

Randy, these methods are a very useful set that are missing from the Revit node library. @riteshchandawar and I also need them for our client project and thought it would be a good idea to introduce them in master. Please review this as this is my first stab at implementing Revit nodes, especially on the element binding part. Thanks!


src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -157,6 +167,88 @@ private void InitFamilyInstance(Autodesk.Revit.DB.FamilySymbol fs, XYZ pos)
...
+
+            InternalSetFamilyInstance(fi);
+
+            TransactionManager.Instance.TransactionTaskDone();
+
+            ElementBinder.SetElementForTrace(InternalElement);
+        }
+
+        private void InitFamilyInstance(FamilySymbol fs, Reference reference, XYZ location,
+            XYZ referenceDirection)
```
aparajit-pratap: I'm not sure how to set the reference direction on the internal family instance in the case of element binding?

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -157,6 +167,88 @@ private void InitFamilyInstance(Autodesk.Revit.DB.FamilySymbol fs, XYZ pos)
...
+            var oldFam =
+                ElementBinder.GetElementFromTrace<Autodesk.Revit.DB.FamilyInstance>(Document);
+
+            //There was a point, rebind to that, and adjust its position
+            if (oldFam != null)
+            {
+                InternalSetFamilyInstance(oldFam);
+                InternalSetFamilySymbol(fs);
+                InternalSetFace(reference);
+                InternalSetPosition(location);
```
aparajit-pratap: I'm not sure how to set the reference direction on the internal family instance for the case of element binding?

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -184,6 +276,29 @@ private void InternalSetPosition(XYZ fi)
...
+        private void InternalSetFace(Reference face)
+        {
+            if (InternalFamilyInstance.HostFace == face)
+                return;
+
+            TransactionManager.Instance.EnsureInTransaction(Document);
+
+            InternalFamilyInstance.get_Parameter(BuiltInParameter.FAMILY_WORK_PLANE_BASED).Set(face.ElementId);
+
+            TransactionManager.Instance.TransactionTaskDone();
```
aparajit-pratap: Not sure if this is the right way to set the face reference on the internal family instance?

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -184,6 +276,29 @@ private void InternalSetPosition(XYZ fi)
...
+            TransactionManager.Instance.EnsureInTransaction(Document);
+
+            //var lp = InternalFamilyInstance.get_Parameter(BuiltInParameter.CURVE_IS_REFERENCE_LINE).Set(pos.Reference.ElementId);
+            var lp = InternalFamilyInstance.Location as LocationCurve;
+            lp.Curve = pos;
+
+            TransactionManager.Instance.TransactionTaskDone();
+        }
+
+        private void InternalSetFace(Reference face)
```
sharadkjaiswal: Should we call this method InternalSetHostFace(), how do you validate that the input reference belongs to face?

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -184,6 +276,29 @@ private void InternalSetPosition(XYZ fi)
             TransactionManager.Instance.TransactionTaskDone();
         }
 
+        private void InternalSetPosition(Line pos)
+        {
+            TransactionManager.Instance.EnsureInTransaction(Document);
+
+            //var lp = InternalFamilyInstance.get_Parameter(BuiltInParameter.CURVE_IS_REFERENCE_LINE).Set(pos.Reference.ElementId);
+            var lp = InternalFamilyInstance.Location as LocationCurve;
+            lp.Curve = pos;
```
sharadkjaiswal: You can make a geometry comparison of the input line and the curve location, to avoid unwanted change notification.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
sharadkjaiswal: Looks like FAMILY_WORK_PLANE_BASED parameter accepts bool value. Have you tested it? https://knowledge.autodesk.com/search-result/caas/CloudHelp/cloudhelp/2015/ENU/Revit-API/files/GUID-DC143EB8-43CB-48AB-938E-7ADE3A9D2E63-htm.html

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
sharadkjaiswal: Updating the Location may not create a host relationship. @Randy-Ma, I am correct? @aparajit-pratap can you please verify this. Create a family on curve using Dynamo and then modify that curve in Dynamo. You will see that family instance is updated. Now close Dynamo and modify the curve in Revit if the family instance is updated then we have a correct hosting relationship.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
aparajit-pratap: You're right this does not work. Removing this code ...

test/Libraries/RevitNodesTests/Elements/FamilyInstanceTests.cs
```diff
@@ -124,5 +124,26 @@ public void FacingOrientation()
...
+            var familytype1 = FamilyType.ByName("WallFamily");
+            var plane1 = Autodesk.DesignScript.Geometry.Plane.XY();
+            var rectangle1 = Autodesk.DesignScript.Geometry.Rectangle.ByWidthLength(plane1, 1000, 1000);
+            var surface1 = Autodesk.DesignScript.Geometry.Surface.ByPatch(rectangle1);
+            var importinstance1 = Revit.Elements.ImportInstance.ByGeometry(surface1);
+            var t2 = importinstance1.Geometry();
+            var s = t2[0] as Autodesk.DesignScript.Geometry.Surface;
+            var point1 = Point.ByCoordinates(0, 0, 0);
+            var vector1 = Vector.YAxis();
+            var famInst = FamilyInstance.ByFace(familytype1, s, point1, vector1);
```
sharadkjaiswal: Can we directly use surface1 instead of s? Also it would be good to validate that famInst is hosted on the input surface.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -157,6 +167,86 @@ private void InitFamilyInstance(Autodesk.Revit.DB.FamilySymbol fs, XYZ pos)
...
+        private void InitFamilyInstance(FamilySymbol fs, Reference reference, Line pos)
+        {
+            //Phase 1 - Check to see if the object exists and should be rebound
+            var oldFam =
+                ElementBinder.GetElementFromTrace<Autodesk.Revit.DB.FamilyInstance>(Document);
+
+            //There was a point, rebind to that, and adjust its position
+            if (oldFam != null)
+            {
+                InternalSetFamilyInstance(oldFam);
```
sharadkjaiswal: Also we need to check if the oldFam is hosted on the same reference? If not, you may want to recreate a new family instance, else this may not be attached to the new face.

test/Libraries/RevitNodesTests/Elements/FamilyInstanceTests.cs
sharadkjaiswal: Thanks for adding this test, we need to have one more test to test the update scenario.

aparajit-pratap:
Adding a new task to write test case for ByFace with Line overload: http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10326

test/Libraries/RevitNodesTests/Elements/FamilyInstanceTests.cs
aparajit-pratap: `surface1` would not have a reference to a Revit face in which case the node will not work. The input surface needs to be associated with a Revit face.

PR #988, opened 2016-05-10
Check loaded assembly versions

mjkkirschner:
### Purpose

http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10072

Nate's addin is ultimately referencing an older version of RestSharp than we reference in multiple locations. There is most likely a missing method or interface in this assembly that throws an exception at various places (like in binding construction in WPF) -  We have not identified any great solutions to this problem in general, we still need to look into strong naming, keeping our own version of the open source libraries where possible, or different app domains.

As an alert to the user that Dynamo is in a bad state and to help them solve the issue this PR adds a subscription to Revit's app domain assembly.load event, and checks if any of the currently loaded dlls are referenced by the dll that is being loaded by the event. If any of the referenced dlls are indeed loaded and with incompatible major version numbers we show a warning to the user.

screen shot 2016-05-09 at 10 39 50 pm
screen shot 2016-05-09 at 10 16 33 pm

I've also added a list of names that we ignore - SSONET, RevitAPI etc... it seems that GregAuth.dll is referencing some old versions of these and the version naming scheme has changed since then, my assumption is that we know these assemblies work. 

When Nate's addin is present we get a bunch of rest sharp warnings where we reference it, and none when it is removed. 

### Notes:

* This does not stop Dynamo from loading but warns the user and suggests uninstalling addins.
* This should be tested with Zero touch nodes ( we could unsubscribe from this event at some point before shutdown if we don't want to get these messages when zero touch libraries load with problems)

### Declarations

### Reviewers

@sm6srw 
@QilongTang 
### FYIs

@jnealb 
@sharadkjaiswal 
@ikeough


src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -336,6 +344,42 @@ public static Assembly ResolveAssembly(object sender, ResolveEventArgs args)
...
+            //get all assemblies that are currently loaded into the appdomain.
+            var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies().Select(x => x.GetName()).ToList();
+            // ignore some assemblies(revit assemblies) that we know work and have changed their version number format or do not align
+            // with semantic versioning.
+            loadedAssemblies.RemoveAll(x => assemblyNamesToIgnore.Contains(x.Name));
+            //build dict- ignore those with duplicate names.
+            var loadedAssemblyDict = loadedAssemblies.GroupBy(assm => assm.Name).ToDictionary(g => g.Key, g => g.First());
+
+            foreach (var currentAssembly in assembly.GetReferencedAssemblies().Concat(new AssemblyName[] { assembly.GetName() }))
+            {
```
QilongTang: Based on our discussion, I am still curious about the N square complexity here. I'll try to install the profiling tool tomorrow. If not happening, we could also put timers around and do a couple run. Thanks for pulling this together.

sm6srw:
I think we should do a quick investigation and see if we can create our own app domain for dynamo before we commit to doing this.

ikeough:
@sm6srw It's a good question, and I've done that analysis in the past. I built an `AppDomain` at startup, and had Dynamo load in that `AppDomain`. I only got part way down the rabbit hole before coming to the conclusion that the amount of work required to create a separate app domain far exceeds the benefit. When you create an `AppDomain`, if you want to communicate with another`AppDomain` at all, like the one in which the addin is loaded, every type you wish to pass needs to be serializable. This means no passing any references to objects representing the addin context or any other Revit API type.

ikeough:
@sm6srw I would love it if someone with more Revit experience could tell me that I'm wrong here :)

sm6srw:
@ikeough well, you are right. I had a chat with Scott that is in charge of the Revit API and he told me that the API team did a similar analysis a couple of years ago with the same conclusion. It is not worth it. Scott also told me that they have seen conflicts like this a couple of times. It is not common and have been resolved by asking 3rd party developers to update their addins. @mjkkirschner I think we should go ahead with this PR. The only thing that I am a little worried about is that we are going to see a bunch of false positives. Lets try it and see what happens..

mjkkirschner:
@kronz lets discuss this message's exact wording for now and I'll merge this and see what false positives or other info we gain from testing this.

kronz:
@mjkkirschner can we have a basic warning message, with a "detail" section?

Dynamo has detected a conflict with <Addin> and may be unstable.  If any issues are detected, please uninstall this addin and contact the provider.

<check box> do not show me this message again.
<expandable dialog button> Details
While loading assembly <assembly name> Dynamo detected that the dependency <dependency> has already loaded with an incompatible or out of date version.  It is likely that another Revit addin has loaded this assembly, please try uninstalling other addins, and starting Dynamo again.  Dynamo may be unstable in the current state.

mjkkirschner:
@sm6srw @QilongTang 

please see:
https://github.com/DynamoDS/Dynamo/pull/6680

most changes here now reference code inside `Dynamo.Applications` and `Dynamo.Logging` apis to check for mismatched assemblies, get exceptions from that check, and then create notifications from them that bubble to the UI through the view extension so changes here are minimal.

src/transform_all.bat
```diff
@@ -26,6 +26,9 @@ IF EXIST "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\11.0\TextTra
 IF EXIST "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\12.0\TextTransform.exe" (
     set TEXTTRANSFORMPATH="%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\12.0\TextTransform.exe"
 )
+IF EXIST "%COMMONPROGRAMFILES(x86)%\microsoft shared\TextTemplating\14.0\TextTransform.exe" (
```
mjkkirschner: add visual studio 2015 support to bat file

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -183,10 +194,21 @@ public Result ExecuteCommand(DynamoRevitCommandData commandData)
...
+                DynamoApplications.Properties.Resources.MismatchedAssemblyVersionShortMessage,
+                x.Message));
+
                 TryOpenAndExecuteWorkspace(extCommandData);
 
                 // Disable the Dynamo button to prevent a re-run
                 DynamoRevitApp.DynamoButtonEnabled = false;
+
+                //unsubscribe to the assembly load
+                AppDomain.CurrentDomain.AssemblyLoad -= AssemblyLoad;
```
mjkkirschner: we only subscribe to this event during construction and load of Dynamo Model and View.

src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -23,6 +23,7 @@
 using Dynamo.Models;
 using RevitServices.EventHandler;
 using System.Collections;
+using System.Windows.Forms;
```
sm6srw: Is this still needed?

src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -283,14 +284,16 @@ private void UnsubscribeApplicationEvents()
...
         {
             AppDomain.CurrentDomain.AssemblyResolve += ResolveAssembly;
         }
 
-        private void UnsubscribeAssemblyResolvingEvent()
+      
+        private void UnsubscribeAssemblyEvents()
         {
             AppDomain.CurrentDomain.AssemblyResolve -= ResolveAssembly;
+            //AppDomain.CurrentDomain.AssemblyLoad -= AssemblyLoad;
```
sm6srw: Remove comment? It looks like this file could be left unchanged now but I don't think the renaming of the methods hurts...

sm6srw:
A few minor things then LGTM!

PR #991, opened 2016-05-12
MAGN-10068: Terminate the evaluation cycle due to transactions in Dynamo

sharadkjaiswal:
### Purpose

This PR fixes the crash due to infinite evaluation cycle caused due to element modification by SetRotation node.

DynamoRevit listens to document changed event on Revit to track the elements modified in Revit and marks the nodes responsible for creating the affected elements dirty so that those nodes can take part in next evaluation cycle. This is to support the ownership of Dynamo over the elements created by Dynamo in Revit, so if a user modifies the elements created by Dynamo using Revit interface then Dynamo should react and re-execute to force the rule set in Dynamo.

In this case of the defect, `FamilyInstance.SetRotation` node modifies the elements created by Dynamo and hence another update is triggered and it creates an infinite evaluation cycle. In fact, if the elements are getting modified due to Dynamo graph evaluation, then Dynamo should not mark the nodes dirty for re-execution in next evaluation cycle. Every evaluation cycle should complete the evaluation clean and no nodes should be left dirty after the evaluation.

The evaluation cycle is now broken by putting a check if the `ApplicationDocumentChanged` event  is triggered by the transaction created by Dynamo. We don't need to process the elements if the notification was due to a transaction completed by Dynamo.

### Declarations

### Reviewers
- [ ] @Randy-Ma 
- [ ] @ikeough 

### FYIs
- [ ] @kronz 
- [ ] @riteshchandawar 
- [ ] @Benglin

### DIFF
--- a/src/Libraries/RevitServices/Transactions/TransactionManager.cs
+++ b/src/Libraries/RevitServices/Transactions/TransactionManager.cs
@@ -226,6 +226,7 @@ namespace RevitServices.Transactions
         internal Transaction Transaction { get; set; }
         private readonly WarningHandler handler;
         internal readonly TransactionHandle Handle;
+        public static readonly string TransactionName = "Dynamo Script";
 
         internal TransactionWrapper()
         {
@@ -299,7 +300,7 @@ namespace RevitServices.Transactions
                     Transaction.Dispose();
                 }
 
-                Transaction = new Transaction(document, "Dynamo Script");
+                Transaction = new Transaction(document, TransactionName);
                 Transaction.Start();
 
                 FailureHandlingOptions failOpt = Transaction.GetFailureHandlingOptions();
--- a/src/Libraries/RevitServices/Transactions/TransactionManager.cs
+++ b/src/Libraries/RevitServices/Transactions/TransactionManager.cs
@@ -226,7 +226,7 @@ namespace RevitServices.Transactions
         internal Transaction Transaction { get; set; }
         private readonly WarningHandler handler;
         internal readonly TransactionHandle Handle;
-        public static readonly string TransactionName = "Dynamo Script";
+        public static readonly string TransactionName = "Dynamo-51297CB5 Script";
 
         internal TransactionWrapper()
         {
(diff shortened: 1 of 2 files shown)


Randy-Ma:
:+1: :+1: Wow! A simple but super cool fix! @ikeough This is also addressing some of the infinite loop issues in a very nice way. Two suggestions: 1). can we revisit the old fix which is to prevent the infinite loop and maybe some of them can be reverted? 2). shall we have a better transaction name? Other than these, it looks good to me! Thanks again!

mjkkirschner:
@sharadkjaiswal I imagine a graph where we create a direct shape or another element. We also have a node that gathers all elements in the Revit Model. When the element is added we should currently trigger the get all elements node, updating our list of elements. This behavior will no longer work with this change, correct?

ikeough:
@sharadkjaiswal @Randy-Ma This is very cool. Let's get this PR in then open a separate one to revisit the other cases. Are those cases centralized or do we have specific logic per node?

sharadkjaiswal:
@mjkkirschner, @ikeough We have there are two kinds of nodes which are taking part in an evaluation. One type of node that creates or modify the Revit document/element and another set of nodes are just reading data from the document/element. In this particular case, we don't want to re-execute the nodes that modify Revit document/element, however, there is no harm in letting the read operation nodes run. To effectively classify such nodes and enable/disable re-execution selectively will be more involved.

First, we need to classify our nodes appropriately and then we need to propagate the context (list of transactions) to the level where we are marking the nodes dirty.

I propose that let's get this change and see user feedback and work accordingly. What do you say?

src/Libraries/RevitServices/Transactions/TransactionManager.cs
```diff
@@ -226,6 +226,7 @@ public class TransactionWrapper
         internal Transaction Transaction { get; set; }
         private readonly WarningHandler handler;
         internal readonly TransactionHandle Handle;
+        public static readonly string TransactionName = "Dynamo-51297CB5 Script";
```
sharadkjaiswal: Unique transaction name.

PR #1009, opened 2016-05-26
MAGN-10128 Revit model only gives NULL when its components are selected

sm6srw:
### Purpose

The purpose of this PR is to make the materials manager more robust if it 1) for some reason fails to initialize or 2) encounter projects with non-unique localized names on materials.

Things we have seen when the materials manager fails includes crashes in Revit preview and nodes that stops generating geometry until Dynamo is restarted.. 
.
* Make sure that the materials manager have working default values even
if it for some reason fail to initialize.
* Stop using dictionaries because material names are not guaranteed to be
unique.
* Fix nasty bug when the material id for the DynamoError material was
assigned to the wrong property if it already existed.
* Fixed bug when GStyle lookup for Generic model fails on localized
projects.

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@jnealb 

This PR is for 2017.


src/Libraries/RevitServices/Materials/Materials.cs
```diff
@@ -10,9 +10,9 @@ public class MaterialsManager
     {
         private static MaterialsManager instance;
 
-        public ElementId DynamoMaterialId { get; private set; }
-        public ElementId DynamoErrorMaterialId { get; private set; }
-        public ElementId DynamoGStyleId { get; private set; }
+        public ElementId DynamoMaterialId { get; private set; } = ElementId.InvalidElementId;
+        public ElementId DynamoErrorMaterialId { get; private set; } = ElementId.InvalidElementId;
+        public ElementId DynamoGStyleId { get; private set; } = ElementId.InvalidElementId;
```
sm6srw: Make sure we have valid invalid values by default.

src/Libraries/RevitServices/Materials/Materials.cs
```diff
@@ -51,28 +51,27 @@ public static void Reset()
...
         {
-            Dictionary<string, Material> materials
-            = new FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument)
-              .OfClass(typeof(Material))
-              .Cast<Material>()
-              .ToDictionary<Material, string>(
-                e => e.Name);
+            var materials = new FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument)
+                .OfClass(typeof(Material))
+                .Cast<Material>();
```
sm6srw: Stop using dictionaries.

src/Libraries/RevitServices/Materials/Materials.cs
```diff
@@ -89,17 +88,16 @@ private void FindOrCreateDynamoMaterial()
...
+               .Cast<Material>();
+
+
+            // First try to get or create the Dynamo material
+            var material = materials.FirstOrDefault(x => x.Name == "DynamoError");
+            if (material != null)
             {
-                DynamoMaterialId = materials["DynamoError"].Id;
+                DynamoErrorMaterialId = material.Id;
                 return;
```
sm6srw: Fix nasty bug!

src/Libraries/RevitServices/Materials/Materials.cs
```diff
@@ -118,18 +116,14 @@ private void FindOrCreateDynamoErrorMaterial()
 
         private void FindDynamoGraphicsStyle()
         {
-            var styles = new FilteredElementCollector(DocumentManager.Instance.CurrentUIDocument.Document);
-            styles.OfClass(typeof(GraphicsStyle));
-
-            var gStyle = styles.ToElements().FirstOrDefault(x => x.Name == "Generic Models");
-
-            if (gStyle != null)
+            var genericModelCategory = Category.GetCategory(DocumentManager.Instance.CurrentDBDocument, BuiltInCategory.OST_GenericModel);
```
sm6srw: Look-up anything built in by name is usually a bad idea. Will not work on localized templates/projects.

jnealb:
@sm6srw @ikeough @monikaprabhu This is very interesting; we had a discussion with Sasha Crotty about this when did the Dynamo localization back in 2015, great to see it get cleaned up/hardened.

src/Libraries/RevitServices/Materials/Materials.cs
```diff
@@ -118,18 +116,14 @@ private void FindOrCreateDynamoErrorMaterial()
...
+            if(genericModelCategory != null)
             {
-                DynamoGStyleId = gStyle.Id;
-            }
-            else
-            {
-                DynamoGStyleId = ElementId.InvalidElementId;
+                var gStyle = genericModelCategory.GetGraphicsStyle(GraphicsStyleType.Projection);
```
mjkkirschner: is the projection style the standard / regular style?

test/Libraries/RevitServicesTests/RevitServicesTests.cs
```diff
@@ -328,5 +329,34 @@ public void TwoInstancesOfSerializableIdAreEqual()
 
         }
 
+        [Test]
+        [Category("UnitTests")]
+        public void MaterialManagerCheckDefaultValues()
```
mjkkirschner: probably not necessary, but it might be worth looking into if we can start the revit in localized german mode using RTF and test this stuff or, can the API switch the localization mode?

mjkkirschner:
@sm6srw looks really solid, just two questions. then LGTM

src/Libraries/RevitServices/Materials/Materials.cs
sm6srw: There are two style types for model category styles: projection & cut. Projection is mandatory. Cut is optional. It allows an element do be drawn with a different style when it is cut in some way compared to when it is only projected.

test/Libraries/RevitServicesTests/RevitServicesTests.cs
sm6srw: I will take a quick look.

sm6srw: @mjkkirschner I added two more tests with Revit 2015 projects that covers the part with duplicated material names. No need to run localized Revit. I will merge this soon if you have no objections and then start cherry-picking for 2015 and 2016 later today if everything looks good.

mjkkirschner:
all looks good to me - but you may not need to cherry pick to 2015, check with @kronz.

sm6srw:
@mjkkirschner @kronz I already have it ready to go for 2015 so no big deal...

Issue #939, opened 2016-03-28
Get Family Parameter node improvement


ksobon:
Team, 

cc: @ikeough @mjkkirschner @kronz  (I am not sure who's working on this)
I think there is a clean way to return parameters that you could implement in Get Family Parameter node. I might be wrong on this, but it has worked for me so far. 

This is the node that I am talking about: https://github.com/DynamoDS/DynamoRevit/blob/91f35d6e3762aa8b7da19043fde41e73e2a86f4d/src/Libraries/RevitNodesUI/RevitDropDown.cs

Please consider this instead: 

Don't cast elements passed in as input into <code>FamilyInstance</code> or <code>WallType</code> etc. As a matter of fact just leave them be as <code> Element </code>. 

Then you can get all parameters from an element and if that element is an instance you can also get its type parameters using this utility method: 

```C#
private void AddElementParams(Element e)
        {
            foreach (Parameter p in e.Parameters)
            {
                if (!(p.StorageType == StorageType.None))
                {
                    AddDropDownItem(p);
                }
            }
            // if element can have type assigned it's safe to assume that it's an instance
            // and add type parameters to the list
            if (e.CanHaveTypeAssigned())
            {
                ElementType et = DocumentManager.Instance.CurrentDBDocument.GetElement(e.GetTypeId()) as ElementType;
                if (et != null)
                {
                    AddTypeParams(et);
                }
            }
        }

private void AddTypeParams(ElementType et)
        {
            foreach (Parameter p in et.Parameters)
            {
                if (p.StorageType == StorageType.None)
                    continue;

                AddDropDownItem(p);
            }
        }
```
As you can see I am making a simple check <code> e.CanHaveTypeAssigned()</code> to see if element has also element type parameters available and this seems to be just good enough. 

This way, I can handle both, instances, as well as types being passed in as inputs and do away with type specific casting. 

Please let me know if this helps. Here are some results: 

![paramlister](https://cloud.githubusercontent.com/assets/529781/14086725/90fd934e-f4f4-11e5-897e-33d0334f6695.PNG)

kronz:
@ksobon @ikeough @Randy-Ma Hey Konrad.  This looks very cool.  It's been a long time since we looked at this one.  Get Family Paramters works really differently than anything else, needing to run once before the UI can be populated has created some problems in the past.  But I love the idea of having a more generalized Get Parameters functionality. Randy/Ian, any obvious problems with this approach?

[...]

PR #1015, opened 2016-06-01
MAGN-10193: Get Family Parameter node enhancement

sharadkjaiswal:
### Purpose

This PR addresses recent regression where Get Family Parameter node hangs DynamoRevit when it's input port is connected to a valid input and it also implements the suggestion provided in #939.

Get Family Parameter is a unique node where based on input provided, it populates the combo box for selection, that means it calls for PopulateItems() when node's CachedValue is updated. With recent changes during the PopulateItems() call the selection index is modified which again marks the node, dirty. This gets into an infinite cycle.

The infinite cycle can be broken by inspecting the input, it the input element is same as old element then we don't really need to call PopulateItems() again and hence no selection change and no further evaluation.

This node now also shows up read-only parameters as well as the parameters of different family types.

**Old**
image

**New Implementation**
image

### Declarations

### Reviewers

- [ ] @ikeough 
- [ ] @Randy-Ma 

### FYIs

@mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitDropDown.cs
+++ b/src/Libraries/RevitNodesUI/RevitDropDown.cs
@@ -30,6 +30,19 @@ using BuiltinNodeCategories = Revit.Elements.BuiltinNodeCategories;
 
 namespace DSRevitNodesUI
 {
+    public class DropDownItemEqualityComparer : IEqualityComparer<DynamoDropDownItem>
+    {
+        public bool Equals(DynamoDropDownItem x, DynamoDropDownItem y)
+        {
+            return string.Equals(x.Name, y.Name);
+        }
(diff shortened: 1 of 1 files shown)


kronz:
It looks like this node is now extending beyound families, right?  We should rename it to Get Element Parameter and create mgrations if that is the case

sharadkjaiswal:
@kronz This node will not work for all the elements, such as Element Type or Category. Isn't wall or wall type is a family(type)??

kronz:
Wall is a system family, like stairs, and rooms, which are found in Element Types selectors.  I always understood GetFamilyParameters to only work on loadable families.  We can keep the name if it isn't really all elements

PR #1016, opened 2016-06-01
Revit2016 rooms

moethu:
### Purpose

Expose Revit Room element to Dynamo

### Declarations

### Reviewers

### FYIs


src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -0,0 +1,220 @@
...
+            }
+            else
+            {
+                // Update Location only
+                Autodesk.Revit.DB.LocationPoint point = (Autodesk.Revit.DB.LocationPoint)RoomElem.Location;
+                point.Point = location;
+            }
+
+            // Apply name and number if set
+            if (name != "") { RoomElem.Name = name; }
```
mjkkirschner: maybe use string.Empty here, and also can you put the If on separate lines. I think this is in our style guide in the wiki - if not apologies.

src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -0,0 +1,220 @@
...
+            if (number != "") { RoomElem.Number = number; }
+
+            InternalSetElement(RoomElem);
+
+            // Commit transaction
+            TransactionManager.Instance.TransactionTaskDone();
+
+            if (RoomElem != null)
+            {
+                ElementBinder.CleanupAndSetElementForTrace(document, this.InternalElement);
```
mjkkirschner: so I'm a bit confused by this logic - can you clarify why if the room element is not null, meaning it was been created or pulled from trace that you then destroy it? and the else statement seems it will never be true...

Can a room not be rebound? I would think that you should get rid of the else statement and use SetElement for trace here instead.

`CleanupAndSetElement` destroys the element - meaning that the ID of the room will change if it was pulled from trace right?

src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -0,0 +1,220 @@
...
+
+        /// <summary>
+        /// Create a Revit Room Element
+        /// </summary>
+        /// <param name="level">Level the room is hosted on</param>
+        /// <param name="location">Location for the center of the room</param>
+        /// <param name="name">Room name</param>
+        /// <param name="number">Room number</param>
+        /// <returns></returns>
+        public static Room Create(Elements.Level level, Autodesk.DesignScript.Geometry.Point location, string name = "", string number = "")
```
mjkkirschner: this should use the "by" naming if its exposed as a node.

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,167 @@
...
+
+using Dynamo.Utilities;
+using Dynamo.Models;
+using ProtoCore.AST.AssociativeAST;
+
+namespace DSRevitNodesUI
+{
+
+
+    public abstract class CustomRevitElementDropDown : RevitDropDownBase
```
mjkkirschner: is this used by the room element somewhere?

test/Libraries/RevitNodesTests/Elements/RoomTests.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using RTF.Framework;
+
+namespace RevitNodesTests.Elements
+{
+    [TestFixture]
+    public class RoomTests : RevitNodeTestBase
+    {
+        [Test]
+        [TestModel(@".\emptyAnnotativeView.rfa")]
```
mjkkirschner: should this be .rvt? you have not attached a .rfa file that I can see.

src/Libraries/RevitNodesUI/GenericClasses.cs
mjkkirschner: please add comments to any public classes (summaries)

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,167 @@
...
+            var node = AstFactory.BuildFunctionCall("Revit.Elements.ElementSelector", "ByElementId",
+                new List<AssociativeNode> { AstFactory.BuildIntNode(Id.IntegerValue) });
+
+            // Return the selected element
+            return new[] { AstFactory.BuildAssignment(GetAstIdentifierForOutputIndex(0), node) };
+        }
+    }
+
+
+    public abstract class CustomGenericEnumerationDropDown : RevitDropDownBase
```
mjkkirschner: summary please

mjkkirschner:
@moethu thanks for the hard work here. I have a few comments and clarifications.
Also, I have not reviewed the generic element class or the UI nodes yet. I will do this next.

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: @mjkkirschner the generic dropdowns are not being used by the Room element but they are part of the later package. I added them to commit some more nodes.

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// Generic Revit Element Class Dropdown Node
+        /// </summary>
+        /// <param name="name">Name of the Node</param>
+        /// <param name="elementType">Type of Revit Element to display</param>
+        public CustomRevitElementDropDown(string name, Type elementType) : base(name) { this.ElementType = elementType; PopulateItems(); }
+
+        /// <summary>
+        /// Constant message for missing Types
+        /// </summary>
+        private const string noTypes = "No Types available.";
```
mjkkirschner: this should be moved to the .resx files in properties/resources for localization.

mjkkirschner: let me know if you need more info here.

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        /// <summary>
+        /// Type of Element
+        /// </summary>
+        public Type ElementType;
```
mjkkirschner: please make this a property, as a general rule make it as conservative as possible - if it can have a private or readonly setter that is preferable.

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// </summary>
+        public void PopulateItems()
+        {
+            if (this.ElementType != null)
+            {
+                // Clear the Items
+                Items.Clear();
+
+                // Set up a new element collector using the Type field
+                var fec = new RVT.FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument).OfClass(ElementType);
```
mjkkirschner: is it possible that DM.instance.CurrentDB will be null?

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+                // If there is nothing in the collector add the missing Type message to the Dropdown menu.
+                if (fec.ToElements().Count == 0)
+                {
+                    Items.Add(new CoreNodeModels.DynamoDropDownItem(noTypes, null));
+                    SelectedIndex = 0;
+                    return;
+                }
+
+                // if the elementtype is a RebarHookType add an initial "None" value which does not come from the collector
+                if (this.ElementType.FullName == "Autodesk.Revit.DB.Structure.RebarHookType") Items.Add(new CoreNodeModels.DynamoDropDownItem("None", null));
```
mjkkirschner: "Autodesk.Revit.DB.Structure.RebarHookType" should be stored as a constant somewhere or better- extracted from a Revit type somehow if possible.

"None" should be moved to a resx for localization, but can you explain why this is being done?

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        /// <summary>
+        /// Cast the selected element to a dynamo node
+        /// </summary>
+        public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
+        {
+            // If there are no elements in the dropdown or the selected Index is invalid return a Null node.
+            if (Items.Count == 0 || 
+            Items[0].Name == noTypes ||
+            SelectedIndex == -1 || Items[SelectedIndex].Name == "None")
```
mjkkirschner: I see the "none" type now is used for this comparison... so it shouldn't goto the .resx unless exposed to the user, instead just pull it out as a constant. any reason not to use `null`?

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+            {
+                return new[] { AstFactory.BuildAssignment(GetAstIdentifierForOutputIndex(0), AstFactory.BuildNullNode()) };
+            }
+
+            // Assume we are going to cast a Revit Element
+            Type genericElementType = typeof(Autodesk.Revit.DB.Element);
```
mjkkirschner: this is not used - did you intend to use it below to do a cast?

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            // Cast the selected object to a Revit Element and get its Id
+            Autodesk.Revit.DB.ElementId Id = ((Autodesk.Revit.DB.Element)Items[SelectedIndex].Item).Id;
+
+            // Select the element using the elementIds Integer Value
+            var node = AstFactory.BuildFunctionCall("Revit.Elements.ElementSelector", "ByElementId",
```
mjkkirschner: you could call this elementSelectionFunction

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// Generic Enumeration Dropdown
+        /// </summary>
+        /// <param name="name">Node Name</param>
+        /// <param name="enumerationType">Type of Enumeration to Display</param>
+        public CustomGenericEnumerationDropDown(string name, Type enumerationType) : base(name) { this.EnumerationType = enumerationType; PopulateItems(); }
+
+        /// <summary>
+        /// Type of Enumeration
+        /// </summary>
+        public Type EnumerationType;
```
mjkkirschner: same, please convert to most conservative property possible.

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+                Items = Items.OrderBy(x => x.Name).ToObservableCollection();
+            }
+        }
+
+        /// <summary>
+        /// Assign the selected Enumeration value to the output
+        /// </summary>
+        public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
+        {
+            // If there the dropdown is still empty try to populate it again
```
mjkkirschner: ~~there~~

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+            if (Items.Count == 0 || Items.Count == -1)
+            {
+                PopulateItems();
```
mjkkirschner: is this logic taken from somewhere else? I'm just curious if it could cause infinite loops.

src/Libraries/RevitNodesUI/RevitTypes.cs
```diff
@@ -0,0 +1,125 @@
...
+    [NodeDescription("Horizontal Text Alignment Style")]
+    [IsDesignScriptCompatible]
+    public class HorizontalAlignment : CustomGenericEnumerationDropDown
+    {
+        public HorizontalAlignment() : base("Horizontal Alignment", typeof(Autodesk.Revit.DB.HorizontalAlignmentStyle)) { }
+    }
+
+    [NodeName("Vertical Text Alignment Style")]
+    [NodeCategory("Revit.Rebar")]
+    [NodeDescription("Vertical Text Alignment Style")]
```
mjkkirschner: I would be a little more verbose with these descriptions.

mjkkirschner: I believe these strings should also be added to the resx dictionaries.

src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -757,4 +757,16 @@ public SelectFaces()
                 "Select faces.",
                 "Faces") { }
     }
+
+    [NodeName("Select Edges"), NodeCategory(Revit.Elements.BuiltinNodeCategories.REVIT_SELECTION),
+NodeDescription("SelectEdgesDescription", typeof(DSRevitNodesUI.Properties.Resources)), IsDesignScriptCompatible]
```
mjkkirschner: same here, for localization.

src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -757,4 +757,16 @@ public SelectFaces()
...
+    public class SelectEdges : ReferenceSelection
+    {
+        public SelectEdges()
+            : base(
+                SelectionType.Many,
+                SelectionObjectType.Edge,
+                "Select edges.",
+                "Edges") { }
```
mjkkirschner: is this propagated to the user?

test/Libraries/RevitNodesTests/Elements/RoomTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            Level lvl = Level.ByElevation(0);
+            Assert.IsNotNull(lvl);
+
+            var room = Revit.Elements.Room.ByLocation(lvl,Point.ByCoordinates(0, 0, 0),"myRoom", "myNumber");
+            
+            Assert.NotNull(room);
+            Assert.AreEqual(room.InternalRevitElement.Name, "myRoom");      
+        }
+
```
mjkkirschner: can you add an additional test that selects a room, and checks if the wrapper is as expected?

mjkkirschner: you can create a Dynamo .dyn, load it, and run it, then check that the returned wrapper is valid.

mjkkirschner:
@moethu nice work! I have some comments on properties and strings/constants for you to check out.

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: @mjkkirschner , right - added a check for this:
```
                // If the active doc is null, throw an exception
                if (DocumentManager.Instance.CurrentDBDocument == null)
                {
                    throw new Exception(Properties.Resources.NoActiveDocumentFound);
                }
```

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: @mjkkirschner , the Rebar constructor accepts null as a hook type value and translates this into a rebar hook type called "none". But querying the existing hook types won't give you "none" as an existing hook type since its just null. I don't know if there are other constructors accepting null as well.
I could drop it entirely or always add a "none" value returning null. What would make more sense?

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: yes, that was some legacy... dropped it.

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: @mjkkirschner do you mean the RevitElementSelectionHelper?

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: if the EnumerationType is null or the enumeration is empty it would lead to an empty list of elements. But I'm not sure if BuildOutputAst is triggered on user interaction only? I guessed so. So far I has been around some time in the Rebar package without causing any infinite loops. I will add a check for null and empty enums:
```if (this.EnumerationType != null && Enum.GetNames(this.EnumerationType).Length > 0)```

File: src/Libraries/RevitNodesUI/Selection.cs (hunk 19 above)
Comment (moethu) [inline src/Libraries/RevitNodesUI/Selection.cs]: yes, why?

File: src/Libraries/RevitNodes/Elements/Room.cs [hunk 21]
```diff
@@ -0,0 +1,219 @@
...
+
+            if (RoomElem == null)
+            {
+                // Create new Room element
+                RoomElem = document.Create.NewRoom(level, new Autodesk.Revit.DB.UV(location.X, location.Y));
+            }
+            else
+            {
+                // Update Location only
+                Autodesk.Revit.DB.LocationPoint point = (Autodesk.Revit.DB.LocationPoint)RoomElem.Location;
```
sharadkjaiswal: How do you update the level? What if Init is called with different level?

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        protected override CoreNodeModels.DSDropDownBase.SelectionState PopulateItemsCore(string currentSelection)
+        {
+            PopulateItems();
+            return SelectionState.Done;
+        }
+        /// <summary>
+        /// Populate the Dropdown menu
+        /// </summary>
+        public void PopulateItems()
```
sharadkjaiswal: I think the base class also has PopulateItems() method, do you intend to hide the base class implementation? I think this belongs to PopulateItemsCore().

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,184 @@
...
+        /// </summary>
+        private Type ElementType
+        {
+            get;
+            set;
+        }
+
+        protected override CoreNodeModels.DSDropDownBase.SelectionState PopulateItemsCore(string currentSelection)
+        {
+            PopulateItems();
```
sharadkjaiswal: PopulateItems() in base class calles the PopulateItemsCore() so it would create a recursive call.

moethu:
@mjkkirschner , thanks for you review, applied all changes. Please let me know how to proceed with the null value for "none" type.

src/Libraries/RevitNodes/Elements/Room.cs
moethu: @sharadkjaiswal , good point. haven't found a way to do this via API at the moment. the properties Level and LevelID are readonly. I could create a new room and delete the old one but this might affect existing properties.

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: @sharadkjaiswal yes that's true but the PopulateItems of the base class is called before the constructor because its triggered by the base constructor. At this point the EnumerationType is not set yet which I need for populating. That's why I have to call Populate again after setting the EnumerationType.

src/Libraries/RevitNodesUI/GenericClasses.cs
moethu: @sharadkjaiswal it does call the correct PopulateItems() method and runs correctly at the moment, but you are right - I will change the name.

mjkkirschner:
@moethu I think this looks pretty good and can be merged, but I'd like to wait until some other PRs are brought in and we can verify that this repo is stable.

src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -0,0 +1,220 @@
...
+        internal Autodesk.Revit.DB.Architecture.Room InternalRevitElement
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Reference to the Element
+        /// </summary>
+        public override Autodesk.Revit.DB.Element InternalElement
```
sharadkjaiswal: [SupressImportIntoVM] is required for InternalElement property and not for InternalRevitElement property. InternalRevitElement property is already marked internal.

src/Libraries/RevitNodes/Elements/Room.cs
sharadkjaiswal: If there is no way to update the level, then we should create new room element if the input level is different than the one from existing room element.

test/Libraries/RevitNodesTests/Elements/RoomTests.cs
sharadkjaiswal: Also, it would be good to have tests to validate the update and element binding. Create a DYN file for room creation, run it and then modify some of the input parameters run it again. Validate that the hosting is maintained, the same room element is updated etc.

src/Libraries/RevitNodes/Elements/Room.cs
mjkkirschner: as @sharadkjaiswal pointed out, internal access properties do not need to be suppressed from import, but I believe we should suppress the import of the public property right below.

test/Libraries/RevitNodesTests/Elements/RoomTests.cs
moethu: @sharadkjaiswal like this?
2016-07-18_14h50_46

moethu:
@mjkkirschner @sharadkjaiswal applied the changes and added a dyn test file

mjkkirschner:
@moethu thanks for making the changes, can you please fix these conflicts here and on 2017 branch  and we will merge.

mjkkirschner:
hmm, I'm not sure what is up with the diff here.... 175 files changed + 4000 changes seems like either the diff is wrong or the merge went a bit wrong... I'll take a look but please investigate as well.

mjkkirschner:
diff appears to just be incorrect on github, appears fine through git CLI.

moethu:
@mjkkirschner sorry, I had to rebase it due to changes in Revit2016, guess that's why the github diff shows many changes.

mjkkirschner:
please make sure the 2017 branch is matched up with all changes in this one, and make a 2018 PR as well.  I know you do not have access yet, but we can deal with changes there later, I'm more concerned with losing some of your work on one branch.

moethu:
@mjkkirschner will do, please have a look at the Enum parser I added in the last commit:
https://github.com/DynamoDS/DynamoRevit/pull/1016/commits/9092827cb0b75dbc0c0bde5b249f5af48c0d77ac

src/Libraries/RevitNodes/Elements/InternalUtilities/ParseEnum.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class ParseEnum
+    {
+        /// <summary>
+        /// Parse any Revit Enum by String
+        /// </summary>
+        /// <param name="value">enum string value</param>
+        /// <param name="typeName">full type name</param>
+        /// <returns></returns>
+        [IsVisibleInDynamoLibrary(false)]
+        public static object ByString(string value, string typeName)
```
mjkkirschner: will this always return an int?

src/Libraries/RevitNodes/Elements/InternalUtilities/ParseEnum.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            // Get Revit Assembly
+            var revitAssembly = System.Reflection.Assembly.GetAssembly(typeof(Autodesk.Revit.DB.Document));
+
+            // Get Type from Assembly by string
+            Type type = revitAssembly.GetType(typeName);
+
+            if (type != null)
+            {
+                return Enum.Parse(type, value);
```
mjkkirschner: can you add some tests for this class specifically... what happens if this is run with an invalid string name for the enum type.

src/Libraries/RevitNodes/Elements/InternalUtilities/ParseEnum.cs
moethu: usually I would implement it using a generic Type, 
public static T ByString<T>(...) where T : ...
But since I'm calling this using AstFactory.BuildStringNode, I guess it won't work.

moethu:
@mjkkirschner added tests & check if enum is defined.

mjkkirschner:
@moethu I'm going to merge this since it's a dep. of other PRs, can you please create a 2018 branch? I only see this one and the 2017 PR.

PR #1017, opened 2016-06-01
Revit2017 rooms

moethu:
### Purpose

Exposing Revit Room element to Dynamo.

### Declarations

### Reviewers

### FYIs


src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -0,0 +1,219 @@
...
+        internal Autodesk.Revit.DB.Architecture.Room InternalRevitElement
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Reference to the Element
+        /// </summary>
+        public override Autodesk.Revit.DB.Element InternalElement
```
sharadkjaiswal: You need to put attribute [SupressImportIntoVM] on the public properties or methods that uses types from Autodesk.Revit.DB namespace. We don't want this type to be imported in DesignScript VM.

src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -0,0 +1,219 @@
...
+            }
+            else
+            {
+                // Update Location only
+                Autodesk.Revit.DB.LocationPoint point = (Autodesk.Revit.DB.LocationPoint)RoomElem.Location;
+                point.Point = location;
+            }
+
+            // Apply name and number if set
+            if (name != string.Empty)
```
sharadkjaiswal: better to use string.IsNullOrEmpty(name).

src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Get room name
+        /// </summary>
+        public string GetName
```
sharadkjaiswal: Property name should not include Get. Can we have Name, Number,Area,Height etc. as property?

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        protected override CoreNodeModels.DSDropDownBase.SelectionState PopulateItemsCore(string currentSelection)
+        {
+            PopulateItems();
+            return SelectionState.Done;
+        }
+        /// <summary>
+        /// Populate the Dropdown menu
+        /// </summary>
+        public void PopulateItems()
```
sharadkjaiswal: Base class already has an implementation of PopulateItems() which calls PopulateItemsCore() method. You can move the implementation of PopulateItems() to PopulateItemsCore() method.

src/Libraries/RevitNodesUI/GenericClasses.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            if (this.ElementType != null)
```
sharadkjaiswal: You may consider if(this.ElementType == null) return;

moethu:
@sharadkjaiswal thanks for the comments - I haven't added the changes from the 2016 branch yet. Will do this now. They incorporate Michael's review.

About the [SupressImportIntoVM] Attribute - my understanding was that internal properties aren't displayed in Dynamo anyway, since none of the other elements have this attribute.

moethu:
@sharadkjaiswal like this? https://github.com/DynamoDS/DynamoRevit/pull/1017/commits/eaa5aca4fa783a7237bb99d034c3f6503a1f7cbd

src/Libraries/RevitNodes/Elements/Room.cs
```diff
@@ -21,6 +21,7 @@ public class Room : Element
         /// <summary>
         /// Internal reference to the Revit Element
         /// </summary>
+        [SupressImportIntoVM]
```
sharadkjaiswal: If the method, property or class is internal then you don't need this attribute. But if it's public then you do need to specify this attribute when you are using Autodesk.Revit.DB namespace type in your method/property.

mjkkirschner:
@moethu has this been updated with changes from the 2016 branch?

moethu:
@mjkkirschner I will check. let me cherry pick the latest commits from 2016 first. Is the 2016 version approved so far?

mjkkirschner:
@moethu yes.

src/Libraries/RevitNodes/Elements/Room.cs
mjkkirschner: this is still missing the Suppress import attribute.

mjkkirschner:
@moethu there are some merge conflicts on this PR, and a missing suppression attribute noted above. Can you try to sync up 2016 and 2017 branches and then we can get this merged to avoid any more code rot.

moethu:
@mjkkirschner applied all changes and had to rebase - same like 2016

mjkkirschner:
github reports merge conflicts here :(

moethu:
@mjkkirschner  resolved

PR #1025, opened 2016-06-07
MAGN-10068 Fix evaluation cycle due to element modification in Dynamo.

sharadkjaiswal:
### Purpose
This PR is in continuation of #991 where element modification caused by SetFamilyRotation triggers an infinite evaluation cycle. This happens because ElementsModified event marks all the nodes related to modified element dirty and forces re-execution. In fact, the constructor nodes are the one that causes this issue. In this case FamilyInstance.ByPoint() node creates a new family instance and resets the rotation value and when SetRotation is called it modifies the family instance again to get back the desired rotation.

In order to avoid re-execution of constructor nodes in Dynamo Transaction context, we need to pass the transaction-specific context to the event handlers so that the event handlers can take a decision on when to skip constructor nodes.

- RevitServicesUpdater now provides a new event `ElementsUpdated(object sender, ElementUpdateEventArgs args)` and marks all the other events such as ElementsAdded, ElementsModified and ElementsDeleted as Obsolete to maintain the binary compatibility. 
- This PR also updates all the other clients of the Obsolete events to use ElementsUpdated event.

### Declarations

### Reviewers

- [ ] @mjkkirschner 
- [ ] @Randy-Ma 
- [ ] @ikeough

### DIFF
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -427,14 +427,12 @@ namespace Dynamo.Applications.Models
 
         private void SubscribeRevitServicesUpdaterEvents()
         {
-            RevitServicesUpdater.Instance.ElementsDeleted += RevitServicesUpdater_ElementsDeleted;
-            RevitServicesUpdater.Instance.ElementsModified += RevitServicesUpdater_ElementsModified;
+            RevitServicesUpdater.Instance.ElementsUpdated += RevitServicesUpdater_ElementsUpdated;
         }
 
         private void UnsubscribeRevitServicesUpdaterEvents()
         {
-            RevitServicesUpdater.Instance.ElementsDeleted -= RevitServicesUpdater_ElementsDeleted;
-            RevitServicesUpdater.Instance.ElementsModified -= RevitServicesUpdater_ElementsModified;
+            RevitServicesUpdater.Instance.ElementsUpdated -= RevitServicesUpdater_ElementsUpdated;
         }
 
         private void SubscribeTransactionManagerEvents()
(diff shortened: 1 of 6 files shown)


src/Libraries/RevitServices/Elements/ModelUpdater.cs
```diff
@@ -241,9 +249,63 @@ public void ApplicationDocumentChanged(object sender, DocumentChangedEventArgs a
...
+            Elements = elements;
+            Transactions = transactions;
+            Operation = operation;
+        }
+
+        /// <summary>
+        /// Returns unique ids for the elements involved in the event.
+        /// </summary>
+        /// <returns>List of unique id strings</returns>
+        public IEnumerable<string> GetUniqueIds() { return Elements.Select(x => RevitDocument.GetElement(x).UniqueId); }
```
mjkkirschner: could this ever return null or throw an exception?

src/Libraries/RevitServices/Elements/ModelUpdater.cs
```diff
@@ -241,9 +249,63 @@ public void ApplicationDocumentChanged(object sender, DocumentChangedEventArgs a
...
+
+        /// <summary>
+        /// The Revit document involved in the event.
+        /// </summary>
+        public Document RevitDocument { get; private set; }
+
+        /// <summary>
+        /// The name of transactions involved in the event.
+        /// </summary>
+        public IEnumerable<string> Transactions { get; set; }
```
mjkkirschner: can this also be a private setter? or can these properties even be read only?

src/Libraries/RevitServices/Elements/ModelUpdater.cs
```diff
@@ -29,36 +29,50 @@ public class RevitServicesUpdater
 
         private readonly HashSet<IUpdater> registeredUpdaters;
 
+        [Obsolete("Use ElementsUpdated event")]
```
mjkkirschner: so to clarify, these events are still raised so old clients can still use them, but they are no longer subscribed to anywhere in in this codebase?

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -809,44 +807,37 @@ private void TransactionManager_FailuresRaised(FailuresAccessor failuresAccessor
...
             var nodes = ElementBinder.GetNodesFromElementIds(
-                deleted,
+                e.Elements,
                 CurrentWorkspace,
                 EngineController);
             foreach (var node in nodes)
             {
+                //Don't re-execute the element construction node, if 
+                //this element is modified due to a Dynamo transaction.
+                if (dynamoTransaction && IsConstructorNode(node)) continue;
```
mjkkirschner: so in the case that we don't have a constructor node, but instead select and wrap a Revit family element from the project and then set its rotation - the set rotation node will not retrigger it's own execution still?

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -809,44 +807,37 @@ private void TransactionManager_FailuresRaised(FailuresAccessor failuresAccessor
...
-            var nodes = ElementBinder.GetNodesFromElementIds(
-                updatedIds,
-                CurrentWorkspace,
-                EngineController);
-            foreach (var node in nodes )
-            {
-            
-                node.OnNodeModified(true);
-            }
+            return func.Controller.Definition.Type == Engine.FunctionType.Constructor;
```
mjkkirschner: how does this play with custom nodes which return new elements?

src/Libraries/RevitServices/Elements/ModelUpdater.cs
```diff
@@ -206,32 +219,27 @@ public Document DocumentToWatch
...
         {
             var empty = new List<string>();
-            ProcessUpdates(doc, empty, deleted, empty, new List<ElementId>());
+            ProcessUpdates(doc, Enumerable.Empty<ElementId>(), deleted, Enumerable.Empty<ElementId>(), empty);
         }
 
-        private void ProcessUpdates(Document doc, IEnumerable<string> modified, 
-            IEnumerable<ElementId> deleted, IEnumerable<string> added, 
-            IEnumerable<ElementId> addedIds )
+        private void ProcessUpdates(Document doc, IEnumerable<ElementId> modified,
```
mjkkirschner: should we change names to be consistent since they are now all ids?

mjkkirschner:
@sharadkjaiswal looks really good to me. Just a few questions.

src/Libraries/RevitServices/Elements/ModelUpdater.cs
sharadkjaiswal: Yes, all the known clients have been updated to subscribe to the new event. For compatibility we are keeping the old events.

src/DynamoRevit/Models/RevitDynamoModel.cs
sharadkjaiswal: Yes because SetRotation caches the rotation value and before re-executing it checks against the input. If the input and the applied rotation is same then it won't update the element.

src/DynamoRevit/Models/RevitDynamoModel.cs
sharadkjaiswal: God point. If a custom node wraps a constructor node, this code will not detect that. Let me think about that.

PR #1031, opened 2016-06-12
MAGN-10168 With Kepler, Dynamo can be loaded when the Project is in perspective View and then closing Dynamo and the project can crash Revit

sm6srw:
### Purpose

1. Fix crash: don't try to clean up watch 3d view if it never was
initialized.
2. Show "Dynamo unavailable" when if a work space are loaded in an
invalid Revit context.
3. Show "Dynamo available" when switching from a invalid context to a
valid context.

Note: There is currently no easy way to add a test around this becuase the test framework can't handle a test that ends in an invalid Revit context (e.g. being in a perspective view).

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

@ramramps @QilongTang

### DIFF
--- a/src/DynamoRevit/Models/RevitDynamoModel.cs
+++ b/src/DynamoRevit/Models/RevitDynamoModel.cs
@@ -146,6 +146,12 @@ namespace Dynamo.Applications.Models
             {
                 node.PropertyChanged += node_PropertyChanged;
             }
+
+            var dm = DocumentManager.Instance;
+            if (dm.CurrentDBDocument != null)
+            {
+                SetRunEnabledBasedOnContext(dm.CurrentUIDocument.ActiveView, false);
+            }
         }
 
         private void node_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
@@ -540,7 +546,7 @@ namespace Dynamo.Applications.Models
         /// <param name="e"></param>
         internal void OnApplicationViewActivating(object sender, ViewActivatingEventArgs e)
         {
-            SetRunEnabledBasedOnContext(e.NewActiveView);
+            SetRunEnabledBasedOnContext(e.NewActiveView, true);
         }
 
         /// <summary>
@@ -611,7 +617,7 @@ namespace Dynamo.Applications.Models
                 workspace.ResetEngine(EngineController, markNodesAsDirty);
         }
 
-        public void SetRunEnabledBasedOnContext(View newView)
+        public void SetRunEnabledBasedOnContext(View newView, bool doContextAvailable)
         {
             DocumentManager.Instance.HandleDocumentActivation(newView);
 
@@ -632,6 +638,11 @@ namespace Dynamo.Applications.Models
             {
                 Logger.Log(
                     string.Format("Active view is now {0}", newView.Name));
+
+                if(doContextAvailable)
+                {
+                    OnRevitContextAvailable();
+                }
 
                 // If there is a current document, then set the run enabled
                 // state based on whether the view just activated is 
@@ -736,7 +747,7 @@ namespace Dynamo.Applications.Models
             var uiDoc = DocumentManager.Instance.CurrentUIDocument;
             if (uiDoc != null)
             {
-                SetRunEnabledBasedOnContext(uiDoc.ActiveView);
+                SetRunEnabledBasedOnContext(uiDoc.ActiveView, false);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -611,7 +617,7 @@ private void OnResetMarkNodesAsDirty(bool markNodesAsDirty)
                 workspace.ResetEngine(EngineController, markNodesAsDirty);
         }
 
-        public void SetRunEnabledBasedOnContext(View newView)
+        public void SetRunEnabledBasedOnContext(View newView, bool doContextAvailable)
```
sm6srw: Add new flag that allow us to show "Dynamo available" when going from bad Revit context to good Revit context.

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -736,7 +747,7 @@ private void HandleApplicationDocumentClosed()
             var uiDoc = DocumentManager.Instance.CurrentUIDocument;
             if (uiDoc != null)
             {
-                SetRunEnabledBasedOnContext(uiDoc.ActiveView);
+                SetRunEnabledBasedOnContext(uiDoc.ActiveView, false);
```
sm6srw: Don't show when a document is closing (this is how it was before).

src/DynamoRevit/ViewModel/RevitWatch3DViewModel.cs
```diff
@@ -363,11 +363,21 @@ private List<GeometryObject> Tessellate(Solid solid)
...
+            // Check for this condition before trying to access the current Revit document because there
+            // are cases when we get here uninitialized with a document that is gone already. 
+            if (keeperId == ElementId.InvalidElementId)
+            {
+                return;
+            }
+   
+            // Never access the current document with an invalid keeperId
+            // See comment at the beginning of this method.
             var dbDoc = DocumentManager.Instance.CurrentDBDocument;
```
sm6srw: Above fixes the crash. See comments.

src/DynamoRevit/Models/RevitDynamoModel.cs
mjkkirschner: since this is a public method (and lets say this is in the API... still up for debate) we should either add a new method and obsolete the old one, or simpler make doContextAvailable an optional parameter.

mjkkirschner: can we also add a summary for this public method.

sm6srw: Thanks, good catch. I didn't realize that is was public.

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -284,7 +284,7 @@ protected IEnumerable<NodeModel> AllNodes
 
         private void CurrentUIApplication_ViewActivating(object sender, ViewActivatingEventArgs e)
         {
-            ((RevitDynamoModel)this.ViewModel.Model).SetRunEnabledBasedOnContext(e.NewActiveView);
+            ((RevitDynamoModel)this.ViewModel.Model).SetRunEnabledBasedOnContext(e.NewActiveView, true);
```
mjkkirschner: depending on how you handle the change of this method can we also make sure the test the other options - either the default call or the old obsoleted method?

src/DynamoRevit/Models/RevitDynamoModel.cs
mjkkirschner: last thought: can you make this either DOContextAvailable or DocumentObject.... I think the intended pursue  will just be more clear.

mjkkirschner:
Hi @sm6srw  - LGTM, just 2 comments first.

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -620,7 +636,22 @@ public void SetRunEnabledBasedOnContext(View newView)
...
+                        currentState = true;
+                        break;
+                    }
+                }
+
+                // Only notify if we are changing state e.g. the Revit context becomes unavailable
+                if (currentState)
+                {
+                    OnRevitContextUnavailable();
+                }
```
sm6srw: Only notify if we actually are changing state.

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -633,6 +664,26 @@ public void SetRunEnabledBasedOnContext(View newView)
...
+                    }
+
+                    // Only notify if we are changing state e.g. the Revit context becomes available
+                    if (!currentState)
+                    {
+                        OnRevitContextAvailable();
+                    }
+                }
+
                 // If there is a current document, then set the run enabled
```
sm6srw: Same here...

sm6srw:
@mjkkirschner PTAL!

mjkkirschner:
LGTM

PR #1032, opened 2016-06-13
Revit2016 materials elements

moethu:
### Purpose

Element material and location properties added
Family instance Family, Type and Host property added
Material properties added

### Declarations

### Reviewers

This is basically the same Pull request than before but in its own branch.
@sharadkjaiswal , please review. Thanks, Max

### FYIs

CC @mjkkirschner


src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -324,12 +324,50 @@ internal static FamilyInstance FromExisting(Autodesk.Revit.DB.FamilyInstance fam
...
         public override string ToString()
         {
             return InternalFamilyInstance.Name;
         }
 
-       #region Public Methods
+        /// <summary>
+        /// Get FamilyInstance Host Element
+        /// </summary>
+        public Element GetHost
```
sharadkjaiswal: This property should be called Host and likewise other properties.

moethu:
@sharadkjaiswal , thanks for your review, applied the changes

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +718,108 @@ protected bool IsAlive
...
+                    Autodesk.DesignScript.Geometry.Point point = (Autodesk.DesignScript.Geometry.Point)geometry;
+                    Autodesk.Revit.DB.LocationPoint pt = (Autodesk.Revit.DB.LocationPoint)this.InternalElement.Location;
+                    pt.Point = point.ToRevitType(true);
+                }
+                else
+                    throw new Exception("Point element required");
+            }
+            else if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationCurve) && geometry is Curve)
+            {
+                if (geometry.GetType() == typeof(Curve))
```
sharadkjaiswal: What if `geometry` type is derived from Curve? Since you are anyway going to cast it later on, you can better cast the geometry to Curve before entering the if() block.
`var dynamoCurve = geometry as Curve; if(dynamoCurve != null){}`

You can do the same for other checks as well.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +718,108 @@ protected bool IsAlive
...
+            TransactionManager.Instance.TransactionTaskDone();
+        }
+
+        /// <summary>
+        /// Get an exsiting element's location
+        /// </summary>
+        /// <returns>Location Geometry</returns>
+        public Geometry Location()
+        {
+            if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationPoint))
```
sharadkjaiswal:
```
var pt = this.InternalElement.Location as Autodesk.Revit.DB.LocationPoint;
if(pt != null) return pt.Point.ToPoint(true);
```

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -101,6 +101,189 @@ public static Material ByName(string name)
...
+
+        /// <summary>
+        /// Get all apperance parameters
+        /// </summary>
+        public List<Parameter> AppearanceParameters
+        {
+            get
+            {
+                // Get the active Document
+                Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
```
sharadkjaiswal: you can user `var document = ...`

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -101,6 +101,189 @@ public static Material ByName(string name)
...
+                }
+
+                return thermals;
+            }
+        }
+
+        /// <summary>
+        /// Get all structural parameters
+        /// </summary>
+        public List<Parameter> StructuralParameters
```
sharadkjaiswal: Better to return `IEnumerable<Parameter>`

src/Libraries/RevitNodes/Elements/Element.cs
moethu: @sharadkjaiswal good point, I changed the checks using typeof(Curve).IsAssignableFrom(geometry.GetType()) to make sure all derived types are catched as well.

moethu:
@sharadkjaiswal applied changes in https://github.com/DynamoDS/DynamoRevit/pull/1032/commits/fe15da7d7cc3687b4aa12e577ef945dd4bf9dc68

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -728,31 +728,21 @@ public void SetLocation(Geometry geometry)
         {
             TransactionManager.Instance.EnsureInTransaction(Application.Document.Current.InternalDocument);
 
-            if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationPoint))
+            if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationPoint) && typeof(Autodesk.DesignScript.Geometry.Point).IsAssignableFrom(geometry.GetType()))
```
sharadkjaiswal: You could have done the casting before the if() condition and then check for null. Here type check is happening twice.
```
var locationPoint = this.InternalElement.Location as Autodesk.Revit.DB.LocationPoint;
var point = geometry as Autodesk.DesignScript.Geometry.Point;
if(locationPoint != null && point != null)
{
    locationPoint.Point = point.ToRevitType(true);
}
```

moethu: @sharadkjaiswal can you explain it more please? I need to check first if my Element has a LocationPoint and if the supplied geometry is type of point. Once those are fine I can apply the geometry to the element.
In the other case both, argument and element.location must be curves. Then the location curve is being applied.

sharadkjaiswal: You can include casting for curves also at the beginning. Also instead of checking IsAssignable, or equality for types, you can directly use `if(geometry is Point){..}`

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -766,17 +756,31 @@ public Geometry Location()
         {
             if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationPoint))
             {
-                Autodesk.Revit.DB.LocationPoint pt = (Autodesk.Revit.DB.LocationPoint)this.InternalElement.Location;
-                return pt.Point.ToPoint(true);
+                Autodesk.Revit.DB.LocationPoint pt = this.InternalElement.Location as Autodesk.Revit.DB.LocationPoint;
```
sharadkjaiswal: These lines are becoming too long, you can user `var pt = ...` , you can also define a new name for this type `using DBLocationPoint = Autodesk.Revit.DB.LocationPoint;`

PR #1038, opened 2016-06-16
manage dynamoRevitModel startup sequence

BogdanZavu:
### Purpose

Mange dynamorevitModel startup sequence based on input JournalKeys.
Expose dynamoRevitModel state so dependent components can take decisions based on it regarding their UI and functionality.

### Declarations

- [X] Snapshot of UI changes, if any.

### Reviewers

@mjkkirschner 
@sharadkjaiswal

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -23,6 +23,7 @@ using Dynamo.Applications.Models;
 using Dynamo.Applications.ViewModel;
 using Dynamo.Controls;
 using Dynamo.Core;
+using Dynamo.Graph.Workspaces;
 using Dynamo.Logging;
 using Dynamo.Models;
 using Dynamo.Updates;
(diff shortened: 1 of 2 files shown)


sm6srw:
I have two questions: I assume this is needed for a 2017 update release (and not 2018), right? So will this work without any issues on 2017 RTM?

BogdanZavu:
Yes , our target is 2017 update release R2 . We payed attention to not break binary compatibility so there should be no problem for 2017 RTM.

sm6srw:
LGTM. @mjkkirschner will have the final say...

mjkkirschner:
@sm6srw thanks for checking this out, @BogdanZavu this looks good to me - lets wait on what @sharadkjaiswal has to say and then get this in. :)

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -112,42 +113,85 @@ public DynamoRevitCommandData()
...
+        private static DynamoViewModel dynamoViewModel;
+        private static RevitDynamoModel revitDynamoModel;
+        private static bool handledCrash;
+        private static List<Exception> preLoadExceptions;
+
+        /// <summary>
+        /// The modelState tels us if the RevitDynamoModel was started and if has the
+        /// the Dynamo UI attached to it or not 
+        /// </summary>
+        public static RevitDynamoModelState ModelState
```
sharadkjaiswal: Can we define ModelState on RevitDynamoModel, so that we can have all the states related to DynamoModel consolidated with the same object. If dynamoViewModel is not initialized can we safely assume that we are running in UI-less mode?

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -112,42 +113,85 @@ public DynamoRevitCommandData()
...
-        // is pulled from the journal file.
-        private static bool shouldShowUi = true;
-        private static bool isAutomationMode;
-
         /// <summary>
         /// The journal file can use this key to specify whether
         /// the Dynamo UI should be visible at run time.
         /// </summary>
-        private const string JournalShowUiKey = "dynShowUI";
+        public const string JournalShowUiKey = "dynShowUI";
```
sharadkjaiswal: you can drop Journal from the name as all of these are Journal keys.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -112,42 +113,85 @@ public DynamoRevitCommandData()
         public IDictionary<string, string> JournalData { get; set; }
     }
 
-    [Transaction(TransactionMode.Manual),
-     Regeneration(RegenerationOption.Manual)]
-    public class DynamoRevit : IExternalCommand
+    /// <summary>
+    /// Defines startup parameters for DynamoRevitModel
+    /// </summary>
+    public class DynamoRevitJournalKeys
```
sharadkjaiswal: JournalKeys could be a short and sweet name.

src/DynamoRevit/DynamoRevitApp.cs
```diff
@@ -110,6 +110,39 @@ private static string GetDynamoRoot(string dynamoRevitRoot)
         private static EventHandlerProxy proxy;
         private AddInCommandBinding dynamoCommand;
 
+        private Result loadDependentComponents()
+        {
+            var dynamoRevitRootDirectory = Path.GetDirectoryName(Path.GetDirectoryName(assemblyName));
+            var versionNumber = ControlledApplication.VersionNumber;
+
+            var dynamoRevitAditionsPath = Path.Combine(dynamoRevitRootDirectory,
```
sharadkjaiswal: assemblyName is full path for DynamoRevit.dll, you could have used that directly.

sharadkjaiswal:
@BogdanZavu thanks for working on it. Just a few minor comments, otherwise looks good.

src/DynamoRevit/DynamoRevit.cs
BogdanZavu: Ok

src/DynamoRevit/DynamoRevit.cs
BogdanZavu: Ok

src/DynamoRevit/DynamoRevit.cs
BogdanZavu: Model state takes into account both revitDynamoModel and dynamoViewModel so for now I would say to keep it outside these two. We can assume that if dynamoViewModel is not initialized we are UI-less mode but in order to be in UI mode it looks that there are some other operations that need to be performed on dynamoViewModel . So we can't rely only on the dynamoViewModel initialization to conclude that we are in UI-less mode.

src/DynamoRevit/DynamoRevitApp.cs
BogdanZavu: Ok

BogdanZavu:
Thank you @sharadkjaiswal ,
I made the adjustments based on your observations and pushed another commit.

mjkkirschner:
@BogdanZavu thanks for this work, will merge soon.

PR #1044, opened 2016-06-20
Fix MAGN-10069 FamilyInstance.SetRotation misplaces family on first run

Randy-Ma:
### Purpose

This pull request fixes MAGN-10069. The family instances created by FamilyInstance.ByPoint and FamilyInstance.ByPointAndLevel do not have correct locations(similar with transforms) as specified by the points when the locations are obtained immediately after the family instances are created. But in the case of the defect, the rotation highly depends on the rotation axises which pass through the locations(the origin of the transforms). If the obtained locations are wrong, so as the rotations. The fix here is to renerate the document before trying to get the transform of the family instance so that the transform obtained is correct.

### Declarations

### Reviewers

@sharadkjaiswal 

### FYIs

@kronz @riteshchandawar

### DIFF
--- a/src/Libraries/RevitNodes/Elements/FamilyInstance.cs
+++ b/src/Libraries/RevitNodes/Elements/FamilyInstance.cs
@@ -330,28 +330,30 @@ namespace Revit.Elements
         /// <returns>The result family instance.</returns>
         public FamilyInstance SetRotation(double degree)
         {
-           if (this == null)
-              throw new ArgumentNullException("familyInstance");
+            if (this == null)
+                throw new ArgumentNullException("familyInstance");
 
-           TransactionManager.Instance.EnsureInTransaction(Document);
+            TransactionManager.Instance.EnsureInTransaction(Document);
 
-           // Rotate the element.
-           var oldTransform = InternalFamilyInstance.GetTransform();
-           double[] oldRotationAngles;
-           TransformUtils.ExtractEularAnglesFromTransform(oldTransform, out oldRotationAngles);
+            Document.Regenerate();
 
-           Double newRotationAngle = degree * Math.PI / 180;
+            // Rotate the element.
+            var oldTransform = InternalFamilyInstance.GetTransform();
+            double[] oldRotationAngles;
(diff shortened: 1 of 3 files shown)


src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -330,28 +330,30 @@ public override string ToString()
...
+                throw new ArgumentNullException("familyInstance");
 
-           TransactionManager.Instance.EnsureInTransaction(Document);
+            TransactionManager.Instance.EnsureInTransaction(Document);
 
-           // Rotate the element.
-           var oldTransform = InternalFamilyInstance.GetTransform();
-           double[] oldRotationAngles;
-           TransformUtils.ExtractEularAnglesFromTransform(oldTransform, out oldRotationAngles);
+            Document.Regenerate();
```
sharadkjaiswal: I'll suggest, let's have an internal method to GetTransform() or GetRotation() and call Document.Regenerate() from there, so that we can reuse the same code later on.

Randy-Ma: I prefer GetRotations() because there will be three rotation angles around x axis, y axis and z axis.

Randy-Ma: InternalGetTransform() looks better as the result will also contain the location information.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -321,39 +321,54 @@ public override string ToString()
...
 
-           TransactionManager.Instance.TransactionTaskDone();
+        #region Private helper methods
 
-           return this;
+        /// <summary>
+        /// Get the transform of the internal family instance
+        /// </summary>
+        /// <returns>The internal transform</returns>
+        private Transform InternalGetTransform()
```
sharadkjaiswal: :) you could have made this method internal and the method name as GetTransform(), so that other classes may use this method if required.

PR #1047, opened 2016-06-20
Fix the build error

Randy-Ma:
### Purpose

This fixes the build error which is caused by a missing namespace.

### Declarations

### Reviewers

@sharadkjaiswal


Randy-Ma:
@sharadkjaiswal 
I am merging this as it is a straight-forward fix.

PR #1056, opened 2016-06-22
Fix MAGN-9855 AdaptiveComponent.ByParametersOnFace and OnCurve creates components that are not hosted.

Randy-Ma:
### Purpose

When adaptive components are created in batch, there is no method to make them hosted on curves or faces. @kronz suggests that hosting adaptive components on the related curves or faces is more important. In this submission, the batch processing behavior is removed for the two AdaptiveComponent nodes(ByParametersOnCurveReference and ByParametersOnFace). But we will make the created components hosted on the curves or the faces.

At the same time, element binding (which is not working before) is fixed for the two nodes.

### Declarations

### Reviewers

@sharadkjaiswal 

### FYIs

@kronz

### DIFF
--- a/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
+++ b/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
@@ -116,6 +116,7 @@ namespace Revit.Elements
             }
 
             TransactionManager.Instance.TransactionTaskDone();
+            ElementBinder.SetElementForTrace(InternalElement);
 
         }
 
@@ -163,6 +164,7 @@ namespace Revit.Elements
             }
 
             TransactionManager.Instance.TransactionTaskDone();
+            ElementBinder.SetElementForTrace(InternalElement);
         }
 
         #endregion
--- a/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
+++ b/src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
@@ -75,10 +75,10 @@ namespace Revit.Elements
         /// <summary>
         /// Initialize an AdaptiveComponent element
         /// </summary>
-        /// <param name="pts">Points to use as reference</param>
+        /// <param name="uvParams">Parameters to use as reference</param>
         /// <param name="f">Face to use as reference</param>
         /// <param name="ft">familyType to place</param>
-        private void InitAdaptiveComponent(double[][] pts, ElementFaceReference f, FamilyType ft)
+        private void InitAdaptiveComponent(double[][] uvParams, ElementFaceReference f, FamilyType ft)
         {
             // if the family instance is present in trace...
             var oldFam =
@@ -90,7 +90,7 @@ namespace Revit.Elements
                 InternalSetFamilyInstance(oldFam);
                 if (ft.InternalFamilySymbol.Id != oldFam.Symbol.Id)
                     InternalSetFamilySymbol(ft);
-                InternalSetUvsAndFace(pts.ToUvs(), f.InternalReference);
+                InternalSetUvsAndFace(uvParams.ToUvs(), f.InternalReference);
 
                 return;
             }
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -50,6 +72,101 @@ private void InitAdaptiveComponent(Autodesk.Revit.DB.FamilyInstance familyInstan
             TransactionManager.Instance.TransactionTaskDone();
         }
 
+        /// <summary>
+        /// Initialize an AdaptiveComponent element
+        /// </summary>
+        /// <param name="pts">Points to use as reference</param>
```
sharadkjaiswal: I think the pts is UV parameters, you can name it as uv or uvParams.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -74,6 +191,78 @@ private static void UpdatePlacementPoints(Autodesk.Revit.DB.FamilyInstance fi, X
...
+        /// </summary>
+        /// <param name="uvs"></param>
+        /// <param name="faceReference"></param>
+        private void InternalSetUvsAndFace(Autodesk.Revit.DB.UV[] uvs, Autodesk.Revit.DB.Reference faceReference)
+        {
+            TransactionManager.Instance.EnsureInTransaction(Document);
+
+            var placePointIds = AdaptiveComponentInstanceUtils.GetInstancePlacementPointElementRefIds(InternalFamilyInstance);
+
+            if (placePointIds.Count() != uvs.Length)
```
sharadkjaiswal: What happens when the node is re-executed with a different number of uv parameters?

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -127,8 +316,7 @@ public List<Point> Locations
         // without breaking replication guides
 
         [IsVisibleInDynamoLibrary(false)]
-        [AllowRankReduction]
-        public static AdaptiveComponent[] ByParametersOnFace(double[][][] uvs, ElementFaceReference faceReference, FamilyType familyType)
```
sharadkjaiswal: To maintain binary compatibility, we can't really remove this method. Can you please update this method`s implementation based on the new method. Also mark this method as [SuppressImportInVM] so this method is not available in DS Engine.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
```diff
@@ -127,8 +316,7 @@ public List<Point> Locations
...
+        public static AdaptiveComponent ByParametersOnFace(double[][] uvs, ElementFaceReference faceReference, FamilyType familyType)
```
sharadkjaiswal: Do we need to migrate the nodes?

sharadkjaiswal: Why is this new method is hidden in the library? Can we take a list of Geometry.UV instead of double?

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
Randy-Ma: This is a private method. If the number of uv parameters does not equal to the number of placement points, an exception will be thrown.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
Randy-Ma: ByParametersOnFace has 4 overloaded functions. This is hidden originally. I guess it is because actually it is only supposed to be supplementory to the other one with the same number of parameters. So if the input is not Geometry.UV[], it still can work.

Randy-Ma: Do we need to migrate? I have thought it will replicate.

src/Libraries/RevitNodes/Elements/AdaptiveComponent.cs
Randy-Ma: So it won't automatically replicate?

aparajit-pratap: Also would be a good idea to add the `Obsolete` attribute with an appropriate message

PR #1065, opened 2016-06-27
Filters added 2016

moethu:
### Purpose

Revit View Filters added

### Declarations

### Reviewers

@mjkkirschner , let me know if this is the way you had the filters in mind.

### FYIs

@sharadkjaiswal @Randy-Ma


src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -135,5 +135,62 @@ public override string ToString()
...
+            RevitServices.Transactions.TransactionManager.Instance.EnsureInTransaction(Application.Document.Current.InternalDocument);
+            this.InternalView.AddFilter(parameterFilter.InternalElement.Id);
+            RevitServices.Transactions.TransactionManager.Instance.TransactionTaskDone();
+        }
+
+        /// <summary>
+        /// Get View Filters
+        /// </summary>
+        /// <returns></returns>
+        public List<Revit.Filter.ParameterFilterElement> Filters
```
mjkkirschner: can this be an IEnumerable<>?

src/Libraries/RevitNodes/Filter/FilterRule.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// <summary>
+    /// Revit Filter Rule
+    /// </summary>
+    public class FilterRule
+    {
+
+        #region Internal Properties
+
+        /// <summary>
+        /// Internal reference to the Revit Element
```
mjkkirschner: this actually isn't a revit element though right?

src/Libraries/RevitNodes/Filter/FilterRule.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// </summary>
+        internal Autodesk.Revit.DB.FilterRule InternalFilterRule
+        { 
+            get; set;
+        }
+
+        /// <summary>
+        /// Reference to the Element
```
mjkkirschner: same comment

mjkkirschner: also this function seems kind of redundant... why do we need this?

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// </summary>
+        internal Autodesk.Revit.DB.OverrideGraphicSettings InternalOverrideGraphicSettings
+        { 
+            get; set;
+        }
+
+        /// <summary>
+        /// Reference to the Element
+        /// </summary>
+        internal OverrideGraphicSettings(Autodesk.Revit.DB.OverrideGraphicSettings internalOverrideGraphicSettings)
```
mjkkirschner: same as above, is this necessary?

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <param name="projectionLineColor">Projection line color</param>
+        /// <param name="cutLineWeight">Cut line weight</param>
+        /// <param name="projectionLineWeight">Projection line weight</param>
+        /// <param name="cutFillPattern">Cut fill pattern</param>
+        /// <param name="projectionFillPattern">Projection fill pattern</param>
+        /// <param name="cutLinePattern">Cut line pattern</param>
+        /// <param name="projectionLinePattern">Projection line pattern</param>
+        /// <returns>OverrideGraphicSettings</returns>
+        public static OverrideGraphicSettings ByProperties(
+            DSCore.Color cutFillColor = null,
```
mjkkirschner: do these nulls work correctly for default values, last I remember was that this caused a partial application of the node instead of running automatically. Maybe this behavior was fixed?

src/Libraries/RevitNodes/Filter/ParameterFilterElement.cs
```diff
@@ -0,0 +1,171 @@
...
+            SafeInit(() => InitElement(elem));
+        }
+
+        /// <summary>
+        /// ParameterFilterElement by Rules
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="ids"></param>
+        /// <param name="rules"></param>
+        private ParameterFilterElement(string name, List<ElementId> ids, List<Autodesk.Revit.DB.FilterRule> rules)
```
mjkkirschner: this could probably accept a IEnumerable<Id>?

src/Libraries/RevitNodes/Filter/ParameterFilterElement.cs
```diff
@@ -0,0 +1,171 @@
...
+            InternalSetElement(element);
+        }
+
+        /// <summary>
+        /// Init ParameterFilterElement by rules
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="ids"></param>
+        /// <param name="rules"></param>
+        private void Init(string name, List<ElementId> ids, List<Autodesk.Revit.DB.FilterRule> rules )
```
mjkkirschner: same question about IEnums where possible.

src/Libraries/RevitNodes/Filter/ParameterFilterElement.cs
```diff
@@ -0,0 +1,171 @@
...
+        #region Public static constructors
+
+        /// <summary>
+        /// Create a Revit Parameter Filter Element
+        /// </summary>
+        /// <param name="name">Name</param>
+        /// <param name="categories">Categories the filter applies to</param>
+        /// <param name="rules">Filter rules</param>
+        /// <returns></returns>
+        public static ParameterFilterElement ByRules(string name, List<Revit.Elements.Category> categories, List<FilterRule> rules)
```
mjkkirschner: same question about IEnums

src/Libraries/RevitNodes/Properties/Resources.resx
```diff
@@ -312,4 +312,7 @@
   <data name="MultipleSurfacesIntroducedAfterConversion" xml:space="preserve">
     <value>There are multiple surfaces converted from the input face.</value>
   </data>
+  <data name="InvalidFilterType" xml:space="preserve">
```
mjkkirschner: this should get into the en-us.resx as well.

test/Libraries/RevitNodesTests/RevitNodesTests.csproj
```diff
@@ -39,6 +39,11 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="DSCoreNodes, Version=1.0.1.1523, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>../../../../Dynamo/bin/AnyCPU/Debug/DSCoreNodes.dll</HintPath>
```
mjkkirschner: I believe there is a configuration parameter in the config.cs props file that defines a DynamoAPI path you can use here.

src/Libraries/RevitNodes/Filter/FilterRule.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            if (!Enum.TryParse<FilterType>(typename, out type))
+            {
+                throw new Exception(Properties.Resources.InvalidFilterType);
+            }
+
+            ElementId parameterId = parameter.InternalParameter.Id;
+
+            // assemble the method name to construct a new filter using the FilterType
+            string methodname = "Create" + type.ToString() + "Rule";
```
mjkkirschner: define these as constants somewhere.

src/Libraries/RevitNodes/Filter/FilterRule.cs
```diff
@@ -0,0 +1,144 @@
...
+        #region Public static constructors
+
+        /// <summary>
+        /// Create a new Filter Rule
+        /// </summary>
+        /// <param name="type">Filter Rule Type</param>
+        /// <param name="value">Value to check</param>
+        /// <param name="parameter">Parameter to filter</param>
+        /// <returns></returns>
+        public static FilterRule ByRuleType(string typename, object value, Elements.Parameter parameter)
```
mjkkirschner: couldn't the typename be a member of the enum directly?

src/Libraries/RevitNodes/Filter/FilterRule.cs
```diff
@@ -0,0 +1,144 @@
...
+            if (
+                type == FilterType.Equals ||
+                type == FilterType.NotEquals ||
+                type == FilterType.Greater ||
+                type == FilterType.Less ||
+                type == FilterType.GeaterOrEqual ||
+                type == FilterType.LessOrEqual
+                )
+            {
+                if (value.GetType() == typeof(int))
```
mjkkirschner: I'm not sure about this method for using the API (relying on reflection) but I don't know much about it and I'm not sure there is any better way.

src/Libraries/RevitNodes/Filter/FilterRule.cs
moethu: @mjkkirschner it is a revit element (Autodesk.Revit.DB.FilterRule) but it doesn't require any bindings. I followed similar existing implementations like Parameters (https://github.com/moethu/DynamoRevit/blob/Revit2015/src/Libraries/RevitNodes/Elements/Parameter.cs) So its basically a wrapper for the FilterRule with an internal property containing the actual Revit element.

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
moethu: this would be the same situation, Revit element wrapper like the parameter implementation.

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
moethu: they don't work in 1.0.0.1180 yet. I changed them to [DefaultArgumentAttribute("null")]

src/Libraries/RevitNodes/Filter/FilterRule.cs
moethu: You mean naming them entirely like the method like "CreateLessOrEqualRule"? I tried to keep it as human readable as possible, but if you want, I'll change that.

src/Libraries/RevitNodes/Filter/FilterRule.cs
moethu: The other solution would be to write every call explicitly, but it would be a switch over at least 18 cases.

src/Libraries/RevitNodes/Filter/FilterRule.cs
sharadkjaiswal: You can use the enum type as input rather than taking a string and converting it to enum. Users may not know which are the valid input string.

src/Libraries/RevitNodes/Filter/FilterRule.cs
sharadkjaiswal: Adding switch case is always better over multiple if conditions.

src/Libraries/RevitNodes/Filter/FilterRule.cs
moethu: @sharadkjaiswal @mjkkirschner can you cast a string into an enum in Dynamo? Because for the rebar package for example most users are constructing their enum values as strings. Maybe we should add a enum cast node allowing the users to get their enum value from strings?

src/Libraries/RevitNodes/Filter/FilterRule.cs
moethu: @sharadkjaiswal but the switch case wouldn't work on types, except switching on their string value. What do you think about using reflection in this case (as it is implemented at the moment) to avoid a lot of switching / if statements?

mjkkirschner:
@moethu I'm not sure what to do here with the reflection switch, seems you could do something like this:

http://stackoverflow.com/questions/5895055/how-to-rewrite-if-to-switch-about-identifying-dynamic-class-type

https://chodounsky.net/2014/01/29/dynamic-dispatch-in-c-number/

if you think these approaches will not work or will take too long then we can stick with the reflection, but please pull out the strings.

Also please create a 2018 PR for each of your remaining PRs as we have recently created this branch and don't want to lose your changes. Feel free to do this after your PRs get LGTM so you don't have to keep them all up to date for no reason.

moethu:
@mjkkirschner sure, I will create the 2017 and 2018 PRs. I put the strings into consts if this is fine.

mjkkirschner:
@moethu let's do the string removal, one more thing, can you please create 1 more set of PRS for 1.1RC for those PRS that get a LGTM?

moethu:
@mjkkirschner strings are done. will place PRs for 1.1 as well, had only one LGTM yet.

mjkkirschner:
this looks pretty good to me, I think we'll need to improve the reflection / switch case bit as a follow up task using a dynamic type or dict of type -> methods, but for now it looks good, I'd like to get it in so it can be tested.

please if you have not done so already, create cherry picked branches for 2016 1.1, 2017, 2017 1.1, 2018.

moethu:
@mjkkirschner sure will do. I don't have 2018 though.

mjkkirschner:
I am going to merge this in to master, not to 1.1. I will cross merge to 2018 and 2017 next.

mjkkirschner:
@moethu does this API exist in revit 2015?

PR #1072, opened 2016-06-30
MAGN-10290 - Fix Hardware Acceleration Crash in Revit 2017

sm6srw:
### Purpose

By default, Dynamo4Revit has WPF Hardware Acceleration turned on when Dynamo loads and when the Dynamo button is clicked. This is causing crashes in Revit 2017.

The proposed fix from our end is to let Revit decide if WPF Hardware acceleration should be on or off. The default is off. Hardware Acceleration can then be turned on in Revit 2017 SP2 or later by adding "UseHardwareAcceleration=1" to the [UserInterface] section of the Revit.ini. UI for controlling this setting will be provided in Revit 2018 or later.

### Declarations

### Reviewers

@QilongTang 

### FYIs

@mjkkirschner @jnealb

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -35,7 +35,7 @@ using RevitServices.Threading;
 using MessageBox = System.Windows.Forms.MessageBox;
 using DynUpdateManager = Dynamo.Updates.UpdateManager;
 using Microsoft.Win32;
-
+using System.Windows.Media;
 
 namespace RevitServices.Threading
 {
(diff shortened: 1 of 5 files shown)


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -253,7 +253,13 @@ public Result ExecuteCommand(DynamoRevitCommandData commandData)
                 if (CheckJournalForUiDisplay(extCommandData))
                 {
                     dynamoViewModel = InitializeCoreViewModel(revitDynamoModel);
+
+                    // Let the host (e.g. Revit) control the rendering mode
+                    var save = RenderOptions.ProcessRenderMode;
                     InitializeCoreView().Show();
+                    RenderOptions.ProcessRenderMode = save;
+                    revitDynamoModel.Logger.Log("WPF Render Mode: " + RenderOptions.ProcessRenderMode.ToString());
```
QilongTang: I know it might not be required, but do you think we should put "WPF Render Mode" into resource string so that they could be localized? Cause the render mode messages seem to be important to user to know.

QilongTang:
Otherthan one small question, LGTM.

src/DynamoRevit/DynamoRevit.cs
sm6srw: That is a good idea. I will add a new string resource.

PR #1092, opened 2016-07-04
Fix CachedValueChanges_HasItemsAndSelectedIndex failure

sharadkjaiswal:
### Purpose

This PR fixes CachedValueChanges_HasItemsAndSelectedIndex test failure. This test makes use of FamilyInstanceParameters node, which has EngineController property that is set from its node view customization class. During the test run, the node view customization is not initialized and hence EngineController is not set properly.

To fix this test scenario, let's use the CachedValue of the input node if EngineController is not set. However, the CachedValue may not represent the true value connected to this node if the input node has multiple output ports. We might need a different way to query the mirror data of any output port.

**Why Regression?**
Get Family Parameters node was getting into an infinite loop because the items for the drop-down were populated on every evaluation which was changing the selection index and hence node was marked dirty. #1015 fixed this issue by avoiding the PopulateItems call if the input element has not changed and to find the input element it needs access to EngineController. Earlier this test case was using the cached element id in DYN file to populate drop-down items. 

### Declarations

### Reviewers
- [ ] @ikeough 
- [ ] @mjkkirschner 

### FYIs
@Benglin


Benglin:
Thanks so much for fixing, @sharadkjaiswal! This looks great!

PR #1126, opened 2016-07-16
fix freeze for multiSerializeableIds

mjkkirschner:
### Purpose
http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10285

Freeze stopped working on nodes that saved multiple element ids into trace. This PR does the right thing when setting the freeze state of an element if the object that comes back out of trace for the node callsite is a `MultiSerializeableID`

### Declarations

### Reviewers

@ramramps 

### FYIs

@kronz


ramramps:
LGTM. Do we need this for revit 2017 and RC 1.1?

kronz:
Yes, Revit 2017 too and RC 1.1.  Thanks!

PR #1192, opened 2016-08-17
Add two test cases for element binding related to re-opening the graph

Randy-Ma:
### Purpose

This is to add two element binding test cases:
1). Run a graph, then close the graph and then open the same graph and run again;
2). Run a graph, then save and close the graph, and then open the same graph and run again.

The graph in use is simply to add one reference point at (0, 0, 0).

For test case 1, finally there will be two reference points; while for test case 2, finally there will be one reference points.

### Declarations

### Reviewers

@QilongTang 

### FYIs

@kronz @sharadkjaiswal @riteshchandawar

### DIFF
--- a/test/Libraries/RevitIntegrationTests/ElementBindingTests.cs
+++ b/test/Libraries/RevitIntegrationTests/ElementBindingTests.cs
@@ -243,6 +243,64 @@ namespace RevitSystemTests
 
         [Test]
         [TestModel(@".\empty.rfa")]
+        public void CreateInDynamoCloseGraphReopenGraphRerun()
+        {
+            //Create a reference point at (0.0, 0.0, 0.0);
+            string dynFilePath = Path.Combine(workingDirectory, @".\ElementBinding\CreateOneReferencePoint.dyn");
+            string testPath = Path.GetFullPath(dynFilePath);
+
+            ViewModel.OpenCommand.Execute(testPath);
+
+            RunCurrentModel();
+
+            //Close the current graph
+            ViewModel.CloseHomeWorkspaceCommand.Execute(null);
+
+            //Open the same graph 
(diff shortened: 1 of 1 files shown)


test/Libraries/RevitIntegrationTests/ElementBindingTests.cs
```diff
@@ -243,6 +243,64 @@ public void CreateInDynamoModifyInRevitReRun()
...
+            ViewModel.CloseHomeWorkspaceCommand.Execute(null);
+
+            //Open the same graph 
+            ViewModel.OpenCommand.Execute(testPath);
+
+            //Run the graph once again
+            RunCurrentModel();
+
+            var points = GetAllReferencePointElements(true);
+            Assert.AreEqual(2, points.Count);
```
QilongTang: Can we simply Assert.AreNotEqual(1, point.Count) here? Just to rule out all other possibilities.

test/Libraries/RevitIntegrationTests/ElementBindingTests.cs
```diff
@@ -243,6 +243,64 @@ public void CreateInDynamoModifyInRevitReRun()
...
+            var pnt = points[0] as ReferencePoint;
+            Assert.IsTrue(pnt.Position.IsAlmostEqualTo(new XYZ(0.0, 0.0, 0.0)));
```
QilongTang: Looks like this test case will create two identical points. Can we check the  second point as well?

QilongTang:
Thanks Randy, Just one comment then LGTM.

QilongTang:
I think it is good to be merged into regardless of the comment. Please cherry-pick this to other branches as well.

QilongTang:
Randy, Revit 2018 branch is there as well. Can you cherry pick there as well. Thanks!

PR #1196, opened 2016-08-18
Preload ASM for Revit

aparajit-pratap:
### Purpose

This is a fix for [MAGN-10546] (http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10546). 

### Background

The issue here was that in the presence of Alias 2017 installed, the wrong ASM version was being inherited (the version from Alias instead of from Revit 2017) and since LibG is incompatible with the version of ASM in Alias, it was refusing to load. As a result all geometry nodes are rendered useless. The reason for this is that installing Alias adds the Alias bin folder containing ASM binaries to the PATH envar, which is why these binaries took precedence over those from Revit. This could happen for any other Autodesk product that comes with an ASM installation. 

### Solution

The fix for this is to preload ASM binaries specifically from the Revit installation that is being used to host Dynamo just like we do in the case of DynamoSandbox. This way there is no confusion about the version of ASM that will be loaded for Dynamo-Revit and we can ensure that LibG will always load no matter which other Autodesk product is installed on the users system.

### Declarations

### Reviewers

@sharadkjaiswal


sharadkjaiswal:
Smart implementation @aparajit-pratap, good stuff.

aparajit-pratap:
Thanks @sharadkjaiswal

jnealb:
@aparajit-pratap @mjkkirschner @sharadkjaiswal @Racel @ramramps @saintentropy FYI: there is still no build on the EC with this PR incorporated, the last build from the 18th does not have this fix.

QilongTang:
@aparajit-pratap Can we cross merge this into Revit2018 branch as well?

aparajit-pratap:
@QilongTang I need to cross-merge a couple of others as well, will do so today. Thanks for pointing it out!

PR #1205, opened 2016-08-22
Disallow Selection nodes from updating for Dynamo transactions

aparajit-pratap:
### Purpose

This fixes [MAGN-10410] (http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10410). There are certain cases where the family created cuts a hole in the host Revit face. If such a family is used in the `FamilyInstance.ByFace` node, it will modify the parent face that hosts it in the first place, thereby causing a graph re-execution leading to an indefinite cyclic loop. 

The host face originates from a Revit `Selection` node however the solution cannot be to block any selection nodes from re-executing. There are workflows where the elements in the Revit document require to be manually edited and cause the graph to re-evaluate, which in this case would be broken if we blocked all selection nodes from re-executing. The solution therefore is to allow selection nodes to update ONLY for non-Dynamo transactions, and prevent update for graph transactions.

### Declarations

### Reviewers

@sharadkjaiswal 

@Randy-Ma


mjkkirschner:
please also merge to 2018 branch if you have not yet.

PR #1209, opened 2016-08-22
Revit2017

aparajit-pratap:
### Purpose

This is a fix for crash - [MAGN-10578] (http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10578). The crash occurred when the `ToString` method of the `InternalFamilyInstance` property of an invalid `FamilyInstance` object was being queried, in which case it was throwing an unhandled object invocation exception. 

The fix is to check for the validity of the `InternalFamilyInstance` object (using `IsValidObject` property) before querying any properties for it.

### Declarations

### Reviewers

@sharadkjaiswal


mjkkirschner:
please also merge to 2018 branch if you have not yet.

PR #1218, opened 2016-08-25
Fix build errors due to API changes in Revit 2018

QilongTang:
### Purpose

Fix build errors due to API changes in Revit 2018.

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/test/Libraries/RevitNodesTests/Elements/TagTests.cs
+++ b/test/Libraries/RevitNodesTests/Elements/TagTests.cs
@@ -28,7 +28,7 @@ namespace RevitNodesTests.Elements
             var wall = Wall.ByCurveAndHeight(line, 300, Level.ByElevation(100), wt);
             Assert.NotNull(wall);
 
-            var tag = Tag.ByElement(Revit.Application.Document.Current.ActiveView, wall, true, false, Autodesk.Revit.DB.HorizontalAlignmentStyle.Center, Autodesk.Revit.DB.VerticalAlignmentStyle.Middle, Vector.ByCoordinates(0, 0, 0));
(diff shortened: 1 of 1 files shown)


test/Libraries/RevitNodesTests/Elements/TagTests.cs
```diff
@@ -28,7 +28,7 @@ public void ByElement_ValidArgs()
             var wall = Wall.ByCurveAndHeight(line, 300, Level.ByElevation(100), wt);
             Assert.NotNull(wall);
 
-            var tag = Tag.ByElement(Revit.Application.Document.Current.ActiveView, wall, true, false, Autodesk.Revit.DB.HorizontalAlignmentStyle.Center, Autodesk.Revit.DB.VerticalAlignmentStyle.Middle, Vector.ByCoordinates(0, 0, 0));
+            var tag = Tag.ByElement(Revit.Application.Document.Current.ActiveView, wall, true, false, Point.ByCoordinates(0, 0, 0), true, Autodesk.Revit.DB.HorizontalAlignmentStyle.Center.ToString(), Autodesk.Revit.DB.VerticalAlignmentStyle.Middle.ToString());
```
QilongTang: The API does not accept Enum any more, instead it asks for string name of the Enum. Also added offSet point and isOffSet Point signature changes.

mjkkirschner: I'm not so sure this are API changes, I think they are just merge conflicts from nodes that max has added. Tag.ByElement should refer to his Revit Element wrapper class.

PR #1225, opened 2016-08-26
Merged pr annotations 2016

moethu:
### Purpose
One big PR containing:
FilledRegions, Revisions, Parameters, Detail Curves and Material&Elements
### Declarations

### Reviewers

@mjkkirschner 

### FYIs


src/Libraries/RevitNodes/Elements/DetailCurve.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        #region Public static constructors
+
+        /// <summary>
+        /// Construct a Revit DetailCurve element from a curve
+        /// </summary>
+        /// <param name="view">View to place the detail curve on</param>
+        /// <param name="curve">Curve to create detailcurve from</param>
+        /// <returns></returns>
+        public static DetailCurve ByCurve(Revit.Elements.Views.View view, Autodesk.DesignScript.Geometry.Curve curve)
```
mjkkirschner: I'm a bit concerned about this class have thoroughly tested it with many types of curves? have you looked at https://github.com/DynamoDS/DynamoRevit/blob/60325b5dd793e1e06668869ab8d4f88bc0c950ff/src/Libraries/RevitNodes/Elements/ModelCurve.cs

there was a lot of work there to identify changing curves and sketch plane reuse - is the `detail curve` similar to the model curve or doesn't have these concerns?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -165,6 +165,14 @@ public string UniqueId
         private ElementId internalId;
 
         /// <summary>
+        /// Get Element Category
+        /// </summary>
+        public Category GetCategory
```
mjkkirschner: is this a Revit Db `Category`? or a Dynamo wrapper being returned here?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+        /// Update an existing element's location
+        /// </summary>
+        /// <param name="geometry">New Location Point or Curve</param>
+        public void SetLocation(Geometry geometry)
+        {
+            TransactionManager.Instance.EnsureInTransaction(Application.Document.Current.InternalDocument);
+
+            if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationPoint))
+            {
+                if (geometry.GetType() == typeof(Autodesk.DesignScript.Geometry.Point))
```
mjkkirschner: can you change these checks to (if geometry is Point) etc..

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+                    Curve dynamoCurve = geometry as Curve;
+                    Autodesk.Revit.DB.LocationCurve curve = (Autodesk.Revit.DB.LocationCurve)this.InternalElement.Location;
+                    curve.Curve = dynamoCurve.ToRevitType(true);
+                }
+                else
+                    throw new Exception("Curve element required");
+            }
+            else
+            {
+                throw new Exception("Element does not have a location that can be updated");
```
mjkkirschner: these strings should be in the resources file please

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+
+        /// <summary>
+        /// Get an exsiting element's location
+        /// </summary>
+        /// <returns>Location Geometry</returns>
+        public Geometry GetLocation()
+        {
+            if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationPoint))
+            {
+                Autodesk.Revit.DB.LocationPoint pt = (Autodesk.Revit.DB.LocationPoint)this.InternalElement.Location;
```
mjkkirschner: var pt = this.InternalElement.Location as Autodesk.Revit.DB.LocationPoint

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+                return pt.Point.ToPoint(true);
+            }
+            else if (this.InternalElement.Location.GetType() == typeof(Autodesk.Revit.DB.LocationCurve))
```
mjkkirschner: same thing here pleas use `is` if you can instead of getting the types

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+            {
+                Autodesk.Revit.DB.LocationCurve curve = (Autodesk.Revit.DB.LocationCurve)this.InternalElement.Location;
+                return curve.Curve.ToProtoType(true);
+            }
+            else
+            {
+                throw new Exception("Element location not extractable");
```
mjkkirschner: strings to resource file

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+
+        /// <summary>
+        /// Move Revit Element by Vector
+        /// </summary>
+        /// <param name="vector">Translation Vector</param>
+        public void MoveByVector(Vector vector)
+        {
+            if (!this.InternalElement.Location.Move(vector.ToXyz(true)))
+            {
+                throw new Exception("Element could not be moved");
```
mjkkirschner: string to resource file

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+        #endregion
+
+        #region Material
+
+        /// <summary>
+        /// Get Material Names from a Revit Element
+        /// </summary>
+        /// <param name="paintMaterials">Paint Materials</param>
+        /// <returns>List of Names</returns>
+        public List<Material> GetMaterials(bool paintMaterials = false)
```
mjkkirschner: any reason this can't be an Ienumerable<Material>

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -492,12 +492,50 @@ internal static FamilyInstance FromExisting(Autodesk.Revit.DB.FamilyInstance fam
...
-        #region Public methods
+        /// <summary>
+        /// Get FamilyInstance Host Element
+        /// </summary>
+        public Element GetHost
+        {
+            get
+            {
+                if (this.InternalFamilyInstance.Host != null)
+                    return ElementWrapper.Wrap(this.InternalFamilyInstance.Host, true);
```
mjkkirschner: tricky... what happens if the host element was actually created by Dynamo? Can you check it's ownership, I can't recall?

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -492,12 +492,50 @@ internal static FamilyInstance FromExisting(Autodesk.Revit.DB.FamilyInstance fam
...
+                else
+                    return null;
+            }
+        }
+
+        /// <summary>
+        /// Get Family
+        /// </summary>
+        public Family GetFamily
```
mjkkirschner: can you add to the summaries here and in family type on the difference between the two.

src/Libraries/RevitNodes/Elements/InternalUtilities/ElementUtils.cs
```diff
@@ -48,5 +51,65 @@ public static void UpdateLevelName(ref string name)
...
             }
         }
+
+
+        /// <summary>
+        /// Get a revit parameters value
+        /// </summary>
+        /// <param name="param">Revit parameter</param>
+        /// <returns></returns>
+        public static object GetParameterValue(Autodesk.Revit.DB.Parameter param)
```
mjkkirschner: this is going to be visible in the library and is going to expose the `Autodesk.Revit.DB` namespace in Dynamo, either this should be internal or suppressed - we should not expose this namespace in Dynamo.

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -101,6 +101,186 @@ public static Material ByName(string name)
...
+        /// </summary>
+        public DSCore.Color Color
+        {
+            get { return ToDSColor(this.InternalMaterial.Color); }
+        }
+
+        /// <summary>
+        /// Get cut pattern id
+        /// </summary>
+        public Autodesk.Revit.DB.ElementId CutPatternId
```
mjkkirschner: I don't think we want to expose the elementID here, don't we have some way to wrap this?

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -101,6 +101,186 @@ public static Material ByName(string name)
...
+        {
+            get
+            {
+                // Get the active Document
+                Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
+
+                List<Parameter> appearances = new List<Parameter>();
+                if (this.InternalMaterial.AppearanceAssetId != Autodesk.Revit.DB.ElementId.InvalidElementId)
+                {
+                    Autodesk.Revit.DB.AppearanceAssetElement appearance = (Autodesk.Revit.DB.AppearanceAssetElement)document.GetElement(this.InternalMaterial.AppearanceAssetId);
```
mjkkirschner:
``` var appearance = document.GetElement(this.InternalMaterial.AppearanceAssetId) as Autodesk.Revit.DB.AppearanceAssetElement;

if(appearance != null)
{
// do stuff

}

File: src/Libraries/RevitNodes/Elements/Material.cs [hunk 15]
```diff
@@ -101,6 +101,186 @@ public static Material ByName(string name)
...
+
+                    foreach (Autodesk.Revit.DB.Parameter parameter in appearance.Parameters)
```
mjkkirschner: var parameter

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -101,6 +101,186 @@ public static Material ByName(string name)
...
+        public IEnumerable<Parameter> ThermalParameters
+        {
+            get
+            {
+                Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
+
+                List<Parameter> thermals = new List<Parameter>();
+                if (this.InternalMaterial.ThermalAssetId != Autodesk.Revit.DB.ElementId.InvalidElementId)
+                {
+                    Autodesk.Revit.DB.PropertySetElement thermal = (Autodesk.Revit.DB.PropertySetElement)document.GetElement(this.InternalMaterial.ThermalAssetId);
```
mjkkirschner: same `null` check,  `var` and `as` style here please.

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -101,6 +101,186 @@ public static Material ByName(string name)
...
+        public IEnumerable<Parameter> StructuralParameters
+        {
+            get
+            {
+                Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
+
+                List<Parameter> structurals = new List<Parameter>();
+                if (this.InternalMaterial.StructuralAssetId != Autodesk.Revit.DB.ElementId.InvalidElementId)
+                {
+                    Autodesk.Revit.DB.PropertySetElement structural = (Autodesk.Revit.DB.PropertySetElement)document.GetElement(this.InternalMaterial.StructuralAssetId);
```
mjkkirschner: same as above

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -25,12 +34,168 @@ public string Name
...
+            else
+            {
+                return new Parameter(param);
+            }
+        }
+
+        /// <summary>
+        /// Set the value of the parameter
+        /// </summary>
+        public static void SetValue(Parameter parameter, object value)
```
mjkkirschner: @moethu I think this implementation basically duplicates the implementation here:
https://github.com/DynamoDS/DynamoRevit/blob/Revit2016/src/Libraries/RevitNodes/Elements/Element.cs#L435

can you find a way to unify them using the dynamic implementation on the element type.

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
             }
             return string.Format("{0} : {1}", Name, value);
         }
+
+        #endregion
+
+        #region Shared Parameters
+
+        /// <summary>
+        /// Get shared parameter file path
```
mjkkirschner: can you explain the logic behind this to users? something created by default etc?

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+        /// <param name="parameterName">Name</param>
+        /// <param name="groupName">Group of the parameter for shared parameters</param>
+        /// <param name="type">Parameter Type</param>
+        /// <param name="group">Parameter Group</param>
+        /// <param name="instance">Is instance parameter, otherwise its a type parameter</param>
+        /// <param name="categoryList">List of categories this parameter applies to</param>
+        public static void CreateSharedParameter(string parameterName, string groupName, string type, string group, bool instance, [DefaultArgumentAttribute("null")]System.Collections.Generic.IEnumerable<Category> categoryList)
+        {
+            // parse parameter type
+            Autodesk.Revit.DB.ParameterType parameterType = Autodesk.Revit.DB.ParameterType.Text;
```
mjkkirschner: please use var

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+            if (!System.Enum.TryParse<Autodesk.Revit.DB.ParameterType>(type, out parameterType))
```
mjkkirschner: am I correct that this will turn into a string comparison later?

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+                throw new System.Exception(Properties.Resources.ParameterTypeNotFound);
+
+            // parse parameter group
+            Autodesk.Revit.DB.BuiltInParameterGroup parameterGroup = Autodesk.Revit.DB.BuiltInParameterGroup.PG_DATA;
```
mjkkirschner: var please

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+                DefinitionGroup groupDef = document.Application.OpenSharedParameterFile().Groups.get_Item(groupName);
+                if (groupDef == null)
+                {
+                    groupDef = document.Application.OpenSharedParameterFile().Groups.Create(groupName);
+                }
+
+                // If the parameter definition does not exist yet, create it
+                if (groupDef.Definitions.get_Item(parameterName) == null)
+                {
+                    ExternalDefinition def = groupDef.Definitions.Create(new ExternalDefinitionCreationOptions(parameterName, parameterType)) as ExternalDefinition;
```
mjkkirschner: please split the lines

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+
+                    // Apply instance or type binding
+                    Binding bin = (instance) ? (Binding)document.Application.Create.NewInstanceBinding(categories) : (Binding)document.Application.Create.NewTypeBinding(categories);
```
mjkkirschner: please split the lines

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+        /// <summary>
+        /// Create Project Parameter
+        /// </summary>
+        /// <param name="parameterName">Name</param>
+        /// <param name="groupName">Group of the parameter for shared parameters</param>
+        /// <param name="type">Parameter Type</param>
+        /// <param name="group">Parameter Group</param>
+        /// <param name="instance">Is instance parameter, otherwise its a type parameter</param>
+        /// <param name="categoryList">List of categories this parameter applies to</param>
+        public static void CreateProjectParameter(string parameterName, string groupName, string type, string group, bool instance, [DefaultArgumentAttribute("null")]System.Collections.Generic.IEnumerable<Category> categoryList)
```
mjkkirschner: can you split this into multiple lines- 

additionally can you make an AllCategories token that we can use instead of null? the intention will become 100% more clear.

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+        {
+            // parse parameter type
+            Autodesk.Revit.DB.ParameterType parameterType = Autodesk.Revit.DB.ParameterType.Text;
```
mjkkirschner: var - var everywhere!

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+            if (!System.Enum.TryParse<Autodesk.Revit.DB.ParameterType>(type, out parameterType))
+                throw new System.Exception(Properties.Resources.ParameterTypeNotFound);
+
+            // parse parameter group
+            Autodesk.Revit.DB.BuiltInParameterGroup parameterGroup = Autodesk.Revit.DB.BuiltInParameterGroup.PG_DATA;
```
mjkkirschner: var

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+            TransactionManager.Instance.TransactionTaskDone();
+        }
+
+
+        #endregion
+
+        #region Project Parameters
+
+        /// <summary>
+        /// Create Project Parameter
```
mjkkirschner: some more info for users here please

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+            // get document and open transaction
+            Document document = Application.Document.Current.InternalDocument;
+            TransactionManager.Instance.EnsureInTransaction(document);
+
+            try
+            {
+                // buffer the current shared parameter file name and apply a new empty parameter file instead
+                string sharedParameterFile = document.Application.SharedParametersFilename;
+                string tempSharedParameterFile = System.IO.Path.GetTempFileName() + ".txt";
+                using (System.IO.File.Create(tempSharedParameterFile)) { }
```
mjkkirschner: does this require a using statement?

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+                throw new System.Exception(Properties.Resources.ParameterTypeNotFound);
+
+            // get document and open transaction
+            Document document = Application.Document.Current.InternalDocument;
+            TransactionManager.Instance.EnsureInTransaction(document);
+
+            try
+            {
+                // buffer the current shared parameter file name and apply a new empty parameter file instead
+                string sharedParameterFile = document.Application.SharedParametersFilename;
```
mjkkirschner: can this be null?

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+                string tempSharedParameterFile = System.IO.Path.GetTempFileName() + ".txt";
+                using (System.IO.File.Create(tempSharedParameterFile)) { }
+                document.Application.SharedParametersFilename = tempSharedParameterFile;
+
+                // Apply selected parameter categories
+                CategorySet categories = ToCategorySet(categoryList);
+
+                // create a new shared parameter, since the file is empty everything has to be created from scratch
+                ExternalDefinition def = document.Application.OpenSharedParameterFile().Groups.Create(groupName).Definitions.Create(new ExternalDefinitionCreationOptions(parameterName, parameterType)) as ExternalDefinition;
```
mjkkirschner: multi line please

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+
+                // Create an instance or type binding
+                Binding bin = (instance) ? (Binding)document.Application.Create.NewInstanceBinding(categories) : (Binding)document.Application.Create.NewTypeBinding(categories);
```
mjkkirschner: multi line

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+                throw e;
+            }
+
+            TransactionManager.Instance.TransactionTaskDone();
+        }
+
+        #endregion
+
+        /// <summary>
+        /// Transform List of categories to CategorySet
```
mjkkirschner: into a ?

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +205,190 @@ public override string ToString()
...
+        /// <param name="categoryList"></param>
+        /// <returns></returns>
+        internal static CategorySet ToCategorySet(System.Collections.Generic.IEnumerable<Category> categoryList)
+        {
+            Document document = Application.Document.Current.InternalDocument;
+
+            // Apply selected parameter categories
+            CategorySet categories = document.Application.Create.NewCategorySet();
+
+            if (categoryList == null)
```
mjkkirschner: if you can create that allCategory token you would handle that case before calling this method I guess

src/Libraries/RevitNodes/Elements/Revision.cs
```diff
@@ -0,0 +1,261 @@
...
+            // Get existing element if any
+            var RevisionElem = ElementBinder.GetElementFromTrace<Autodesk.Revit.DB.Revision>(document);
+
+            if (RevisionElem == null)
+            { 
+                RevisionElem = Autodesk.Revit.DB.Revision.Create(document); 
+            }
+            
+            // Apply properties
+            RevisionElem.Visibility = visibility;
```
mjkkirschner: should check that these properties are not the same before applying them.

src/Libraries/RevitNodes/Elements/Revision.cs
```diff
@@ -0,0 +1,261 @@
...
+        /// <param name="revDate">Revision Date</param>
+        /// <param name="description">Description</param>
+        /// <param name="issued">Issuing status</param>
+        /// <param name="issuedBy">Issued by</param>
+        /// <param name="issuedTo">Issued to</param>
+        /// <param name="numberType">Number type</param>
+        /// <returns></returns>
+        public static Revision ByName(string name, string revDate, string description, bool issued, string issuedBy, string issuedTo, string visibility = "",string numberType = "")
+        {
+            RevisionVisibility revVisibility = RevisionVisibility.CloudAndTagVisible;
```
mjkkirschner: any reason to initialize them if you are setting them to defaults below?

src/Libraries/RevitNodes/Elements/RevisionCloud.cs
```diff
@@ -0,0 +1,209 @@
...
+        internal Autodesk.Revit.DB.RevisionCloud InternalRevitElement
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Reference to the Element
+        /// </summary>
+        public override Autodesk.Revit.DB.Element InternalElement
```
mjkkirschner: suppress this

src/Libraries/RevitNodes/Elements/RevisionCloud.cs
```diff
@@ -0,0 +1,209 @@
...
+    /// </summary>
+    [DynamoServices.RegisterForTrace]
+    public class RevisionCloud : Element
+    {
+        #region Internal Properties
+
+        /// <summary>
+        /// Internal reference to the Revit Element
+        /// </summary>
+        [SupressImportIntoVM]
```
mjkkirschner: this doesn't need to be suppressed

src/Libraries/RevitNodes/Elements/RevisionCloud.cs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+
+
+        #endregion
+
+        #region Internal static constructors
+
+        /// <summary>
+        /// Create a Rebar from an existing reference
```
mjkkirschner: typo

src/Libraries/RevitNodes/Properties/Resources.resx
```diff
@@ -321,13 +321,22 @@
...
-  <data name="CannotGetBaseOrSurveyPoint" xml:space="preserve">
-    <value>Cannot get base or survey point from document.</value>
+  <data name="NoSharedParameterFileFound" xml:space="preserve">
+    <value>No shared parameter file found.</value>
+  </data>
+  <data name="ParameterGroupNotFound" xml:space="preserve">
+    <value>Invalid parameter group.</value>
+  </data>
+  <data name="ParameterTypeNotFound" xml:space="preserve">
+    <value>Invalid parameter type.</value>
```
mjkkirschner: resource files do not match?

test/Libraries/RevitNodesTests/Elements/DetailCurveTests.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    [TestFixture]
+    public class DetailCurveTests : RevitNodeTestBase
+    {
+        /// <summary>
+        /// Revit Detail Curve Test
+        /// </summary>
+        [Test]
+        [TestModel(@".\emptyAnnotativeView.rvt")]
+        public void ByCurve_ValidArgs()
```
mjkkirschner: can you add some more complex curve tests?

test/Libraries/RevitNodesTests/Elements/FilledRegionTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Linq;
+
+using Autodesk.DesignScript.Geometry;
+
+using Revit.Elements;
+using NUnit.Framework;
+
+using RevitTestServices;
+
+using RTF.Framework;
```
mjkkirschner: where is this used?

mjkkirschner:
@moethu these look solid. Please address the issues I've identified here and it would be great to increase the number of node tests to cover more of the public methods and properties (nodes) you have added...

src/Libraries/RevitNodes/Elements/Material.cs
mjkkirschner: or just an int?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -710,5 +670,108 @@ protected bool IsAlive
...
+            {
+                throw new Exception("Element location not extractable");
+            }
+        }
+
+        /// <summary>
+        /// Move Revit Element by Vector
+        /// </summary>
+        /// <param name="vector">Translation Vector</param>
+        public void MoveByVector(Vector vector)
```
mjkkirschner: **have you tested this method does not create an infinite loop when used in automatic run mode?**

src/Libraries/RevitNodes/Elements/FilledRegion.cs
```diff
@@ -0,0 +1,174 @@
...
+            if (region == null)
+            { 
+                // Create new region
+                region = Autodesk.Revit.DB.FilledRegion.Create(document, typeId, view.Id, boundary.ToList()); 
+            }
+            else
+            {
+                // Delete and replace region because there are no properties to update the filled region
+                document.Delete(region.Id);
+                region = Autodesk.Revit.DB.FilledRegion.Create(document, typeId, view.Id, boundary.ToList());
```
mjkkirschner: I think you'll need to add some logic for comparing the inputs and bailing if they are the same to avoid infinite loops when in run auto mode. There should be some curve comparison logic in the curve utilities classes we already have.

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -101,6 +101,186 @@ public static Material ByName(string name)
...
+        /// </summary>
+        public int Transparency
+        {
+            get { return this.InternalMaterial.Transparency; }
+        }
+
+        /// <summary>
+        /// Get SurfacePatternColor
+        /// </summary>
+        public DSCore.Color SurfacePatternColor
```
mjkkirschner: did you have to add a new reference for this or we already referenced - just curious.

src/Libraries/RevitNodes/Elements/RevisionCloud.cs
```diff
@@ -0,0 +1,209 @@
...
+            if (element == null)
+            {
+                element = Autodesk.Revit.DB.RevisionCloud.Create(document, view, id, curves.ToList());
+            }
+            else
+            {
+                // there is no way of updateing the properties.
+                // So the cloud is being recreated.
+                document.Delete(element.Id);
+                element = Autodesk.Revit.DB.RevisionCloud.Create(document, view, id, curves.ToList());
```
mjkkirschner: I think you should potentially do a comparison between geometry, can you check the import instance code to see what is done there? it's a similar situation where the element is always destroyed and recreated I think.

mjkkirschner:
@moethu I sent a few more comments where I think you need to do comparisons between the set properties on the element and the input arguments to make sure not to go into infinite loops in run auto mode. There are a few spots where I'm not sure if its necessary when you always destroy the element since it can't be re-bound, check the import instance code to see what is done, I believe this is a similar situation.

src/Libraries/RevitNodes/Elements/DetailCurve.cs
moethu: @mjkkirschner added planarity check, PolyCurve check and pulling it onto XY plane first now. There is no sketchplane needed for DetailCurves since they should be on XY.

src/Libraries/RevitNodes/Elements/Element.cs
moethu: Wrapper is returned (see statement below)
``` return new Category(this.InternalElement.Category);```

File: src/Libraries/RevitNodes/Elements/FamilyInstance.cs (hunk 10 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/FamilyInstance.cs]: that's why I'm checking for null - I added a nicer exception for this.

File: src/Libraries/RevitNodes/Elements/InternalUtilities/ElementUtils.cs (hunk 12 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/InternalUtilities/ElementUtils.cs]: sure, will do. I assumed the class attribute ([IsVisibleInDynamoLibrary(false)]) would be inherited for all methods and therefore be invisible as well.

File: src/Libraries/RevitNodes/Elements/Material.cs (hunk 13 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Material.cs]: will return an Int

File: src/Libraries/RevitNodes/Elements/Material.cs (hunk 15 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Material.cs]: ok, but then I'll have to cast in the line below:
```Parameter p = new Parameter(parameter as Autodesk.Revit.DB.Parameter);```
because a var won't be accepted by the Parameter constructor.

File: src/Libraries/RevitNodes/Elements/Parameter.cs (hunk 18 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Parameter.cs]: In a way yes, but I'd say my implementation is more flexible and handles more cases (eg. if the parameter expects an ElementId you can submit an int or an Element. Same for Int, you can submit a double, int or string to get to the correct result.

File: src/Libraries/RevitNodes/Elements/Parameter.cs (hunk 21 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Parameter.cs]: yes, this is the string comparison already. It does a tryparse on the enum string and continues with a default value if it didn't work out.

File: src/Libraries/RevitNodes/Elements/Parameter.cs (hunk 25 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Parameter.cs]: yes that makes sense, but since I'm using a list of categories, I'd need a AllElements category which doesn't exist. Or what should I pass?

File: src/Libraries/RevitNodes/Elements/Parameter.cs (hunk 30 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Parameter.cs]: yes, handled directly below

File: src/Libraries/RevitNodes/Elements/Parameter.cs (hunk 34 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Parameter.cs]: ah, an overload is a nice solution, will do that.

File: src/Libraries/RevitNodes/Elements/Revision.cs (hunk 36 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Revision.cs]: doing a tryparse afterwards. the default is only a fallback if the tryparse fails.

File: src/Libraries/RevitNodes/Elements/Element.cs (hunk 43 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Element.cs]: works in manual tests (Automatic mode) it runs only once and is not being executed again. Since the node has been run. Its not being invalidated and therefore re-executed.

File: src/Libraries/RevitNodes/Elements/FilledRegion.cs (hunk 44 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/FilledRegion.cs]: but I can't update the element anyway. This doesn't cause an infinite loop for me.

File: src/Libraries/RevitNodes/Elements/Material.cs (hunk 45 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/Material.cs]: DSCore.Color? Sometimes I had to add it. I think the DSCore.Color hasn't been used for Revit so far. It was the only color I found in the DS Library.

File: src/Libraries/RevitNodes/Elements/RevisionCloud.cs (hunk 46 above)
Comment (moethu) [inline src/Libraries/RevitNodes/Elements/RevisionCloud.cs]: ImportInstance just deletes the element first. Changed this in both (RevCloud & FilledRegion).

**Comment by moethu** [comment]
@mjkkirschner applied all the review comments. Some are still open (see questions above)

File: src/Libraries/RevitNodes/Elements/FamilyInstance.cs (hunk 10 above)
Comment (mjkkirschner) [inline src/Libraries/RevitNodes/Elements/FamilyInstance.cs]: I mean, what if you manually created something in Revit on top of a element that was created by Dynamo.

IE. someone creates a wall with Dynamo so this is Dynamo owned, then they create a window on this wall, and then call getHost on it?

File: src/Libraries/RevitNodes/Elements/InternalUtilities/ElementUtils.cs [hunk 47]
```diff
@@ -58,6 +59,7 @@ public static void UpdateLevelName(ref string name)
         /// </summary>
         /// <param name="param">Revit parameter</param>
         /// <returns></returns>
+        [IsVisibleInDynamoLibrary(false)]
         public static object GetParameterValue(Autodesk.Revit.DB.Parameter param)
```
mjkkirschner: I think you need to `suppress` this to stop the Revit.DB type from leaking in.

src/Libraries/RevitNodes/Elements/RevisionCloud.cs
```diff
@@ -188,7 +179,7 @@ public IEnumerable<Autodesk.DesignScript.Geometry.Curve> Curves
         #region Internal static constructors
 
         /// <summary>
-        /// Create a Rebar from an existing reference
+        /// Create Rebar from existing reference
```
mjkkirschner: ... the type is that this is a revision cloud not a rebar right?

src/Libraries/RevitNodes/Elements/Parameter.cs
mjkkirschner: That is true, but I prefer the clarity of the dynamic methods. In either case, can you unify on one implementation and call it from both places?

test/Libraries/RevitNodesTests/Elements/ParameterTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            var wall = fec.FirstElement();    
+            Assert.IsNotNull(wall.LookupParameter("MySharedParameter"));
+        }
+
+        /// <summary>
+        /// Set a parameters value
+        /// </summary>
+        [Test]
+        [TestModel(@".\element.rvt")]
+        public void SetParameterValue_ValidArgs()
```
mjkkirschner: please add tests for different parameter type inputs

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
moethu: This would return the wall since the element exists in Revit. The window always has a Host element.

mjkkirschner:
@moethu I think I have addressed all questions - let me know if I missed some.

about the parameters - My opinion is that since this functionality is highly used *(parameter setting)* and I'd like to only have one implementation in the code base -  and the version we have now is highly tested and has been around for a while, can you please unify the solution using the existing dynamic dispatch implementation?

We can incorporate your more forgiving solution in the future, thanks :)

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
mjkkirschner: right - but it would be tagged as owned by Revit now I think?

src/Libraries/RevitNodes/Elements/FilledRegion.cs
mjkkirschner: okay, thanks for testing.

src/Libraries/RevitNodes/Elements/InternalUtilities/ElementUtils.cs
mjkkirschner: isVisible won't stop the input parameter from being imported into the VM I don't think, so the revit type will still leak into the library search.

src/Libraries/RevitNodes/Elements/Material.cs
mjkkirschner: ahh, okay leave as is.

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
mjkkirschner: if this is not feasible to solve quickly, do not worry about it. But please add a comment / TODO and I will add a task to followup.

src/Libraries/RevitNodes/Elements/Parameter.cs
mjkkirschner: ok

src/Libraries/RevitNodes/Elements/Revision.cs
mjkkirschner: ok

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
moethu: ok, thanks - added a comment . I haven't figured a way to determine if its Revit owned or not.

src/Libraries/RevitNodes/Elements/Parameter.cs
moethu: sure, I'll unify and move it to the element utils

moethu: I used the existing implementation to speed things up

moethu:
@mjkkirschner I moved the set Parameter by value from element to elementUtils and I'm using it in Element and Parameters now.

src/Libraries/RevitNodes/Elements/Parameter.cs
mjkkirschner: @moethu was this overload created?

src/Libraries/RevitNodes/Elements/Revision.cs
```diff
@@ -0,0 +1,288 @@
...
+        internal Autodesk.Revit.DB.Revision InternalRevitElement
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Reference to the Element
+        /// </summary>
+        public override Autodesk.Revit.DB.Element InternalElement
```
mjkkirschner: please suppress

mjkkirschner:
@moethu it's looking good! - I have one issue I missed and one outstanding question.
1 - suppress the public element in revision
2 - overload for the categorySet so you don't have to pass null

src/Libraries/RevitNodes/Elements/Parameter.cs
moethu: yes, there is a one in parameters line 199

moethu:
@mjkkirschner the overload is for createSharedParameter and createProjectParameter, the categorySet isn't exposed anyway. its internal only.

moethu:
@mjkkirschner added two methods, one for all categories another for transforming it.

src/Libraries/RevitNodes/Elements/Parameter.cs
```diff
@@ -40,5 +165,244 @@ public override string ToString()
...
+        /// <summary>
+        /// Create a new Shared Parameter in the current Revit document
+        /// </summary>
+        /// <param name="parameterName">Name</param>
+        /// <param name="groupName">Group of the parameter for shared parameters</param>
+        /// <param name="type">Parameter Type</param>
+        /// <param name="group">Parameter Group</param>
+        /// <param name="instance">Is instance parameter, otherwise its a type parameter</param>
+        /// <param name="categoryList">List of categories this parameter applies to, If no category is supplied, all possible categories are selected</param>
+        public static void CreateSharedParameter(string parameterName, string groupName, string type, string group, bool instance, System.Collections.Generic.IEnumerable<Category> categoryList)
```
mjkkirschner: I see about the overload now I did not realize it was for the public method, is there any issue with actually getting rid of the public overloads? Can you simply change the name of this method or the other one?

mjkkirschner:
@moethu thanks for work here - I misunderstood about the public overload - is it an issue to get rid of them? I'd like to avoid public overloads where possible ... it's a VM future proofing issue.

moethu:
@mjkkirschner np, just renamed them.

mjkkirschner:
@QilongTang @jnealb @kronz @sharadkjaiswal  I am merging these PRs in... things might get rocky, @QilongTang, let's make sure we run the 2018 tests on that branch since it's not automated yet.

PR #1237, opened 2016-08-30
filter dynamo versions that support playlist on startup

BogdanZavu:
### Purpose

When I start Revit I want to see the Playlist button enabled if there is at least one version of Dynamo installed that supports Playlist feature. 
If there are more than 1 version , then I'll see the usual version selector dialog when I press the Playlist button. The dialog will show only Dynamo versions that support Playlist and not all Dynamo versions installed ( as it is the case for Dynamo button ).

### Declarations

### Reviewers

@mjkkirschner 

### FYIs
@kronz

### DIFF
--- a/src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
+++ b/src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
@@ -36,9 +36,11 @@ namespace Dynamo.Applications
     [Regeneration(RegenerationOption.Manual)]
     public class VersionLoader : IExternalApplication
     {
-        private static UIControlledApplication uiApplication;
+        private static UIControlledApplication uiControlledApplication;
         private AddInCommandBinding dynamoCommand;
+        private AddInCommandBinding dynamoPlayerCommand;
         internal List<DynamoProduct> Products { get; private set; }
+        internal List<DynamoProduct> PlaylistProducts { get; private set; }
 
         internal static string GetDynamoRevitPath(DynamoProduct product, string revitVersion)
         {
@@ -50,7 +52,7 @@ namespace Dynamo.Applications
 
         public Result OnStartup(UIControlledApplication application)
         {
-            uiApplication = application;
+            uiControlledApplication = application;
             var revitVersion = application.ControlledApplication.VersionNumber;
 
             var revitFolder =
@@ -60,6 +62,7 @@ namespace Dynamo.Applications
             var dynamoProducts = FindDynamoRevitInstallations(debugPath, revitVersion);
 
             Products = new List<DynamoProduct>();
+            PlaylistProducts = new List<DynamoProduct>();
             foreach (var p in dynamoProducts)
             {
                 var path = VersionLoader.GetDynamoRevitPath(p, revitVersion);
@@ -67,7 +70,20 @@ namespace Dynamo.Applications
                     continue;
 
                 Products.Add(p);
+
+                if (p.VersionInfo.Build >= 1 && p.VersionInfo.Major >= 1)
+                {
+                    if (Convert.ToInt64(revitVersion) >= 2017)
+                    {
+                        PlaylistProducts.Add(p);
+                    }
+                }
             }
+
+            if (Products.Count == 0)
+                return Result.Failed;
+
+            Result preliminaryLoadResult = Result.Succeeded;
 
             // If there are multiple versions installed, then do the command 
             // binding to prompt for version selector task dialog for user to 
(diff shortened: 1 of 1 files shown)


kronz:
Related question but might not be addressed in this PR.  What happens if a user starts an unsupported version of Dynamo in a session (Revit with then ONLY use the unsupported version until restart of Revit) and then uses Playlist?

BogdanZavu:
@kronz If by an unsupported version of Dynamo you refer to a previous version of Dynamo like 0.9 which doesn't support Playlist then when 0.9 is chosen the Playlist button becomes disabled for the current Revit session.

kronz:
@BogdanZavu ok, thanks.  Seems sensible

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -36,9 +36,11 @@ public string VersionString
     [Regeneration(RegenerationOption.Manual)]
     public class VersionLoader : IExternalApplication
     {
-        private static UIControlledApplication uiApplication;
+        private static UIControlledApplication uiControlledApplication;
         private AddInCommandBinding dynamoCommand;
+        private AddInCommandBinding dynamoPlayerCommand;
         internal List<DynamoProduct> Products { get; private set; }
+        internal List<DynamoProduct> PlaylistProducts { get; private set; }
```
mjkkirschner: can you add a comment on this property?

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -60,15 +62,29 @@ public Result OnStartup(UIControlledApplication application)
...
+            PlaylistProducts = new List<DynamoProduct>();
             foreach (var p in dynamoProducts)
             {
                 var path = VersionLoader.GetDynamoRevitPath(p, revitVersion);
                 if (!File.Exists(path))
                     continue;
 
                 Products.Add(p);
+
+                if (p.VersionInfo.Build >= 1 && p.VersionInfo.Major >= 1)
```
mjkkirschner: shouldn't this be `Major` and `Minor` ? https://msdn.microsoft.com/en-us/library/system.version(v=vs.110).aspx#Anchor_6

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -60,15 +62,29 @@ public Result OnStartup(UIControlledApplication application)
...
+                {
+                    if (Convert.ToInt64(revitVersion) >= 2017)
```
mjkkirschner: can we store all these numbers in some `constants somewhere`

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -79,29 +95,52 @@ public Result OnStartup(UIControlledApplication application)
...
 
         void executed(object sender, ExecutedEventArgs e)
         {
-            var product = PromptVersionSelectorDialog();
+            var product = PromptVersionSelectorDialog(Products);
             if(product.HasValue)
                 LaunchDynamoCommand(product.Value, e);
         }
 
+        void executedPlaylist(object sender, ExecutedEventArgs e)
```
mjkkirschner: can you add an explicit private access modifier

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -112,11 +151,25 @@ void canExecute(object sender, CanExecuteEventArgs e)
...
+            var revitApp = ass.CreateInstance("Dynamo.Applications.DynamoRevitApp");
+            revitApp.GetType().GetMethod("OnStartup").Invoke(revitApp, new object[] { uiControlledApplication });
+
+            return Result.Succeeded;
+        }
+
         /// <summary>
         /// Prompts for version selection task dialog
         /// </summary>
         /// <returns>DynamoProduct to launch or null</returns>
```
mjkkirschner: please update the summary (missing parameter)

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
```diff
@@ -174,24 +229,35 @@ private bool LaunchDynamoCommand(DynamoProduct product, ExecutedEventArgs e)
...
+            {
+                //Invoke command
+                string message = string.Empty;
+                result = type.GetMethod("ExecuteDynamoCommand").Invoke(revitApp, new object[] { e.GetJournalData(), uiApp });
+                if ((Result)result != Result.Succeeded)
+                    return false;
+            }
+            else
+            {
+                //search it again since it was re-binded in dependent components
```
mjkkirschner: can you clarify this comment a bit more?

mjkkirschner:
@BogdanZavu a few questions - 
also, I don't think it's easily possible now with the current state of RTF - but I want to investigate if we can write a test to check this functionality in the future. I will checkout what we have existing around startup sequence.

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
BogdanZavu: ok

src/Legacy/DynamoRevitVersionSelector/VersionSelectorApp.cs
BogdanZavu: yes

BogdanZavu:
Regarding RTF , if I'm not mistaken you can't have a test right now that chooses Dynamo interactively ; instead you have to specify at the beginning which Dynamo you want to use . 
I think Revit journaling allows some things to be done in addins ( in this case DynamoRevitVersionSelector ) in order for him to know how to press a button in an addin . So just like we have now a Revit journal that starts Revit and Dynamo we could have a journal that starts Revit , presses Dynamo and chooses a version. 
I need to further check this out if it's really possible , I might be wrong.

mjkkirschner:
@BogdanZavu looks good, can you generate PRs to cherry pick these commits to the 1.2 branches and I'll merge them all at the same time.

jnealb:
Does this need testing?

mjkkirschner:
@jnealb definitely.

mjkkirschner:
@BogdanZavu do you need to make any other changes or are these branches all ready for merge?

BogdanZavu:
@mjkkirschner no other changes are needed ; all branches are ready to merge ; Thank you!

PR #1239, opened 2016-08-31
Exculde template view during View3d wrapping

QilongTang:
### Purpose

[MAGN-5431](http://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-5431) All Elements Of Type throws exception when handling views. Calling IsPerspective() on 3Dviews with IsTemplate marked as true will return exception instead of true/ false. This is idea but waiting for Revit API guild to response the reason behind that. As an attempt to fix on our side, we should avoid such query on template views so the graph can still execute. Added such case in unit tests so we could catch it when it's broken again later.

### Declarations

New results filtering out views:
image

### Reviewers

@sm6srw 

### FYIs

@mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
+++ b/src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
@@ -152,14 +152,14 @@ namespace Revit.Elements
 
         public static View3D Wrap(Autodesk.Revit.DB.View3D view, bool isRevitOwned)
         {
-            if (view.IsPerspective)
+            if (!view.IsTemplate)
             {
-                return PerspectiveView.FromExisting(view, isRevitOwned);
+                if (view.IsPerspective)
+                    return PerspectiveView.FromExisting(view, isRevitOwned);
+                else
+                    return AxonometricView.FromExisting(view, isRevitOwned);
             }
-            else
-            {
-                return AxonometricView.FromExisting(view, isRevitOwned);
-            }
+            return null;
         }
 
         public static Element Wrap(Autodesk.Revit.DB.ViewPlan view, bool isRevitOwned)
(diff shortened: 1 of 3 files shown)


src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
```diff
@@ -152,14 +152,14 @@ public static WallType Wrap(Autodesk.Revit.DB.WallType ele, bool isRevitOwned)
 
         public static View3D Wrap(Autodesk.Revit.DB.View3D view, bool isRevitOwned)
         {
-            if (view.IsPerspective)
+            if (!view.IsTemplate)
```
QilongTang: Exclude template views for now, waiting for API guild's response whether we should totally ignore them or there are other work around.

test/Libraries/RevitIntegrationTests/SelectionTests.cs
```diff
@@ -456,6 +456,11 @@ public void DynamoAllSelectionNodeTests_WithPreSelectedEntities()
                 Assert.IsNotNull(refPt);
             }
 
+            // ElementsOfCategories & Views, as output of Categories 
+            // passed to ElementsOFCategory
+            var elementsOfViewNodeID = "2569434a-b34f-4512-a4fe-f065c7a10175";
+            AssertPreviewCount(elementsOfViewNodeID, 33);
```
QilongTang: Add a new case filtering out all views in the sample.rvt, since template views will appear null, by pass the null check.

src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
sm6srw: I think it is OK to ignore them. I think our users only cares about normal views (that may have a view template applied).

sm6srw:
LGTM

PR #1240, opened 2016-09-01
Add regression test case for element binding of multiple custom node instance

ke-yu:
### Purpose

Regression test case `MultipleCustomNodeInstance` for element binding of multiple custom node instances (MAGN-10165): Open revit project which contains 8 walls that generated from 2 custom node instances and run that .dyn file, verify no new wall will be added; update code block and re-execute, verify no new wall will be added. 

### Declarations

### FYIs

@riteshchandawar


QilongTang:
@ke-yu Do we need this in other branches like 2018? Thanks.

ke-yu:
@QilongTang oh yeah, forget! :-( thanks for reminding!

PR #1291, opened 2016-09-17
Mark detail views as annotative

dimven:
### Purpose

Drafting views can host annotative elements like filled regions and detail curves too.

### Declarations

### Reviewers

@moethu


kronz:
@dimven Thanks so much for the contribution! I'm assuming that this addresses the issue we talked about with Marcello that FilledRegion.ByCurves does not work in Drafting view?
@mjkkirschner If the code looks good, I totally support getting this in

dimven:
Yes, it was just a tiny addition.

mjkkirschner:
@dimven - did not notice that your PR was to 1.2RC 2017 so I cherry picked there already, Thanks for this contribution.

PR #1309, opened 2016-09-23
Preload conflict dll between Dynamo4Revit and Revit Host

QilongTang:
### Purpose

[MAGN-10678](https://adsk-oss.myjetbrains.com/youtrack/issue/MAGN-10678) GOT REPRO : Blank and un-editable Code Block nodes with D4R upgrade to 1.2

### Declarations

### Reviewers

@mjkkirschner @sm6srw 

### FYIs

@ramramps @jnealb

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -1,7 +1,3 @@
-using Dynamo.Scheduler;
-
-using Greg.AuthProviders;
-
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -12,12 +8,11 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Windows.Interop;
+using System.Windows.Media;
 using System.Windows.Threading;
-
 using Autodesk.Revit.Attributes;
 using Autodesk.Revit.DB;
 using Autodesk.Revit.UI;
-
 using Dynamo.Applications;
 using Dynamo.Applications.Models;
 using Dynamo.Applications.ViewModel;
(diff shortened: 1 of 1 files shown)


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -367,8 +361,27 @@ public static string GetGeometryFactoryPath(string corePath)
                 new object[] { corePath, Versions.ShapeManager }) as string);
         }
 
+        private static void PreloadDynamoCoreDlls()
+        {
+            // Assume Revit Install folder as look for root. Assembly name is compromised.
```
mjkkirschner: I dont understand this comment.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -367,8 +361,27 @@ public static string GetGeometryFactoryPath(string corePath)
...
+                if(File.Exists(assemblyPath))
+                    Assembly.LoadFrom(assemblyPath);
+            }
+        }
+
         private static RevitDynamoModel InitializeCoreModel(DynamoRevitCommandData commandData)
         {
+            // Temporary fix to pre-load DLLs that were also referenced in Revit folder. 
+            // To do: Need to align with Revit when provided a chance.
+            PreloadDynamoCoreDlls();
```
mjkkirschner: @QilongTang from your analysis did you ever see a assembly resolve event or exception thrown when we load the second dll (whichever it may be)?

mjkkirschner:
I think we should create a followup task to investigate how to get this class of dll loading issue into Dynamo as a notification ... we can re-subscribe the assembly resolve event probably, also we should add a test around this if possible, but I am fine if these are follow up tasks and we merge this in.

src/DynamoRevit/DynamoRevit.cs
QilongTang: I mean although ideally the "assemblyList"stores assembly name, in the future you may see all kinds of stuff here, e.g.
"SDA\\bin\\ICSharpCode.AvalonEdit.dll"
"RevitAPI.dll"
They are all actually relative path comparied to Revit install path.

src/DynamoRevit/DynamoRevit.cs
QilongTang: Nope. No obvisou exception we could catch in this case. I've seen RevitAPI exceptions.

QilongTang:
Agree. I actually was thinking we need to do a full scan to align with Revit dlls before next major release, so we have less wierd behavior in Revit Environment. I am sure it saves us more time than looking at reports one by one and reproduce.

Issue #1264, opened 2016-09-08
Views list not ordered alphabetically


andydandy74:
## Dynamo version

1.1 stable
## Revit version

2017
## Operating system

Win 7
## What did you do?

Used the `Views` node
## What did you expect to see?

An alphabetically sorted list of views
## What did you see instead?

Chaos.

Please return the list of views in alphabetical order.

kronz:
@andydandy74 I'm seeing the same View order listing in 0.9.2 and latest daily builds.  Can you show what you want to return to?

andydandy74:
@kronz I don't want to return to anything - it's not a regression. I'd like to move towards an alphabetical order. This has been bugging me for as long as I've been using Dynamo but somehow I never posted an issue.

kronz:
@andydandy74 ah, sorry, I mis-read your "please return".  i thought you were refering to a previous state.  Looks like this dropdown probalby uses element id for sorting, which is almost never what anyone wants

ksobon:
@andydandy74  this was fixed. Can you close?

andydandy74:
Sure

PR #1313, opened 2016-09-25
Revit2017_UserRequest to sort Views alphabetically.

ksobon:
### Purpose

User @andydandy74 has requested in this issue https://github.com/DynamoDS/DynamoRevit/issues/1264 to display views in alphabetical order. We already to that for most of the dropdown nodes so it makes sense to also do that for views. 

Also, I added a small name filter to remove <RevisionSchedule> items from that list as well. Those are the small revision tables that are automatically placed on every sheet. I have never found anything useful for them, but they clatter the Views dropdown.

Also, changing the output from `SelectionState.Restore` to `SelectionState.Done` seems to automatically set the selected view to index 0 which in this case is OK since there would always be at least one view in the node. Is it ok to do that? 

Ps. What is the difference between these two selection states and what are recommended use cases? 

Before:
image

After:
image

### Declarations

### Reviewers

@ikeough 
@mjkkirschner 

### FYIs

This is my first PR so its nothing major, but I want to test the process. Based on our conversation with Ian on Friday, I might have some Revit based functionality that can be submitted to the repo. I want to give this a whirl before I submit something larger that will possibly cause conflicts and all sorts of things. Let me know if I am doing anything wrong here.

### DIFF
--- a/src/Config/CS.props
+++ b/src/Config/CS.props
@@ -10,9 +10,9 @@
 	<REVIT_VERSION>Revit_$(RevitVersionNumber)</REVIT_VERSION>
 	<OutputPath Condition=" '$(OutputPath)' == '' ">$(SolutionDir)..\bin\$(Platform)\$(Configuration)\$(REVIT_VERSION)</OutputPath>
     <NunitPath Condition=" '$(NunitPath)' == '' ">$(SolutionDir)..\extern\NUnit</NunitPath>
-	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOAPI>
+	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo 0.7</DYNAMOAPI>
 	<DynamoExternPath  Condition=" '$(DynamoExternPath)' == '' ">$(SolutionDir)..\extern</DynamoExternPath>
-	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOTESTAPI>
+	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">C:\Program Files\Dynamo 0.7</DYNAMOTESTAPI>
 	<TestOutputPath Condition=" '$(TestOutputPath)' == '' ">$(SolutionDir)..\bin\$(Platform)\$(Configuration)\$(REVIT_VERSION)</TestOutputPath>
     <REVITAPI  Condition=" '$(REVITAPI)' == '' ">C:\Program Files\Autodesk\Revit Architecture $(RevitVersionNumber)</REVITAPI>
     <REVITAPI Condition=" !Exists('$(REVITAPI)') ">C:\Program Files\Autodesk\Revit $(RevitVersionNumber)</REVITAPI>
@@ -33,9 +33,9 @@
 	<REVIT_VERSION>Revit_$(RevitVersionNumber)</REVIT_VERSION>
 	<OutputPath Condition=" '$(OutputPath)' == '' ">$(SolutionDir)..\bin\$(Platform)\$(Configuration)\$(REVIT_VERSION)</OutputPath>
     <NunitPath Condition=" '$(NunitPath)' == '' ">$(SolutionDir)..\extern\NUnit</NunitPath>
-	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOAPI>
+	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo 0.7</DYNAMOAPI>
 	<DynamoExternPath  Condition=" '$(DynamoExternPath)' == '' ">$(SolutionDir)..\extern</DynamoExternPath>
-	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOTESTAPI>
+	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">C:\Program Files\Dynamo 0.7</DYNAMOTESTAPI>
 	<TestOutputPath Condition=" '$(TestOutputPath)' == '' ">$(SolutionDir)..\bin\$(Platform)\$(Configuration)\$(REVIT_VERSION)</TestOutputPath>
     <REVITAPI  Condition=" '$(REVITAPI)' == '' ">C:\Program Files\Autodesk\Revit Architecture $(RevitVersionNumber)</REVITAPI>
     <REVITAPI Condition=" !Exists('$(REVITAPI)') ">C:\Program Files\Autodesk\Revit $(RevitVersionNumber)</REVITAPI>
--- a/src/Config/CS.props
+++ b/src/Config/CS.props
@@ -10,9 +10,9 @@
 	<REVIT_VERSION>Revit_$(RevitVersionNumber)</REVIT_VERSION>
 	<OutputPath Condition=" '$(OutputPath)' == '' ">$(SolutionDir)..\bin\$(Platform)\$(Configuration)\$(REVIT_VERSION)</OutputPath>
     <NunitPath Condition=" '$(NunitPath)' == '' ">$(SolutionDir)..\extern\NUnit</NunitPath>
-	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo 0.7</DYNAMOAPI>
+	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOAPI>
 	<DynamoExternPath  Condition=" '$(DynamoExternPath)' == '' ">$(SolutionDir)..\extern</DynamoExternPath>
-	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">C:\Program Files\Dynamo 0.7</DYNAMOTESTAPI>
+	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOTESTAPI>
 	<TestOutputPath Condition=" '$(TestOutputPath)' == '' ">$(SolutionDir)..\bin\$(Platform)\$(Configuration)\$(REVIT_VERSION)</TestOutputPath>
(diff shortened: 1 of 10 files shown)


src/DynamoRevit/DynamoRevit.csproj
```diff
@@ -47,7 +47,12 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="DynamoApplications">
-      <HintPath>$(DYNAMOAPI)\DynamoApplications.dll</HintPath>
+      <HintPath>..\..\..\..\..\..\..\Program Files\Dynamo\Dynamo Core\1.2\DynamoApplications.dll</HintPath>
```
sharadkjaiswal: This change is not required, we still need to use $(DYNAMOAPI) variable to reference any dynamo specific dll

src/DynamoRevit/DynamoRevit.csproj
```diff
@@ -57,8 +62,9 @@
       <HintPath>$(DYNAMOAPI)\DynamoServices.dll</HintPath>
       <Private>False</Private>
     </Reference>
-    <Reference Include="Greg, Version=1.0.5900.25656, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Greg.1.0.5900.25656\lib\net45\Greg.dll</HintPath>
+    <Reference Include="Greg, Version=1.0.5959.26179, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\..\..\..\..\..\Program Files\Dynamo\Dynamo Core\1.2\Greg.dll</HintPath>
```
sharadkjaiswal: Greg.dll needs to be referenced from packages folder because it gets pulled from NuGet package at build time.

sharadkjaiswal: In fact, you don't need to submit any change in this file.

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -735,16 +735,22 @@ public class Views : RevitDropDownBase
...
             
-            //there must always be at least 1 view in a Revit document, so we can exclude the empty list check
+            // there must always be at least 1 view in a Revit document, so we can exclude the empty list check
             foreach (var v in views)
             {
                 Items.Add(new DynamoDropDownItem(v.Name, v));
             }
-            return SelectionState.Restore;
+            Items = Items.OrderBy(x => x.Name).ToObservableCollection();
+            SelectedIndex = 0;
```
sharadkjaiswal: By setting the selected index here you are re-setting the selection index stored in DYN file, so previous selection will no longer be saved. Is there anything specific you wanted to achieve here, otherwise returning SelectionState.Restore was good.

src/Libraries/RevitNodesUI/RevitNodesUI.csproj
```diff
@@ -36,7 +36,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="DynamoUtilities">
-      <HintPath>$(DYNAMOAPI)\DynamoUtilities.dll</HintPath>
+      <HintPath>..\..\..\..\..\..\..\..\Program Files\Dynamo\Dynamo Core\1.2\DynamoUtilities.dll</HintPath>
```
sharadkjaiswal: we don't need to change any of the csproj.

src/Libraries/RevitNodesUI/RevitDropDown.cs
ksobon: yes, when you drop a node on canvas it will have automatically selected the first item like so: 

image

but if selection state is set to restore it will look like this: 

image

src/Libraries/RevitNodesUI/RevitNodesUI.csproj
ksobon: So what happened here is that when I opened the solution i was missing bunch of references to Dynamo dlls. When I re-pointed them to the proper location that of course updated all of the paths in the csproj. Any idea how to make my solution recognize the $(DYNAMOAPI) variable and fix the missing references without manually re-setting them?

src/Libraries/RevitNodesUI/RevitDropDown.cs
ksobon: I removed the `SelectedIndex = 0` and set the `SelectionState.Restore` to be the output.

src/Libraries/RevitNodesUI/RevitNodesUI.csproj
sharadkjaiswal: Once $(DYNAMOAPI) variable is defined correct then all the references will resolve automatically. You need to close the solution and reopen after making changes to csprops file.

src/DynamoRevit/DynamoRevit.csproj
ksobon: I think I restored them all back to what they were. I see now, how you use the $(DYNAMOAPI) variable to set all paths from the user_local.props file. My apologies for messing it up.

src/Libraries/RevitNodesUI/RevitNodesUI.csproj
ksobon: got it. Should be all fixed in latest commit.

src/Config/user_local.props
```diff
@@ -3,9 +3,9 @@
   <PropertyGroup>
     <RevitVersionNumber Condition=" '$(RevitVersionNumber)' == '' ">2017</RevitVersionNumber>
 	<REVIT_VERSION>Revit_$(RevitVersionNumber)</REVIT_VERSION>
-	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">$(SolutionDir)..\..\Dynamo\bin\$(Platform)\$(Configuration)</DYNAMOTESTAPI>
+	<DYNAMOTESTAPI Condition=" '$(DYNAMOTESTAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOTESTAPI>
```
mjkkirschner: this is not correct, there should be no changes to these csproj files at all, you can reset them using git to the previous commit.

then to make sure the build works, pull Dynamo master and build that first.

src/DynamoRevit/DynamoRevit.csproj
```diff
@@ -50,6 +50,11 @@
       <HintPath>$(DYNAMOAPI)\DynamoApplications.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="DynamoCore, Version=1.1.0.2177, Culture=neutral, processorArchitecture=MSIL">
```
mjkkirschner: same here, you dont want any changes in this file.

ksobon:
Oh snap, that i didnt know(that i have to build dynamo first). Let me revisit it again. 

Ps. Can we think of a way to not have to do that? For users that are interested in only building dynamorevit and dont want to mess with dynamo at all. I am thinking of lowering the entry barrier for users to start contributing per Ians comments and goal for next year.

mjkkirschner:
I think a build script used to exist, something that would pull dynamo master, pull dynamo revit master, and then build them in the right order. - not sure what the current state of it is.

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -735,15 +735,20 @@ public class Views : RevitDropDownBase
...
         {
             Items.Clear();
-            //find all views in the project
-            var fec = new FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument);
-            var views = fec.OfClass(typeof(View)).ToElements();
+            // find all views in the project
+            // exclude <RevisionSchedule>
+            var views = new FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument)
+                .OfClass(typeof(View))
+                .Where(x => !x.Name.Contains('<'))
```
mjkkirschner: does this only exclude revision schedule?

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -735,15 +735,20 @@ public class Views : RevitDropDownBase
...
+                .ToList();
             
-            //there must always be at least 1 view in a Revit document, so we can exclude the empty list check
+            // there must always be at least 1 view in a Revit document, so we can exclude the empty list check
             foreach (var v in views)
             {
                 Items.Add(new DynamoDropDownItem(v.Name, v));
             }
+            Items = Items.OrderBy(x => x.Name).ToObservableCollection();
```
mjkkirschner: @sm6srw what do you think of this sorting ? Is this going to cause the issue we talked about the other day?

ksobon:
It should only exclude revisiion schedules. No user created view can have "<" special characters and i have only seen it used for these types of views. Same for default 3d view that uses "{" in is name. Users cant create views named like that.

ksobon:
An alternative would be to first check viewType == schedule and then cast to viewSchedule and check isTitleblockRevisionSchedule but thats expensive. Your call.

src/Config/CS.props
```diff
@@ -10,9 +10,9 @@
 	<REVIT_VERSION>Revit_$(RevitVersionNumber)</REVIT_VERSION>
 	<OutputPath Condition=" '$(OutputPath)' == '' ">$(SolutionDir)..\bin\$(Platform)\$(Configuration)\$(REVIT_VERSION)</OutputPath>
     <NunitPath Condition=" '$(NunitPath)' == '' ">$(SolutionDir)..\extern\NUnit</NunitPath>
-	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo 0.7</DYNAMOAPI>
+	<DYNAMOAPI  Condition=" '$(DYNAMOAPI)' == '' ">C:\Program Files\Dynamo\Dynamo Core\1.2</DYNAMOAPI>
```
sharadkjaiswal: This change is good, except that you need to update the condition to " !Exists('$(DYNAMOAPI)')", then you don't need to change user_local.props

Your current changes to user_local.props will break our automated build system.

ksobon:
@mjkkirschner so I pulled the Dynamo master and trying to build that thing. So far I am missing the two Revit API references and looking at the CS.props file it seem to be setup to look for Revit 2014 which I don't have on my machine. I only have versions 2016/2017. Is it OK to change that reference? Also, why is Dynamo built with Revit dependency? I thought it was a standalone? 

image

ksobon:
OK, you know what? I will start from scratch again, this time in the right order. I just discovered that Dynamo and DynamoRevit need to live in the same directory otherwise they won't build properly. I will close this PR and try again in a bit. Sorry for the trouble.

mjkkirschner:
@ksobon totally fine to keep adding to this PR instead of creating a new one, its up to you. That way the old comments are still associated, but no worries either way.

It looks to me like you are building a super old version of the Dynamo repo.

ksobon:
default branch is "master" that's what I was building.

ksobon:
should I be on RC1.2.0_master?

mjkkirschner:
@ksobon ... neither have any references to the revit api...

mjkkirschner:
also, I'm pretty sure there are is no longer a project called IronPythonEval or UI...

ksobon:
well...this looks like a good start: 
image

ksobon:
and this: 
image

src/Libraries/RevitNodesUI/RevitDropDown.cs
sm6srw: I don't know for this case. I need to see it. I will check it out later today.

PR #1316, opened 2016-09-25
R2017_Sort Views alphabetically

ksobon:
### Purpose

User @andydandy74 has requested here: https://github.com/DynamoDS/DynamoRevit/issues/1264 to return views in an alphabetical order. We already do that in most of the dropdown nodes so It makes sense to grant his wish. 

I also took the opportunity here to remove Revision Schedules from the list since they are mostly useless in Revit while clattering this drop-down list. These views are automatically generated and placed on every Sheet and can really add up fast. 

Before: 

image

After: 

image

Another request from @andydandy74 https://github.com/DynamoDS/Dynamo/issues/941 that we should have ability to set View Scale. This should take care of that: 

image

Another user request https://github.com/DynamoDS/Dynamo/issues/5638 from @nmeek that asked to add ability to retrieve an element Type. It actually comes really handy to have that ability and I have used a node from Clockwork to do so for a long time. 

image

### Declarations

### Reviewers

@mjkkirschner @ikeough 

### FYIs

This would be my second attempt at building this repo from source so again, let me know if I have missed anything. Cheers! 
Also, I don't think I will be submitting small PRs like that. I don't intend to bug the shit out of your guys, but for this first PR I am interested in seeing what the process looks like, the requirements, concerns and standards are so I can adhere to them properly.

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -175,7 +175,7 @@ namespace Revit.Elements
         /// <summary>
         /// Get Element Type
         /// </summary>
-        public Element GetElementType
+        public Element ElementType
         {
             get
             {
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -173,6 +173,21 @@ namespace Revit.Elements
         }
 
         /// <summary>
+        /// Get Element Type
+        /// </summary>
+        public Element GetElementType
+        {
+            get
+            {
+                // Get the active Document
+                Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
+
+                return document.GetElement(this.InternalElement.GetTypeId()).ToDSType(true);
+            }
+
+        }
+
+        /// <summary>
         /// The element id for this element
         /// </summary>
         protected ElementId InternalElementId
--- a/src/Libraries/RevitNodes/Elements/Views/View.cs
+++ b/src/Libraries/RevitNodes/Elements/Views/View.cs
@@ -223,9 +223,12 @@ namespace Revit.Elements.Views
         {
             if (Autodesk.Revit.DB.View.IsValidViewScale(scale))
             {
-                RevitServices.Transactions.TransactionManager.Instance.EnsureInTransaction(Application.Document.Current.InternalDocument);
-                this.InternalView.Scale = scale;
-                RevitServices.Transactions.TransactionManager.Instance.TransactionTaskDone();
+                if (this.InternalView.Scale != scale)
+                {
+                    RevitServices.Transactions.TransactionManager.Instance.EnsureInTransaction(Application.Document.Current.InternalDocument);
+                    this.InternalView.Scale = scale;
+                    RevitServices.Transactions.TransactionManager.Instance.TransactionTaskDone();
+                }
 
                 return this.InternalView.ToDSType(true);
             }
--- a/src/Libraries/RevitNodes/Elements/Views/View.cs
+++ b/src/Libraries/RevitNodes/Elements/Views/View.cs
@@ -162,7 +162,6 @@ namespace Revit.Elements.Views
         /// <summary>
         /// Add Filter to View
         /// </summary>
-        /// <param name="view">View</param>
         /// <param name="parameterFilter">Parameter filter</param>
         public void AddFilter(Revit.Filter.ParameterFilterElement parameterFilter)
         {
@@ -192,7 +191,6 @@ namespace Revit.Elements.Views
         /// <summary>
         /// Set Filter overrides
         /// </summary>
-        /// <param name="view">View</param>
         /// <param name="parameterFilter">Parameter filter</param>
         /// <param name="overrides">overrides settings</param>
         public void SetFilterOverrides(Revit.Filter.ParameterFilterElement parameterFilter, Revit.Filter.OverrideGraphicSettings overrides)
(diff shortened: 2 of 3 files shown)


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -173,6 +173,21 @@ public Category GetCategory
         }
 
         /// <summary>
+        /// Get Element Type
+        /// </summary>
+        public Element GetElementType
```
sharadkjaiswal: No need to use Get in property name. ElementType is sufficient.

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -213,5 +211,30 @@ public Revit.Filter.OverrideGraphicSettings FilterOverrides(Revit.Filter.Paramet
...
+        #region Scale
+
+        /// <summary>
+        /// Set View Scale
+        /// </summary>
+        /// <param name="scale">View scale is the ration of true model size to paper size.</param>
+        /// <returns name="View">View</returns>
+        public Element SetScale(int scale=100)
+        {
+            if (Autodesk.Revit.DB.View.IsValidViewScale(scale))
```
sharadkjaiswal: you might want to return if the input scale is same as current scale on view, to avoid unnecessary document update.

ksobon: done.

sharadkjaiswal:
@ikeough, @mjkkirschner  do we need to cherry-pick this to other branches?

kronz:
@sharadkjaiswal please cherry pick to 2018 and 2016
@ksobon thanks so much for the contribution!

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -173,6 +173,21 @@ public Category GetCategory
...
+        /// Get Element Type
+        /// </summary>
+        public Element ElementType
+        {
+            get
+            {
+                // Get the active Document
+                Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
+
+                return document.GetElement(this.InternalElement.GetTypeId()).ToDSType(true);
```
ikeough: According to the Revit API docs, `GetTypeId()` returns:
>The id of the element's type, or invalid element id if the element cannot have type assigned. 

That means that `GetElement(...)` will fail if an invalid element id is returned. Can we add a check first like:
```c#
var typeId = this.InternalElement.GetTypeId();
if(typeId = ElementId.InvalidElementId){
  return null;
}else{
  return document.GetElement(typeId).toDSType(true);
}
```

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -735,15 +735,21 @@ public class Views : RevitDropDownBase
...
             Items.Clear();
+
             //find all views in the project
-            var fec = new FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument);
-            var views = fec.OfClass(typeof(View)).ToElements();
+            //exclude <RevisionSchedule> (revision tables on sheets) from list
+            var views = new FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument)
+                .OfClass(typeof(View))
+                .Where(x => !x.Name.Contains('<'))
+                .ToList();
```
ksobon: Yeah by default it returns `Element` collection but also i removed it because I don't think I can call that method after I have used a Linq method so I used the `ToList()` instead.

src/Libraries/RevitNodes/Elements/Element.cs
ikeough: Also, we should update the comments on this property to explain that it might return null (if that's what we decide to return). Otherwise, this will return null and the user won't have any way of knowing why.

ikeough: We should update the description of this property to describe that it could return null. Something like:
> Returns the FamilyType for this Element. Returns null if the Element cannot have a FamilyType assigned.

ksobon: you are correct, although the error would be in `ToDSType()` method getting called on `null` value since `GetElement(ElementId.InvalidElementId)` would just return 'null'. Should that method be able to better handle `null` values?

ksobon: How do you want to do the `FamilyType` return? You can't cast `Element` to `FamilyType` and `Element` is what `GetElement()` returns. You also can't use the static method `FamilyType.FromExisting()` because again, we have an `Element` not the `FamilyType`. `ToDSType()` actually returns a `FamilyType` because it uses a dynamic variable. Should we implement that here?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -173,6 +173,21 @@ public Category GetCategory
...
 
         /// <summary>
+        /// Get Element Type
+        /// </summary>
+        public Element ElementType
+        {
+            get
+            {
+                // Get the active Document
+                Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
```
ksobon: fixed

src/Libraries/RevitNodes/Elements/Element.cs
ksobon: image

It returns the right thing but the port name is not correct.

ksobon: this works: 
image

ikeough: I think we can do:
```c#
return (FamilyType)(document.GetElement(typeId).toDSType(true));
```

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -213,5 +211,33 @@ public Revit.Filter.OverrideGraphicSettings FilterOverrides(Revit.Filter.Paramet
...
+            {
+                if (this.InternalView.Scale != scale)
```
mjkkirschner: is scale a double? if so use an epsilon value tolerance value to compare them please.

mjkkirschner:
@ksobon I think this is a good change, but! I'm curious what will happen with other localizations of Revit for views with builtin names, @sm6srw could that be an issue once these are sorted alphabetically?

mjkkirschner:
@ksobon lets get this in, I will generate cherry picks to 2017 and 2018 branches.

mjkkirschner:
@ksobon - you wanted me to hold off merging this PR?

ksobon:
I think this is all contained in my other one, so you can probably just merge the other one and be done with it. Should I just close this?

PR #1341, opened 2016-10-10
Family instance by geo 2016

moethu:
### Purpose

Create FamilyType By Geometry
### Declarations

### Reviewers

@mjkkirschner 
### FYIs


src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -60,6 +65,29 @@ private void InitFamilySymbol(Autodesk.Revit.DB.FamilySymbol symbol)
             InternalSetFamilySymbol(symbol);
         }
 
+        /// <summary>
+        /// Get Solid from Element helper method
+        /// </summary>
+        /// <param name="element"></param>
+        /// <returns></returns>
+        private static Solid GetSolidFromElement(Autodesk.Revit.DB.Element element)
```
Randy-Ma: There is a property method called Element.Solids. Is there any specific reason that we need another method for ElementType?

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -221,6 +249,145 @@ public static FamilyType ByName(string name)
             };
         }
 
+        /// <summary>
+        /// Create new Family Instance from Geometry (via SAT import)
```
Randy-Ma: Uhmmmm,  from the method, it looks like that it is to create a FamilyType object. But the comment and the pull request title is saying to create family instance. 

Also if the parameter if geometry, then it is not only via SAT import.

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -252,4 +419,23 @@ public override string ToString()
...
+
+        public bool OnFamilyFound(bool familyInUse, out bool overwriteParameterValues)
+        {
+            overwriteParameterValues = true;
+            return true;
+        }
+
+        public bool OnSharedFamilyFound(Autodesk.Revit.DB.Family sharedFamily, bool familyInUse, out FamilySource source, out bool overwriteParameterValues)
+        {
+            source = FamilySource.Project;
```
Randy-Ma: If this is the case, shall an warning/error message be displayed so that users know that the geometries they passed in is not used at all.

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -221,6 +249,145 @@ public static FamilyType ByName(string name)
...
+        /// <param name="category"></param>
+        /// <param name="templatePath"></param>
+        /// <param name="isVoid"></param>
+        /// <param name="material"></param>
+        /// <param name="subcategory"></param>
+        /// <returns></returns>
+        public static FamilyType ByGeometry(Autodesk.DesignScript.Geometry.Geometry geometry, string name, Category category, string templatePath, Material material, bool isVoid = false, string subcategory = "")
+        {
+            Autodesk.Revit.DB.Document document = DocumentManager.Instance.CurrentDBDocument;
+            TransactionManager.Instance.ForceCloseTransaction();
```
Randy-Ma: Now I understand that it is because a new transaction need to be started for another db document. But it is still not so good to close the current transaction which is not done in any other nodes.

src/Libraries/RevitNodes/Elements/FamilyType.cs
Randy-Ma: For the first point about creating FamilyType objects, now I understand from @kronz that the new requirement is to create FamilyType objects. But the comment still need to be updated.

src/Libraries/RevitNodes/Elements/FamilyType.cs
moethu: could be, as long as compute references is set to true for elements as well

src/Libraries/RevitNodes/Elements/FamilyType.cs
moethu: @Randy-Ma yes, updated it. its not limited to Sat geometry the comment means via SAT import - meaning as an info that its exporting the geometry to SAT and importing it into the new family

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -60,6 +65,29 @@ private void InitFamilySymbol(Autodesk.Revit.DB.FamilySymbol symbol)
...
+            GeometryElement geo = element.get_Geometry(new Options() { ComputeReferences = true });
+
+            // Get geometry instance from geometry element
+            var enumerator = geo.GetEnumerator();
+            enumerator.MoveNext();
+            GeometryInstance i = (GeometryInstance)enumerator.Current;
+
+            // Get solid from geometry instance
+            var geom2 = i.GetInstanceGeometry();
+            var enumarator2 = geom2.GetEnumerator();
```
moethu: yes, only required the the first geometryinstance & solid at this point because all of the imported should be in there.

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -221,6 +249,145 @@ public static FamilyType ByName(string name)
...
         }
 
+        /// <summary>
+        /// Create new Family Instance from Geometry (via SAT import)
+        /// </summary>
+        /// <param name="geometry"></param>
+        /// <param name="name"></param>
+        /// <param name="category"></param>
+        /// <param name="templatePath"></param>
+        /// <param name="isVoid"></param>
```
moethu: Do you think we should have two methods? CreateVoidByGeo and CreateSolid ? sure @kronz would prefer that?

src/Libraries/RevitNodes/Elements/FamilyType.cs
moethu: sure that the TransactionManager.Instance handles two transactions at the same time? I understood it as a singleton

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -221,6 +249,145 @@ public static FamilyType ByName(string name)
...
+
+            // Get the conversion factor based on the display units
+            DisplayUnitType units = document.GetUnits().GetFormatOptions(UnitType.UT_Length).DisplayUnits;
+            double factor = UnitUtils.ConvertToInternalUnits(1, units);
+
+            // scale the incoming geometry
+            geometry = geometry.Scale(factor);
+
+            // get a displacement vector
+            Vector vector = Vector.ByTwoPoints(Autodesk.DesignScript.Geometry.BoundingBox.ByGeometry(geometry).MinPoint, Autodesk.DesignScript.Geometry.Point.Origin());
```
moethu: the min point sofar created the correct offset

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -221,6 +249,145 @@ public static FamilyType ByName(string name)
...
+
+            // translate the geometry to origin
+            geometry = geometry.Translate(vector);
```
moethu: Do you mean  Robustify(ref newGeometries, ref translation); ?

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -221,6 +249,145 @@ public static FamilyType ByName(string name)
...
+                var cutWithVoidsParameter = freeFormElement.get_Parameter(BuiltInParameter.FAMILY_ALLOW_CUT_WITH_VOIDS);
+                if (cutWithVoidsParameter != null && !cutWithVoidsParameter.IsReadOnly)
+                {
+                    Revit.Elements.InternalUtilities.ElementUtils.SetParameterValue(cutWithVoidsParameter, 1);
+                }
+            }
+            else
+            {
+                // Apply material if supplied
+                if (material != null)
```
moethu: yes but its still in the family document - should it always be applied?

src/Libraries/RevitNodes/Elements/FamilyType.cs
Randy-Ma: Have you tried a case that there are two separate solids for this element type?

src/Libraries/RevitNodes/Elements/FamilyType.cs
Randy-Ma: I think so. It should be consistent instead of applying for one case but not for other cases.

src/Libraries/RevitNodes/Elements/FamilyType.cs
Randy-Ma: No, my point is that if users have a list of geometries and want to convert them to a list of FamilyType objects, the program should automatically figure out which ones are void and which ones are not void.

src/Libraries/RevitNodes/Elements/FamilyType.cs
Randy-Ma: Great! Thanks!

src/Libraries/RevitNodes/Elements/FamilyType.cs
Randy-Ma: The wrapper is a singleton. Right now there is an undocumented rule that everything should be completed in one transaction.  If you insist to implement in this way, at least all Revit test cases need to be verified to pass.

src/Libraries/RevitNodes/Elements/FamilyType.cs
Randy-Ma: The main point is that all the new geometries are disposed explicitly.

Randy-Ma:
@moethu 
Thanks a ton! Amongst the comments, there are two which are especially important:
- temporary geometries need to be disposed explicitly, if not, it will cause random crash
- the conversion from geometry solid to Revit solid is very creative but I can not say that I am totally ok with it because:
    a). multiple transactions are used;
    b). sometimes only the 1st element from a collections is taken;
For the second point, it is better at least to isolate the conversion function as a separate class so that it can be tested and more unit/integration test cases are required to ensure that it is robust.

src/Libraries/RevitNodes/Elements/FamilyType.cs
```diff
@@ -221,6 +249,145 @@ public static FamilyType ByName(string name)
...
+                    view = v;
+                }
+            }
+
+            // Import the sat file to origin in feet
+            TransactionManager.Instance.EnsureInTransaction(familyDocument);
+            ElementId importedElementId = familyDocument.Import(tempFile, new SATImportOptions() { Placement = ImportPlacement.Origin, Unit = ImportUnit.Foot }, view);
+
+            // get the solid element from the imported sat file
+            Solid solid = GetSolidFromElement(familyDocument.GetElement(importedElementId));
```
moethu: Nice Idea, would make the implementation much cleaner. Will Move it over there.

src/Libraries/RevitNodes/Elements/FamilyType.cs
moethu: I will change this to support all elements

moethu:
@Randy-Ma please review latest commit -

PR #1342, opened 2016-10-10
Performance adviser 2016

moethu:
### Purpose

Performance Adviser exposed
### Declarations

### Reviewers

@mjkkirschner 
### FYIs


src/Libraries/RevitNodes/Elements/PerformanceAdviserRule.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            get
+            {
+                PerformanceAdviser adviser = PerformanceAdviser.GetPerformanceAdviser();
+                return adviser.GetRuleDescription(this.InternalId);
+            }
+        }
+
+        /// <summary>
+        /// Check is the rule is activated
```
QilongTang: Since the property is a boolean, can you change the comment to something like: A boolean indicating if the rule is actived. You can move the current comment above the **return**

src/Libraries/RevitNodes/Elements/PerformanceAdviserRule.cs
```diff
@@ -0,0 +1,134 @@
...
+            get
+            {
+                PerformanceAdviser adviser = PerformanceAdviser.GetPerformanceAdviser();
+                return adviser.IsRuleEnabled(this.InternalId);
+            }
+        }
+
+        /// <summary>
+        /// Activate or deactivate the rule
+        /// </summary>
```
QilongTang: Same thing here, the current property comment fit more above the **adviser.SetRuleEnabled(this.InternalId, value);**

src/Libraries/RevitNodes/Elements/FailureMessage.cs
```diff
@@ -0,0 +1,104 @@
...
+using RevitServices.Transactions;
+using RevitServices.Persistence;
+using System.Collections.Generic;
+using System.Collections;
+
+
+namespace Revit.Elements
+{
+    /// <summary>
+    /// Performance Adviser Failure Message
```
QilongTang: Class name indicating it's for general purpose but comments seem to indicate it's only used for performance adviser, can we align these two?

src/Libraries/RevitNodes/Elements/PerformanceAdviserRule.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <param name="guid"></param>
+        /// <returns></returns>
+        public static PerformanceAdviserRule ById(string guid)
+        {
+            return new PerformanceAdviserRule(new PerformanceAdviserRuleId(new Guid(guid)));
+        }
+
+        [SupressImportIntoVM]
+        public override string ToString()
```
QilongTang: Some comments what it output would be helpful.

PR #1345, opened 2016-10-10
Floor revit2016

moethu:
### Purpose

Floor and Roof Shape Points
### Declarations

### Reviewers

@mjkkirschner 
### FYIs


Randy-Ma:
Generally it is better to have some integration test cases.

src/Libraries/RevitNodes/Elements/Floor.cs
```diff
@@ -161,6 +165,81 @@ public static Floor ByOutlineTypeAndLevel(PolyCurve outline, FloorType floorType
...
+            TransactionManager.Instance.EnsureInTransaction(DocumentManager.Instance.CurrentDBDocument);
+            this.InternalFloor.SlabShapeEditor.Enable();
+            this.InternalFloor.SlabShapeEditor.DrawPoint(point.ToXyz());
+            TransactionManager.Instance.TransactionTaskDone();
+        }
+
+        /// <summary>
+        /// Move existing point by offset
+        /// </summary>
+        public void MovePoint(Pt point, double offset)
```
Randy-Ma: It is good to return a floor.

src/Libraries/RevitNodes/Elements/Roof.cs
```diff
@@ -0,0 +1,289 @@
...
+                    points.Add(v.Position.ToPoint());
+                }
+                return points;
+            }
+        }
+
+        /// <summary>
+        /// Add Point to Slab Shape
+        /// </summary>
+        public void AddPoint(Pt point)
```
Randy-Ma: Similarly it is better to return a roof for this method and the MovePoint method.

src/Libraries/RevitNodes/Elements/Floor.cs
```diff
@@ -161,6 +165,81 @@ public static Floor ByOutlineTypeAndLevel(PolyCurve outline, FloorType floorType
...
+                    points.Add(v.Position.ToPoint());
+                }
+                return points;
+            }
+        }
+
+        /// <summary>
+        /// Add Point to Slab Shape
+        /// </summary>
+        public void AddPoint(Pt point)
```
moethu: @Randy-Ma but this would act differently since the slab shape editor affects the original element. The floor you used as an input is modified afterwards. I could still return it though.

src/Libraries/RevitNodes/Elements/RoofType.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        #region Internal static constructors
+
+        /// <summary>
+        /// Create a RoofType from a user selected Element.
+        /// </summary>
+        /// <param name="RoofType"></param>
+        /// <param name="isRevitOwned"></param>
+        /// <returns></returns>
+        internal static RoofType FromExisting(Autodesk.Revit.DB.RoofType RoofType, bool isRevitOwned)
```
moethu: @Randy-Ma sorry, can you specify this? RoofType is an element wrapper itself.

moethu:
@Randy-Ma incorporated your changes

src/Libraries/RevitNodes/Elements/RoofType.cs
Randy-Ma: You can refer other wrappers there for FamilyType, FloorType, ...

src/Libraries/RevitNodes/Elements/Floor.cs
Randy-Ma: In that case, it is good to return the original modified element.

src/Libraries/RevitNodes/Elements/Floor.cs
```diff
@@ -196,34 +196,39 @@ public IEnumerable<Pt> Points
         }
 
         /// <summary>
-        /// Add Point to Slab Shape
+        /// Add a new point to the slab shape editor.
+        /// Behaves the same as adding a point manually to a slab in
+        /// Revit's slab shape edit mode.
         /// </summary>
-        public void AddPoint(Pt point)
+        public static Floor AddPoint(Floor floor, Pt point)
```
Randy-Ma: Sorry if I am not making it very clear. But you can refer FamilyInstance.SetRotation. You do not need to pass in a floor as a parameter.

Issue #1334, opened 2016-09-30
Graphics Overrides Node not implemented fully


ksobon:
## Dynamo version

Dynamo 1.2
## Revit version

2017
## Operating system

Windows 7

I was trying to create a new View Filter, add it to a View and then override some Graphics for that filter. However, Graphics Override node is not fully implemented, in my opinion. Some of the inputs are coded to accept Unwrapped `Autodesk.Revit.DB` elements like `FillPatternElement` or `LinePatternElement`. That's not exactly possible to use in current Dynamo context. First we don't have access to `InternalElement` from CodeBlock level. You could try and unwrap an element from Python node, but the default implementation is such that it will always try and wrap the element back when its passed OUT of the node so we always get this: 

![image](https://cloud.githubusercontent.com/assets/529781/19005159/09851390-8727-11e6-9403-baba18fa99ca.png)

Even when code says this: 

```
import clr

# Import Element wrapper extension methods
clr.AddReference("RevitNodes")
import Revit
clr.ImportExtensions(Revit.Elements)

#The inputs to this node will be stored as a list in the IN variables.
dataEnteringNode = IN

#Assign your output to the OUT variable.
OUT = UnwrapElement(IN[0])
```

Now, that makes it quite impossible to get a bare Revit API element to use here, hence rendering that node impossible to use. It should be updated to accept input type `Revit.Elements.Element` since its quite easy to collect Fill Pattern elements from the DB and wrap them up. Here's an example of my personal implementation that gets the job done: 

![image](https://cloud.githubusercontent.com/assets/529781/19005329/fae6faf0-8727-11e6-99bd-18849ae0baa8.png)

Unfortunately at the time that I was writing it, `Revit.Elements.OverrideGraphicsSettings` element didn't exist so mine actually passes out a bare Revit API element, but that would be really simple to change.

PR #1378, opened 2016-10-31
fixes to view schedule + export functionality

ksobon:
### Purpose

These are two fixes to View Schedule:

1. ViewSchedule.AddField node returns ViewSchedule object for every field it added. Fix to that was to change input type on fields from single to a List input. 
2. ViewSchedule.Create node will not rebind to previously created View, but instead try and create a new one causing an error. I added a binding routine to it, that fixed that issue. 

New Functionality includes classes and objects related to Exporting View Schedules to textual format. It also adds two new overloads for creating regular schedules as well as material takeoffs. 

New functionality for overriding view visibility by Category: 

image

Uses existing Graphics override settings created by @moethu for the View Filters. I only thought that it would be a good idea to extend them a little bit. 

image

Now, there were Transparency, Detail Level and Halftone settings missing from the Override Graphics before. Those are quite critical when working with Categories. The only catch with these is that one cannot apply Detail Level i think to a View Filter. View Filters still use the same Override class but they don't have that particular setting exposed. 

Last small fix was to add Visibility setting to View.SetFilterOverrides node similar and consistent with View.SetCategoryOverrides node. This allows for turning on/off visibility for a given View Filter. 

Also, this same settings can be used for overriding elements in view. There was a node called OverrideColorInView which was quite poorly implemented. There is no reason not to just re-write it to use the current override graphics class: 

image

Fixed this issue as well: https://github.com/DynamoDS/DynamoRevit/issues/1334

image

Added drop-downs for Fill Patterns and Line Patterns.

Fix to this issue: https://github.com/DynamoDS/Dynamo/issues/1131
Fix to this issue: https://github.com/DynamoDS/Dynamo/issues/2081

### Declarations

### Reviewers

@ikeough @mjkkirschner 

### FYI

@mjkkirschner that's the new PR that we talked about last week.


src/AssemblySharedInfoGenerator/AssemblySharedInfo.tt
```diff
@@ -69,6 +69,6 @@ using System.Runtime.InteropServices;
 <#+
 int MajorVersion = 1;
 int MinorVersion = 2;
-int BuildNumber = 1;
+int BuildNumber = 2;
```
mjkkirschner: lets not mess with this number here on master branch, drop this for now.

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
```diff
@@ -156,34 +171,87 @@ public static ScheduleView CreateSchedule(Category category, string name, string
...
+
+        /// <summary>
+        ///     Export View Schedule to CSV, TSV etc.
+        /// </summary>
+        /// <param name="path">A valid file path with file extension.</param>
+        /// <param name="exportOptions">Export Options. If null, default will be used.</param>
+        /// <returns></returns>
+        public ScheduleView Export(
+            string path, 
+            [DefaultArgumentAttribute("null")] ScheduleExportOptions exportOptions)
```
mjkkirschner: I really don't like this use of null, something we did with @moethu's prs where he needed similar functionality was to define a static constructor for a default token:

https://github.com/DynamoDS/DynamoRevit/blob/7270187e8688fbd8ec538b4617810dccdb9a695f/src/Libraries/RevitNodes/Elements/TextNoteType.cs#L106

then create one of those as your optional argument:
https://github.com/DynamoDS/DynamoRevit/blob/7270187e8688fbd8ec538b4617810dccdb9a695f/src/Libraries/RevitNodes/Elements/TextNote.cs#L206

so you would build a default constructor for export options, it could be internal or hidden from the node library.

mjkkirschner: let me also clarify why I don't "like" it.

* It's not explicit what is going to happen to a user - and it's a bit confusing to a developer as well.
* You have redefine this behavior potentially in multiple export nodes... lets say we had 2 for example.
* what happens when another node upstream fails and passes null, this node now works, but finding the source of the error becomes more difficult... I would rather have immediate failure in a graph and understand what broke then have magical things happen that make debugging harder than it already is.
just my thoughts. :)

ksobon: I want to leave this input empty/null. I can have a utility method that I call that returns a null value like this: `[DefaultArgument("archilab.View.GetNull()")] Color projectionLineColor` and a method: 

```
public static object GetNull()
{
    return null;
}
```

Does that work?

mjkkirschner: it's not how you get the null, that I'm opposed to. It's that I don't think the default value should ever be null.

mjkkirschner: @ikeough @kronz  any opinion on this type of design intention for graphs and how it effects nulls propagating the graph?

src/Libraries/RevitNodes/Schedules/ScheduleExportOptions.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        ///     Creates View Schedule Export Options element.
+        /// </summary>
+        /// <param name="columnHeaders">How to export column headers. Default is MultipleRows.</param>
+        /// <param name="fieldDelimiter">How to delimit fields. Default is Tab.</param>
+        /// <param name="headersFootersBlanks">Whether to export group headers, footers, and blank lines. Default is false.</param>
+        /// <param name="textQualifier">How to qualify text fields. Default is DoubleQuote.</param>
+        /// <param name="title">Whether or not to export the schedule title. Default is false.</param>
+        /// <returns name="exportOptions">View Schedule Export Options.</returns>
+        public static ScheduleExportOptions ByProperties(
```
mjkkirschner: thank you for the tooltips!

src/Libraries/RevitNodes/Properties/Resources.resx
```diff
@@ -369,4 +369,7 @@
   <data name="ScheduleTypeArgumentException" xml:space="preserve">
     <value>Invalid Schedule Type.</value>
   </data>
+  <data name="ScheduleExportError" xml:space="preserve">
```
mjkkirschner: en-us, and this resx file don't seem to be the same, is this one missing the  `ScheduleExportError`?

src/Libraries/RevitNodesUI/Properties/Resources.resx
```diff
@@ -349,4 +349,10 @@
   <data name="FilterTypeSelectorDescription" xml:space="preserve">
     <value>Filter Types for Revit Filters</value>
   </data>
+  <data name="ExportColumnHeadersDescription" xml:space="preserve">
```
mjkkirschner: maybe it is here by mistake?

ksobon: ExportColumnHeadersDescription is only needed for UI dropdown nodes so it should not be in the Revit Nodes resex. Right?

src/Libraries/RevitNodes/Properties/Resources.resx
ksobon: I might have deleted it from one but not the other. Let me check.

src/Libraries/RevitNodesUI/Properties/Resources.resx
ksobon: fixed.

src/Libraries/RevitNodes/Properties/Resources.resx
ksobon: fixed

mjkkirschner:
Hi @ksobon please don't bundle more and more functionality into these PRs, they become hard to review, more difficult to reason about, revert, test etc. I know the functionality is related, but thats not a good enough reason to make the PR this large especially because this PR fixes a bug in your previous PR, merging the fix independently of other features and changes is good practice. I will review this PR and we can get these changes in together but going forward please try to limit them in size. You can use branches which branch multiple times for this purpose.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -344,27 +344,20 @@ public object GetParameterValueByName(string parameterName)
         }
 
         /// <summary>
-        /// Override the element's color in the active view.
+        ///     Override Elements Graphics Settings in Active View.
         /// </summary>
-        /// <param name="color">The color to apply to a solid fill on the element.</param>
-        public Element OverrideColorInView(Color color)
```
mjkkirschner: please don't remove this public function. For now, I would like to keep both versions, removing a public node method will break old graphs, and I'd rather not figure out how to migrate this node in this PR.

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
```diff
@@ -76,7 +77,13 @@ private void InitScheduleView(Category category, string name, ScheduleType type)
             var doc = DocumentManager.Instance.CurrentDBDocument;
             TransactionManager.Instance.EnsureInTransaction(doc);
 
-            var vs = CreateViewSchedule(category, name, type);
+            // Get existing view if possible
+            var vs = ElementBinder.GetElementFromTrace<Autodesk.Revit.DB.ViewSchedule>(doc);
```
mjkkirschner: does this fix the reload bug you found?

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
```diff
@@ -156,34 +171,78 @@ public static ScheduleView CreateSchedule(Category category, string name, string
...
+        /// <param name="fields"></param>
         /// <returns name="View">View Schedule</returns>
-        public ScheduleView RemoveField(Revit.Schedules.ScheduleField field)
+        public ScheduleView RemoveField(List<Revit.Schedules.ScheduleField> fields)
         {
-            if (this.Fields.Any(x => x.Name == field.Name))
+            var doc = DocumentManager.Instance.CurrentDBDocument;
+            TransactionManager.Instance.EnsureInTransaction(doc);
+
+            foreach (Revit.Schedules.ScheduleField field in fields)
```
mjkkirschner: `var field in fields` if possible

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
```diff
@@ -156,34 +171,78 @@ public static ScheduleView CreateSchedule(Category category, string name, string
...
             {
-                var doc = DocumentManager.Instance.CurrentDBDocument;
-                TransactionManager.Instance.EnsureInTransaction(doc);
-                this.InternalViewSchedule.Definition.RemoveField(field.InternalScheduleField.FieldId);
-                TransactionManager.Instance.TransactionTaskDone();
+                if (this.Fields.Any(x => x.Name == field.Name))
+                {
+                    this.InternalViewSchedule.Definition.RemoveField(field.InternalScheduleField.FieldId);
```
mjkkirschner: could this fail?, should this be wrapped in a try catch?

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -7,13 +7,12 @@
 namespace Revit.Elements.Views
 {
     /// <summary>
-    /// An abstract Revit View - All view types inherit from this type
+    ///     An abstract Revit View - All view types inherit from this type
     /// </summary>
-    //[SupressImportIntoVM]
```
mjkkirschner: should this really be removed? What doesn't work if this type is not imported?

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -200,16 +199,18 @@ public IEnumerable<Revit.Filter.ParameterFilterElement> Filters
...
         ///     disabling or modifying the View Template.
         /// </summary>
-        /// <param name="parameterFilter">Parameter filter</param>
-        /// <param name="overrides">overrides settings</param>
+        /// <param name="parameterFilter">Parameter Filter</param>
+        /// <param name="overrides">Graphic Overrides Settings</param>
+        /// <param name="hide">If True given Filter will be hidden.</param>
         /// <returns name="view">View</returns>
-        public Element SetFilterOverrides(Revit.Filter.ParameterFilterElement parameterFilter, Revit.Filter.OverrideGraphicSettings overrides)
+        public Revit.Elements.Views.View SetFilterOverrides(Revit.Filter.ParameterFilterElement parameterFilter, Revit.Filter.OverrideGraphicSettings overrides, bool hide = false)
```
mjkkirschner: same here

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -224,14 +225,54 @@ public Revit.Filter.OverrideGraphicSettings FilterOverrides(Revit.Filter.Paramet
...
+
         #region Scale
 
         /// <summary>
         ///     Set View Scale
         /// </summary>
         /// <param name="scale">View scale is the ration of true model size to paper size.</param>
         /// <returns name="view">View</returns>
-        public Element SetScale(int scale=100)
+        public Revit.Elements.Views.View SetScale(int scale=100)
```
mjkkirschner: same here

src/Libraries/RevitNodes/Elements/Views/View3D.cs
```diff
@@ -1,16 +1,13 @@
...
 using System.Linq;
-
 using Autodesk.DesignScript.Runtime;
 using Autodesk.Revit.DB;
 using RevitServices.Persistence;
 using RevitServices.Transactions;
 
 namespace Revit.Elements.Views
 {
-    //[SupressImportIntoVM]
```
mjkkirschner: same question, is this required?

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
```diff
@@ -1,32 +1,26 @@
...
+    ///     Override Graphic Settings
     /// </summary>
     public class OverrideGraphicSettings
     {
 
         #region Internal Properties
 
         /// <summary>
-        /// Internal reference to the Revit Element
+        ///     Internal reference to the Revit Element
```
mjkkirschner: formatting changes, try to get rid of these.

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
```diff
@@ -50,18 +44,24 @@ internal OverrideGraphicSettings(Autodesk.Revit.DB.OverrideGraphicSettings inter
...
         public static OverrideGraphicSettings ByProperties(
             [DefaultArgumentAttribute("null")]DSCore.Color cutFillColor,
             [DefaultArgumentAttribute("null")]DSCore.Color projectionFillColor,
             [DefaultArgumentAttribute("null")]DSCore.Color cutLineColor,
             [DefaultArgumentAttribute("null")]DSCore.Color projectionLineColor,
-            [DefaultArgumentAttribute("null")]FillPatternElement cutFillPattern,
-            [DefaultArgumentAttribute("null")]FillPatternElement projectionFillPattern,
-            [DefaultArgumentAttribute("null")]LinePatternElement cutLinePattern,
-            [DefaultArgumentAttribute("null")]LinePatternElement projectionLinePattern,
+            [DefaultArgumentAttribute("null")]Revit.Elements.Element cutFillPattern,
```
mjkkirschner: why this change?

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
```diff
@@ -50,18 +44,24 @@ internal OverrideGraphicSettings(Autodesk.Revit.DB.OverrideGraphicSettings inter
         /// <param name="projectionFillPattern">Projection fill pattern</param>
         /// <param name="cutLinePattern">Cut line pattern</param>
         /// <param name="projectionLinePattern">Projection line pattern</param>
-        /// <returns>OverrideGraphicSettings</returns>
+        /// <param name="transparency">Transparency as integer between 1-100.</param>
+        /// <param name="detailLevel">Detail Level setting, ex: Coarse, Fine etc.</param>
+        /// <param name="halftone">Halftone. True = halftone.</param>
+        /// <returns name="overrides">Override Graphic Settings</returns>
         public static OverrideGraphicSettings ByProperties(
```
mjkkirschner: hmm, again, concern about node resolution opening old graphs. @kronz is it acceptable to break Revit graphs at this time?

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -478,7 +478,10 @@ protected override SelectionState PopulateItemsCore(string currentSelection)
                 if (category != null)
                 {
                     string name = getFullName(category);
-                    Items.Add(new DynamoDropDownItem(name, categoryId));
+                    if (name != "Cameras")
```
mjkkirschner: hmm... I don't think this will work in all versions of Revit... @sm6srw what do you think?

mjkkirschner: also, why do this?

mjkkirschner: I think just because you can't do anything with them today is not a particularly great reason to hide them, but I have no insight one way or the other. - wonder what @sm6srw thinks.

mjkkirschner:
@ksobon I have some questions, especially with the public methods and signatures that you have modified. I need to check with @kronz if this is okay, otherwise we will need to keep both your improved versions and the originals, or figure out how to migrate... not always possible.

Also, you will need to add tests for the override graphics stuff and the changes to categories dropdown if possible.

src/Libraries/RevitNodes/Elements/Element.cs
ksobon: mhmmmm, I thought that the next release was going to be Dynamo 2.0, and API breaking changes were allowed....actually, just kidding. I had no clue that you were worried about this. I can bring that old clunker back so long as there is room for this Corvette. :-P

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
ksobon: yeah, that binds to the view so that when re-opening a graph it doesn't try to create a new one again.

src/Libraries/RevitNodes/Elements/Element.cs
mjkkirschner: @ksobon haha 😛 , don't invest the time just yet, let me verify with @kronz.

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -7,13 +7,12 @@
 namespace Revit.Elements.Views
 {
     /// <summary>
-    /// An abstract Revit View - All view types inherit from this type
+    ///     An abstract Revit View - All view types inherit from this type
```
ksobon: ahhhh...personal touch not allowed? :-( ...i will get rid of it.

src/Libraries/RevitNodes/Elements/Views/View.cs
ksobon: it was commented out, I just removed that line as per @ikeough earlier comments to not submit commented out code into production branch.

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -166,15 +165,15 @@ public override string ToString()
         /// </summary>
         /// <param name="parameterFilter">Parameter filter</param>
         /// <returns name="view">View</returns>
-        public Element AddFilter(Revit.Filter.ParameterFilterElement parameterFilter)
+        public Revit.Elements.Views.View AddFilter(Revit.Filter.ParameterFilterElement parameterFilter)
```
ksobon: nothing should happen except that you are avoiding a cast `ToDSType()` when you just return the `View` object which this node works with to begin with. Why the extra cast?

src/Libraries/RevitNodes/Elements/Views/View.cs
ksobon: please see above. You are basically skipping the `ToDSType()` method which takes the `View` object and casts it back into the `View`...its redundant.

src/Libraries/RevitNodes/Elements/Views/View3D.cs
ksobon: it was commented out and I just deleted that comment. as for the using statement, it was not used anywhere in the code, why import it?

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
ksobon: done

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
ksobon: This node WAS NOT WORKING to begin with. I stated this before. There is no way for user to obtain unwrapped `Autodesk.Revit.DB.FillPatternElement` which is what this input was declared as. All Python based nodes automatically wrap elements and there is no access to this from a CodeBlock...basically it was a dead end. You could not provide these inputs for this node. By changing the type to `Revit.Elements.Element` it allows you to query Fill Patterns using Python or the DropDown nodes that I added in this PR.

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
ksobon: because unwrapper Revit API elements are not accessible in Dynamo...not for everyday users at least.

src/Libraries/RevitNodesUI/RevitDropDown.cs
ksobon: I might be wrong on this, but I can't for love of god find an object that you could assign OST_Cameras category to. I thought things like Security Devices would be allowed to have that category but its not available. Perhaps something like a NodeView or a CameraInfo object is of that category...I don't know. I got misled by Marcello trying to use that category to collect "Camera Views" which in reality are just View3D objects but that's not a great reason to hide it. I will reverse these changes.

ksobon:
There is already a test for that node and it has nothing to do with the actual implementation of the node, hence you guys didn't catch the fact that it was impossible to make it work. :-) 

image

A OverrideGraphicsSetting object with all nulls is a valid object that is used to reset all overrides to "by category" state.

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
moethu: @ksobon @mjkkirschner agree, there was no option to actually get an element of type FillPatternElement - I would rather suggest to add FillPatternElements to Dynamo instead of allowing any kind of element, because currently they could even plug in a wall element.

ksobon: I agree its a good idea to add Fill Pattern Element to Dynamo. That will make it easier for people. 
Now, I don't agree with getting rid of `null` input. In this particular case I think it makes sense because it actually has a use and is a valid input. It resets the override to "by category". Also, a null used in this node will never escape into the wider context of the definition so I would not worry about it corrupting the graph.

src/Libraries/RevitNodesUI/RevitDropDown.cs
sm6srw: Looks like this is obsoleted already but I will make a couple of comments anyway. First, it is a bad idea to filter by name. This string is localized so this will not work on non-English versions of Revit. It has to be based on the category Id. Second, we are aware of that some of the categories we expose in this list doesn't make sense and are not exposed in the Revit UI today because they  are internal categories. The problem is that there is no way right now to know if a category is internal because those properties has not been exposed in the public revit api (yet). I don't think we want our own, mirrored,  by category filtering mechanism on the Dynamo side. We want Revit to provide that information and until then I think that it is OK as it is right now.

src/Libraries/RevitNodes/Elements/Views/View.cs
mjkkirschner: cool, I did not notice they were already commented out.

src/Libraries/RevitNodes/Elements/Views/View.cs
mjkkirschner: I understand that, but the signature of this public method changed, if you can verify that an old graph using this node before you changed the signature opens fine after your change is made we can keep this.

mjkkirschner:
@ksobon @kronz indicated we should limit changes to public methods for now, please scan over this PR and limit these changes, and add secondary nodes with different names. (avoid overloads with the same name)

if you're confident that the old node was totally useless(and no one could have used it) I'm okay with just changing it here.

src/Libraries/RevitNodes/Elements/Views/View.cs
ksobon: I just tested that by creating and adding a ViewFilter to a View using Dynamo 1.2 out of the box. Then I built this, swapped DLLs, re-opened the same definition and it still worked just fine. Only the description and output name changes which seems to not have any effect on resolution of the graph. Are we good?

mjkkirschner: nice - thanks for testing.

src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
mjkkirschner: ok

ksobon:
No, that view color node was not totally useless, although it was a very poor implementation and given most recent changes - especially addition of `OverrideGraphicsSettings` node - it's just really bad. Anyways, I reverted the change, and instead just added a new node that will take advantage of the new functionality.

mjkkirschner:
@ksobon this is looking very solid. Do you have any tests around the add/remove fields methods and the new override in view node?

Are these easy to write?

src/Libraries/RevitNodesUI/RevitDropDown.cs
ksobon: Understood. I realized that it was a bad idea to begin with, and rolled that back. My bad, and thanks for the explanation.

ksobon:
Yeah, but it would be good to have a different test model than the Empty.rvt file. How do I submit a different test model?

mjkkirschner:
that model just lives in this folder:
https://github.com/DynamoDS/DynamoRevit/tree/Revit2017/test/System

you can just add your own .RVT file to that folder - please use a revit 2016 file so it can be converted to 2017 and 2018.

mjkkirschner:
@ksobon checking in here: are you planning to submit tests for this soon? I'd like to get this in, without tests if needed.

ksobon:
@mjkkirschner I added the wrappers that we talked about. I can take care of the unit tests in the next few days.

src/Libraries/RevitNodesUI/RevitNodesUI.csproj
```diff
@@ -156,7 +156,7 @@
   <ItemGroup>
     <EmbeddedResource Include="Properties\Resources.resx">
       <Generator>ResXFileCodeGenerator</Generator>
-      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+      <LastGenOutput>Resources1.Designer.cs</LastGenOutput>
```
mjkkirschner: this looks wrong... what is Resources1?

mjkkirschner:
@ksobon - yes, a test that will never fail is not such a great test 😉 .

If there was a case where the previous implementation would have failed but will succeed now, that would be a good test.

Also, I would like to see some tests on the add/remove fields methods.

mjkkirschner:
also, new wrappers look good.

src/Libraries/RevitNodesUI/RevitNodesUI.csproj
ksobon: oh my god, my bad. My mouse glitched up and instead of double click to open i dragged the resource file creating a copy of it. I deleted that thing as soon as I tried to build and ran into all the errors realizing what i did. I will try and build again so that gets overridden.

ksobon:
added these ScheduleView tests that you wanted. the thing with the OverrideGraphicsSettings node test is that it will not fail because all default arguments are null which means they are all correct. I don't think there was an instance where this would fail in a previous implementation, it was just not possible to satisfy all inputs. That was the issue. I think its a valid test for the component as that's its default status - with all inputs set to `null`. I would leave it as is. @moethu what you think?

mjkkirschner:
@ksobon this looks good. I will merge it and cherry pick to the other revit branches.

PR #1416, opened 2016-12-07
fix customization xml

mjkkirschner:
### Purpose

fix customization to move new schedule namespace under `Revit`.

### Reviewers

@QilongTang 

### FYIs

@ksobon @jnealb

### DIFF
--- a/src/Libraries/RevitNodes/RevitNodes_DynamoCustomization.xml
+++ b/src/Libraries/RevitNodes/RevitNodes_DynamoCustomization.xml
@@ -19,6 +19,9 @@
     <namespace name="Revit.Filter">
       <category>Revit.Filter</category>
     </namespace>
+    <namespace name="Revit.Schedules">
+      <category>Revit.Schedules</category>
+    </namespace>
     <namespace name="Revit.Geometry">
       <category>Revit.Geometry</category>
     </namespace>


ksobon:
did I screw this one up?

mjkkirschner:
nope, I forgot to mention this, this change sticks your new namespace under the revit cateogory instead of exposing the RevitNodes.Revit.Schedules category.

QilongTang:
Cool. Did not know that we have our own customized mapping of namespaces. LGTM.

src/Libraries/RevitNodes/RevitNodes_DynamoCustomization.xml
```diff
@@ -19,6 +19,9 @@
     <namespace name="Revit.Filter">
       <category>Revit.Filter</category>
     </namespace>
+    <namespace name="Revit.Schedules">
+      <category>Revit.Schedules</category>
+    </namespace>
```
QilongTang: @Racel @mjkkirschner Could this be Revit.Views.ScheduleView so it appear with other schedule Node at the same sub category? Should these three nodes be prefixed with ScheduleView for consistency even though this creates long names?

jnealb:
@mjkkirschner @QilongTang For example the ExportOptions that are needed for Export are under a different Revit subcategory, which is ackward.

Issue #1434, opened 2016-12-23
Can't build Revit2017 from source.


ksobon:
I made an update from 1.2 to 1.2.1 and I am having trouble building the repo. My steps were as follows. Build Dynamo 1.2.1_master (no issues here) then try building the Dynamo.Revit from Revit2017 branch and it fails with unresolved references. Here's the console output: 

1>------ Rebuild All started: Project: AssemblyInfoGenerator, Configuration: Debug Any CPU ------
2>------ Rebuild All started: Project: DynamoRevitIcons, Configuration: Debug Any CPU ------
3>------ Rebuild All started: Project: DynamoAddinGenerator, Configuration: Debug Any CPU ------
3>  DynamoAddinGenerator -> C:\Users\ksobon\Google Drive\Work\Dynamo\bin\AnyCPU\Debug\DynamoAddInGenerator.exe
2>  DynamoRevitIcons -> C:\Users\ksobon\Google Drive\Work\Dynamo\bin\AnyCPU\Debug\Revit_2017\DynamoRevitIcons.dll
1>  'C:\Users\ksobon\Google' is not recognized as an internal or external command,
1>  operable program or batch file.
1>C:\Users\ksobon\Google Drive\Work\DynamoRevit\src\AssemblySharedInfoGenerator\AssemblyInfoGenerator.csproj(78,5): warning MSB3073: The command "C:\Users\ksobon\Google Drive\Work\DynamoRevit\src\restorepackages.bat" exited with code 9009.

mjkkirschner:
@sharadkjaiswal 
```
Drive\Work\DynamoRevit\src\AssemblySharedInfoGenerator\AssemblyInfoGenerator.csproj(78,5): warning MSB3073: The command "C:\Users\ksobon\Google Drive\Work\DynamoRevit\src\restorepackages.bat" exited with code 9009.
```

sharadkjaiswal:
@ksobon are you using windows 10? Can you please run the `C:\Users\ksobon\Google Drive\Work\DynamoRevit\src\restorepackages.bat` file with admin privilege before starting your build?

sharadkjaiswal:
It seems space in the folder name `Google Drive` is creating the issue. @ellensi please look into it.

ellensi:
This build issue should have been fixed.

ksobon:
it still can't get it to build....

[...]

PR #1438, opened 2017-01-03
Symlink fix for directory names that contain spaces

ellensi:
### Purpose

To fix issue #1434 *Can't build Revit2017 from source.*

```
> 'C:\Users\ksobon\Google' is not recognized as an internal or external command, operable program
  or batch file.
> C:\Users\ksobon\Google Drive\Work\DynamoRevit\src\AssemblySharedInfoGenerator\AssemblyInfoGener
  ator.csproj(78,5): warning MSB3073: The command "C:\Users\ksobon\Google Drive\Work\DynamoRevit\
  src\restorepackages.bat" exited with code 9009.
```

### Reviewers

@sharadkjaiswal 

### FYIs

@ksobon

### DIFF
--- a/src/AssemblySharedInfoGenerator/AssemblyInfoGenerator.csproj
+++ b/src/AssemblySharedInfoGenerator/AssemblyInfoGenerator.csproj
@@ -75,6 +75,6 @@
   </Target>
   -->
   <Target Name="BeforeBuild">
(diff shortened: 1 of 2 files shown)


ksobon:
this still doesn't work for me. I keep getting the same error about missing command that is caused by a space in the name...that's what it looks like: 

image

mjkkirschner:
It seems to be failing in a slightly different way now. @ellensi should the packages.aget file exist?

PR #1447, opened 2017-01-10
Symlink fix for directory names that contain spaces (2)

ellensi:
### Purpose

To fix the remaining issue from #1434 *Can't build Revit2017 from source* which was addressed in PR #1438 *Symlink fix for directory names that contain spaces*.

```
> 'C:\Users\ksobon\Google' is not recognized as an internal or external command, operable program
  or batch file.
> Latest pre-release version of "DynamoVisualProgramming.Core" package is ""
> The system cannot find the file specified.
> Error occurred while processing: C:\Users\ksobon\Google.
> The system cannot find the file specified.
```

### Reviewers

@sharadkjaiswal 

### FYIs

@ksobon

### DIFF
--- a/src/restorepackages.bat
+++ b/src/restorepackages.bat
@@ -27,16 +27,16 @@ set NugetConfig=%ConfigDir%\dynamo-nuget.config
 REM 2. replace "LatestBeta" strings in packages-template.aget with the latest
 REM    pre-release version of DynamoVisualProgramming.Core package,
 REM    and save the replaced file as packages.aget
-set versionQuery=%NugetExe% list DynamoVisualProgramming.Core -prerelease -config "%NugetConfig%"
-for /F "tokens=2 delims= " %%F in ( '%versionQuery%' ) do (set LatestBeta=%%F)
+set versionQueryParams=list DynamoVisualProgramming.Core -prerelease -config "%NugetConfig%"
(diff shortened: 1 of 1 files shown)


ksobon:
I think we are getting there...one error at a time. Here's the latest. 

image

ksobon:
wow. This took a while but i tried your latest suggestion and it worked. However, i deleted the packages.aget and ran it the first time around and it failed due to lack of admin privileges. I ran this again, and it exited with the same error that i posted above. It seems that something is appending more stuff than needed to the *.aget file. Anyways, so i deleted that file again, re-launched Visual Studio as an Admin, and ran it.....viola! It built properly. I guess there are still things to be fixed with the aget file and how it gets appended on failed builds and the admin thing is painful. I don't see a whole lot of people having admin rights on their machines at work. That could be an issue.

mjkkirschner:
@ellensi - I just noticed this sadly fails for me since I store my source on a mac drive and access it using a windows VM, and windows cannot create symbolic links there 😢 .

I'm not sure this is a common setup, but I guess we should document it.

ksobon:
@mjkkirschner this should probably be opened again, as the latest Revit2018 has a broken version of this where restorepackages.bat fails to run if the path has a space in it....

PR #1450, opened 2017-01-12
new constructors for dimension type

ksobon:
### Purpose

This adds new wrappers around `DimensionType` object. It creates a constructor for creating new Dimension Types as well as retrieving its style. It was requested here: 

https://forum.dynamobim.com/t/change-dimension-style/8680/10?u=konrad_k_sobon

image

### Declarations

### Reviewers

@mjkkirschner @ikeough

### DIFF
--- a/src/Libraries/RevitNodes/Elements/DimensionType.cs
+++ b/src/Libraries/RevitNodes/Elements/DimensionType.cs
@@ -1,0 +1,153 @@
+﻿using System;
+using System.Linq;
+using Autodesk.DesignScript.Runtime;
+using DynamoServices;
+using RevitServices.Persistence;
+using RevitServices.Transactions;
+
+namespace Revit.Elements
+{
+    /// <summary>
+    /// Dimension Type element.
+    /// </summary>
+    [RegisterForTrace]
+    public class DimensionType : Element
+    {
+        #region Internal Properties
+
+        /// <summary>
+        /// Internal reference to the Revit Element
+        /// </summary>
+        internal Autodesk.Revit.DB.DimensionType InternalRevitElement
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


kronz:
tracking review at https://jira.autodesk.com/browse/DYN-486

kronz:
@QilongTang this looks really small, any chance you have to take a look and merge?  Does this need tests?

QilongTang:
@kronz Sure, I will move DYN-486 to the current sprint then and review this PR.

src/Libraries/RevitNodes/Elements/DimensionType.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            var document = DocumentManager.Instance.CurrentDBDocument;
+            TransactionManager.Instance.EnsureInTransaction(document);
+
+            // get element from trace
+            var element = ElementBinder.GetElementFromTrace<Autodesk.Revit.DB.DimensionType>(document);
+
+            if (element == null)
+            {
+                element = dimType.Duplicate(name) as Autodesk.Revit.DB.DimensionType;
```
QilongTang: @ksobon Does this mean each time user run the graph we unconditionally create a new DimensionType with the new name? Does not seem to scan through the existing DimensionTypes. I imagine it could cause trouble in certain cases. Please correct me if I was wrong.

ksobon: it's quite the opposite. We are retrieving an element from trace, if that element is null, then we create a new one. that ensures that we are not creating a new element every time graph executes.

QilongTang: Great. I was reading through the conversation you had with @mjkkirschner about using trace to avoid creating duplicate elements. Thanks for the comments!

PR #1480, opened 2017-01-25
Removed Additional Elements and returning acutal elements

moethu:
### Purpose

Removed Additional Elements and returning actual elements for failure messages.
This is from Aaron Tang's Review.

### Declarations

### Reviewers

@mjkkirschner @kronz

### DIFF
--- a/src/Libraries/RevitNodes/Elements/FailureMessage.cs
+++ b/src/Libraries/RevitNodes/Elements/FailureMessage.cs
@@ -37,23 +37,23 @@ namespace Revit.Elements
         /// <summary>
         /// The Failing Elements of the message.
         /// </summary>
-        public ICollection<ElementId> FailingElements
+        public ICollection<Element> FailingElements
         {
             get
             {
-                return InternalElement.GetFailingElements();
-            }
-        }
+                var failingIds = InternalElement.GetFailingElements();
+                var failingElements = new List<Element>();
 
+                foreach (var failingId in failingIds)
+                {
+                    if (failingId != ElementId.InvalidElementId)
+                    {
+                        var element = DocumentManager.Instance.CurrentDBDocument.GetElement(failingId);
+                        failingElements.Add(element.ToDSType(true));
+                    }
+                }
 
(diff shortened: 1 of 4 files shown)


QilongTang:
Hey, @moethu The changes to FailingElements looks good. Just to be clear about other note @kronz mentioned in the email:
1. We found PostInDocument Node is not that useful, Remove that
2. The use for AdditionalElements is not super clear, can you give an example?
3. The node category for PerformanceAdviserRules is wrong, can we put the node not in Revit Root Category but PerformanceAdviser category?
4. Also a trigger to run does not refresh the PerformanceAdviserRule.Execute results, we had to change the rule from dropdown list and change back to make it work. Can you take a look?

moethu:
@QilongTang (CC @kronz ) thanks, got it.
1. removed
2. are elements that are affected but not directly conflicting (removed node)
3. you mean so it results in: Revit.PerformanceAdviser ? (done)
4. Yes, I was not sure how you want me to handle this.

kronz:
@moethu @QilongTang for 4 . . . I definitely don't want to create an opportunity for the node to just be running endlessly.  How does this work for things like the Family Types dropdown . . . where the contents refresh when new items are loaded.  Or "categories" passed to "all elements of category" that adds new elements when instances are placed?  
The other option is to add a "refresh" input port, as some packages have done, that takes a boolean toggle to kick it, and give it a default value so it at least runs once, as it currently does.

moethu:
@kronz @mjkkirschner did you guys implement an invalidate trigger to force the node to rerun? We talked about this a lot but I don't remember what the latest status was.

kronz:
@moethu I don't think there has been any functionality to specifically mark a node as Dirty, but defer to @mjkkirschner to say

QilongTang:
Same note as @sm6srw mentioned try out the RevitServicesUpdater.Instance  may get what we need here @moethu

QilongTang:
Merging this in right now. I talked with @mjkkirschner on the refreshing mechanism and will try to implement it.

PR #1491, opened 2017-01-28
R2017 view schedule filters

ksobon:
### Purpose

This PR adds new functionality to `ScheduleView` class enabling ability to add or remove Schedule Filters. It was requested here: https://forum.dynamobim.com/t/schedules-filtering-replace-filter/9077 and in many other forum conversations. 

### Declarations

### Reviewers

@aparajit-pratap @mjkkirschner @ikeough 

### FYIs

Schedule Filter class does not inherit from `Revit.Element` class. It also doesn't have certain restrictions like unique name etc. so unfortunately there is no Element Binding support here and I had to override the `Equals` method on it, to enable `Contains` checks. Let me know if what I did there is acceptable.


ksobon:
@mjkkirschner @ikeough @aparajit-pratap @kronz is anyone there? It's been two weeks guys. Can someone review these?

kronz:
@ksobon sorry, this got lost in the shuffle.  We can look later this week.

QilongTang:
Actually Looks pretty good, thoughts @mjkkirschner @ikeough ?

kronz:
tracking review process at https://jira.autodesk.com/browse/DYN-484

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
```diff
@@ -145,10 +154,10 @@ private static Autodesk.Revit.DB.ViewSchedule CreateViewSchedule(Category catego
         /// <summary>
         /// Create Schedule by Category, Type and Name.
         /// </summary>
-        /// <param name="category"></param>
-        /// <param name="name"></param>
-        /// <param name="scheduleType"></param>
-        /// <returns name="scheduleView">Schedule</returns>
+        /// <param name="category">Category that Schedule will be associated with.</param>
```
mjkkirschner: nice, thanks for the tooltips.

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
```diff
@@ -249,6 +258,62 @@ public ScheduleView AddFields(List<Revit.Schedules.SchedulableField> fields)
...
+        /// </summary>
+        /// <param name="scheduleFilters">List of Schedule Filters.</param>
+        /// <returns name="scheduleView">Schedule View.</returns>
+        public ScheduleView AddFilters(List<ScheduleFilter> scheduleFilters)
+        {
+            var doc = DocumentManager.Instance.CurrentDBDocument;
+            TransactionManager.Instance.EnsureInTransaction(doc);
+
+            var currentFilters = this.InternalScheduleFilters.Select(x => new ScheduleFilter(x)).ToList();
+            foreach (var sf in scheduleFilters)
```
mjkkirschner: are these lists bounded by a small number? Like say 1000?

src/Libraries/RevitNodes/Schedules/ScheduleFilter.cs
```diff
@@ -0,0 +1,161 @@
+﻿using System;
+using Autodesk.DesignScript.Runtime;
+
+namespace Revit.Schedules
```
mjkkirschner: where does this end up in the library tree? can you attach an image?

src/Libraries/RevitNodes/Schedules/ScheduleFilter.cs
```diff
@@ -0,0 +1,161 @@
...
+                else
+                {
+                    return null;
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            return string.Format("ScheduleFilter(FilterType: {0}, FieldId: {1}, Value: {2})", this.FilterType, this.FiledId, this.Value);
```
mjkkirschner: is it possible that any of these properties are null?

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
ksobon: Yeah, I haven't tested exactly, but in Revit UI you can only add some 10 filters. I would venture a guess that API is limited to a similar number as well.

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
ksobon: you are welcome!

src/Libraries/RevitNodes/Schedules/ScheduleFilter.cs
ksobon: I don't think so since they are all required for the filter to be properly defined. The `value` can be an empty string for example but not null.

src/Libraries/RevitNodes/Schedules/ScheduleFilter.cs
ksobon: You overrode that namespace in Customization file last time so it falls into ScheduleView: 
image

mjkkirschner:
Thanks @ksobon - this looks really solid to me. Thanks for the hard work. We can create the cherry picks to 2016 and 2018.

Do you anticipate any issues with 2016?

ksobon:
Nah, it is old functionality. It should work just fine in 2016.

src/Libraries/RevitNodes/Elements/Views/ScheduleView.cs
ksobon: I tested this and it looks that API is equally constrained as the UI version. The limit is 8 filters. If you try adding more this errors appears which is pretty self explanatory. 

image

Would you prefer to catch it and throw our own exception? Should we add a unit test for this condition?

mjkkirschner: nope, this is fine, my only concern was the nested contains check, but since the number is tiny it doesn't matter.

mjkkirschner:
I'm going to merge this in for 2017 and I will do the cherry picks after to 2016,2018.

jnealb:
@ksobon @mjkkirschner Can you please add a system test or provide a rvt/dyn sample for testing?

ksobon:
@jnealb my PR had Unit Tests in it. I also submitted a Revit model to test on. Is there anything else that I missed?

mjkkirschner:
system test would involve opening a .dyn file and running on that model.

ksobon:
do I write those? i can share a dyn file to run this on. that's not an issue. if i need to write one of these just point me to an example.

jnealb:
@ksobon Sorry, where is the Unit Test and Revit model? Am I missing something: I do not see anything in PR 1491? Perhaps you could email me the samples: neal.burnham@autodesk.com?

ksobon:
it's this part of the code in `ScheduleViewTests.cs` file. 

```
[Test]
 +        [TestModel(@".\SampleModel.rvt")]
 +        public void AddScheduleFilter_ValidArgs()
 +        {
 +            var view = ScheduleView.CreateSchedule(Category.ByName("OST_Walls"), "WallSchedule_Test", ScheduleView.ScheduleType.RegularSchedule.ToString());
 +            Assert.NotNull(view);
 +
 +            // get a list of schedulable fields and add "Width" field to schedule
 +            var schedulableFields = view.SchedulableFields;
 +            Assert.Greater(schedulableFields.Count, 0);
 +
 +            var fieldToAdd = schedulableFields.Where(x => x.Name == "Width").FirstOrDefault();
 +            Assert.NotNull(fieldToAdd);
 +
 +            view.AddFields(new List<Revit.Schedules.SchedulableField>() { fieldToAdd });
 +            Assert.Greater(view.Fields.Count, 0);
 +
 +            // create new schedule filter and add it to schedule
 +            var field = view.Fields.Where(x => x.Name == "Width").FirstOrDefault();
 +            Assert.NotNull(field);
 +
 +            var filter = Revit.Schedules.ScheduleFilter.ByFieldTypeAndValue(field, Autodesk.Revit.DB.ScheduleFilterType.Equal.ToString(), 0.1);
 +            Assert.NotNull(filter);
 +
 +            view.AddFilters(new List<Revit.Schedules.ScheduleFilter>() { filter });
 +            Assert.Greater(view.ScheduleFilters.Count, 0);
 +        }
 +
 +        [Test]
 +        [TestModel(@".\SampleModel.rvt")]
 +        public void ClearAllScheduleFilters_ValidArgs()
 +        {
 +            var rView = new Autodesk.Revit.DB.FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument)
 +                .OfClass(typeof(Autodesk.Revit.DB.ViewSchedule))
 +                .Cast<Autodesk.Revit.DB.ViewSchedule>()
 +                .Where(x => x.Name == "Wall Schedule")
 +                .FirstOrDefault();
 +
 +            var view = ScheduleView.FromExisting(rView, true);
 +            Assert.NotNull(view);
 +
 +            // remove all schedule filters
 +            view.ClearAllFilters();
 +            Assert.AreEqual(view.ScheduleFilters.Count, 0);
 +        }
```

I can email you the DYN and RVT files when I get home.

jnealb:
Thanks much ☺

N

ksobon:
please see this: https://github.com/DynamoDS/DynamoRevit/pull/1590

Issue #1410, opened 2016-11-28
ImportInstance selected using SelectModelElement node is reported as UnknownElement


ThomasMahon:
## Dynamo version

1.2

## Revit version

Revit 2016

## Operating system

Win 10

## What did you do? 

Selected an ImportInstance (a linked DWG file) using Revit library `SelectModelElement` node

## What did you expect to see?

For the output to identify as a ImportInstance

## What did you see instead?

Dynamo reports an Unknown Element. This causes a number of problems, in particular for Custom Nodes (which triggered my investigation) as any input type declaration that uses ImportInstance breaks the node.

![image](https://cloud.githubusercontent.com/assets/18601851/20670971/5a32e77a-b573-11e6-9fe4-0e984564f00d.png)

mjkkirschner:
@ThomasMahon did this used to work?

ThomasMahon:
Not since I begun testing on V1.0. However, I only discovered the cause of the problem, i.e. UnknownElement, after looking at ways to solve it for a upcoming package release.

ThomasMahon:
@mjkkirschner one other thing to clarify in respect to the error with the CN: the simple solution to prevent the exception is not to declare the type in the Input node. Downside is, there are no hints to the user as to the type of object that needs to be input. Either way, its certainly a problem.

ThomasMahon:
@mjkkirschner also, I've found that if the base class is used (`Revit.Elements.Element`) then no exception is thrown. Only `ImportInstance` class causes this issue.

ksobon:
@ThomasMahon @mjkkirschner this is all happening because for some reason the `ImportInstance` wrapper class was implemented differently than all other wrappers. Mainly, there were initiate methods, fromExisting and Wrapper missing from the class. Once that was added it works as expected: 

![image](https://cloud.githubusercontent.com/assets/529781/21952423/0618461c-d9ea-11e6-8908-46ef53348e48.png)

@mjkkirschner is there a reason not to add these to `ImportInstance` class: 

```
        /// <summary>
        /// Set the internal Element, ElementId, and UniqueId
        /// </summary>
        /// <param name="element"></param>
        private void InternalSetElement(Autodesk.Revit.DB.ImportInstance element)
        {
            InternalImportInstance = element;
            InternalElementId = element.Id;
            InternalUniqueId = element.UniqueId;
        }

        /// <summary>
        /// ImportInstance from existing
        /// </summary>
        /// <param name="element"></param>
[...]

PR #1492, opened 2017-01-28
import instance wrapper

ksobon:
### Purpose

This adds wrapper to `ImportInstance` class so that its not returned as `UnknownElement`. It was raised here: https://github.com/DynamoDS/DynamoRevit/issues/1410

### Declarations

### Reviewers

@ikeough @mjkkirschner @aparajit-pratap 

### FYIs

- None


aparajit-pratap:
@ksobon could you attach an example Revit file and DYN as a potential test case for this fix? We can then have that included in our test suite. Thanks!

ksobon:
@aparajit-pratap This is really a minor change where we were missing a `FromExisting` implementation on the `ImportInstance` class. That's the only thing that I added here. There are no other changes and whatever tests you were performing previously should not be affected. 

Do you need a Revit + Dyn file to be placed in the usual repo location with the rest of the test models? I can make a small model, but I am not sure its necessary.

kronz:
tracking review process in https://jira.autodesk.com/browse/DYN-483

src/Libraries/RevitNodes/Elements/ImportInstance.cs
```diff
@@ -27,7 +27,22 @@ public override Autodesk.Revit.DB.Element InternalElement
...
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Set the internal Element, ElementId, and UniqueId
+        /// </summary>
+        /// <param name="element"></param>
+        private void InternalSetElement(Autodesk.Revit.DB.ImportInstance element)
```
aparajit-pratap: @ksobon I don't think this method is necessary. We already have `InternalSetImportInstance`

src/Libraries/RevitNodes/Elements/ImportInstance.cs
```diff
@@ -82,6 +106,15 @@ private void InitImportInstance(string satPath, XYZ translation = null)
             ElementBinder.SetElementForTrace(importInstance);
         }
 
+        /// <summary>
+        /// Initilize an ImportInstance from existing element
+        /// </summary>
+        /// <param name="instance"></param>
+        private void InitImportInstance(Autodesk.Revit.DB.ImportInstance instance)
```
aparajit-pratap: I think we can get rid of this as well and just use `InternalSetImportInstance`

aparajit-pratap:
@ksobon some minor comments. Other than that it is good to go.

aparajit-pratap:
@ksobon I'm not sure if it's just me but I tested your changes and am getting the following error with a simple rvt file. Can you reproduce this?
image

aparajit-pratap:
Ok the issue was lurking from earlier - `InternalSetImportInstance` was incorrect. It should be changed to:
```
private void InternalSetImportInstance(Autodesk.Revit.DB.ImportInstance ele)
        {
            this.InternalImportInstance = ele;
            this.InternalElementId = ele.Id;
            this.InternalUniqueId = ele.UniqueId;
        }
```

aparajit-pratap:
@ksobon please take a look at https://github.com/DynamoDS/DynamoRevit/pull/1575 for the test case. Thanks for the fix!

ksobon:
@aparajit-pratap oh wow! I didn't see that coming. My bad. Why does the order matter here? It's kind of disturbing that we would remember something like this for the future and there was no "obvious" reason for this specific order.

aparajit-pratap:
@ksobon that seems strange to me too but I don't have an answer right now. Will have to investigate this in more detail to get more clarity but for now, I'm going ahead and closing this PR and merging #1575 which has the test case. Thanks again!

PR #1556, opened 2017-02-16
Add journal keys for node values to be pushed into revit dynamo model

BogdanZavu:
### Purpose
We want to be able to push nodes values when we execute a DynamoRevit command. We receive these values from a third party ( like Dynamo Player ) as a command parameter and we push them into nodes using UpdateModelValueCommand

### Reviewers
@mjkkirschner

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -27,6 +27,7 @@ using Dynamo.ViewModels;
 using DynamoInstallDetective;
 using Greg.AuthProviders;
 using Microsoft.Win32;
+using Newtonsoft.Json;
 using RevitServices.Persistence;
 using RevitServices.Threading;
 using DynUpdateManager = Dynamo.Updates.UpdateManager;
@@ -149,6 +150,21 @@ namespace Dynamo.Applications
         /// needs to be shutdown before performing any action.
         /// </summary>
         public const string ModelShutDownKey = "dynModelShutDown";
+
+        /// <summary>
+        /// The journal file can specify the values of Dynamo nodes.
+        /// </summary>
+        public const string ModelNodesInfo = "dynModelNodesInfo";
+    }
+
+    /// <summary>
+    /// Defines parameters for Dynamo nodes
+    /// </summary>
+    public class JournalNodeKeys
+    {
+        public const string Id = "Id";
+        public const string Name = "Name";
+        public const string Value = "Value";
     }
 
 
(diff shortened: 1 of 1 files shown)


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -546,6 +562,27 @@ private static void TryOpenAndExecuteWorkspaceInCommandData(DynamoRevitCommandDa
                     dynamoViewModel.ShowStartPage = false;
                 }
 
+                //If we have information about the nodes and their values we want to push those values after the file is opened.
+                if (commandData.JournalData.ContainsKey(JournalKeys.ModelNodesInfo))
+                {
+                    try
+                    {
+                        var allNodesInfo = JsonConvert.DeserializeObject<List<Dictionary<string, string>>>(commandData.JournalData[JournalKeys.ModelNodesInfo]);
```
BogdanZavu: We use this generic dictionary here which comes from a json string received as parameter instead of a more specific data structure that we taught we'll be harder to transmit as a parameter.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -546,6 +562,27 @@ private static void TryOpenAndExecuteWorkspaceInCommandData(DynamoRevitCommandDa
...
+                        if (allNodesInfo != null)
+                        {
+                            foreach (var nodeInfo in allNodesInfo)
+                            {
+                                if (nodeInfo.ContainsKey(JournalNodeKeys.Id) && nodeInfo.ContainsKey(JournalNodeKeys.Name) && nodeInfo.ContainsKey(JournalNodeKeys.Value))
+                                {
+                                    DynamoModel.UpdateModelValueCommand modelCommand = new DynamoModel.UpdateModelValueCommand(nodeInfo[JournalNodeKeys.Id], nodeInfo[JournalNodeKeys.Name], nodeInfo[JournalNodeKeys.Value]);
```
mjkkirschner: small comment: can you use `var` here and split this into multiple lines - these are getting a bit too long.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -546,6 +562,27 @@ private static void TryOpenAndExecuteWorkspaceInCommandData(DynamoRevitCommandDa
...
+                                    modelCommand.Execute(revitDynamoModel);
+                                }
+                            }
+                        }
+                    }
+                    catch { }
```
mjkkirschner: hmm, can we log the console at least?

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -149,6 +150,21 @@ public class JournalKeys
...
         public const string ModelShutDownKey = "dynModelShutDown";
+
+        /// <summary>
+        /// The journal file can specify the values of Dynamo nodes.
+        /// </summary>
+        public const string ModelNodesInfo = "dynModelNodesInfo";
+    }
+
+    /// <summary>
+    /// Defines parameters for Dynamo nodes
```
mjkkirschner: maybe mention UpdateModelValue command in this comment?

BogdanZavu:
@mjkkirschner I made the suggested changes. Also that's the branch from which RC2.0_Revit2018 will be created right ?

mjkkirschner:
@BogdanZavu this looks good to me, yes I believe so about the branch.

Can you also create Prs cherry picking this work to other branches - I would guess to 2017? and I will merge it.

BogdanZavu:
@mjkkirschner We actually don't want these changes in 2017 ; They need to be only for Revit2018 and greater. Do you think it's worth having them as generic enhancements for 2017 anyway ?

mjkkirschner:
Yes, I think it's worth having them around if they cause no harm they open possibilities to other dynamo users, and make future cherry picks potentially easier. Where possible I think having the branches similar makes it easier to reason about.

BogdanZavu:
@mjkkirschner Ok

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -27,6 +27,7 @@
 using DynamoInstallDetective;
 using Greg.AuthProviders;
 using Microsoft.Win32;
+using Newtonsoft.Json;
```
mjkkirschner: @BogdanZavu looks like the 2017 branch is missing the using statement for json.net

BogdanZavu:
@mjkkirschner Yes . I created another PR to fix it.

PR #1595, opened 2017-03-02
DYN-488 Revit 2018

sm6srw:
### Purpose
Parameter names are not unique on a given element. There are several valid cases where 
duplicated parameter names can be found in the Parameter Set.

The most common ones are:
 * Multiple built-in parameters with the same name
This is a common implementation pattern when a different parameter behavior is wanted
for different scopes. For example, lets say that you want a parameter to be writable
in the property palette but read-only in a schedule view. The easiest way to accomplish
this would be to add two parameters. One that is read-write and one that is read-only. 
These parameters will both have the same name and they will share the same getter. 
* Built-in parameters and User parameters with the same name
This happens when a loadable family defines a user parameter with the same name
as a built-in parameter.

Proposed solution:
1. Get all parameters for the given name
2. Sort parameters by ElementId - This will give us built-in parameters first (ID's for built-ins are always < -1)
3. If it exist: Use the first writable parameter 
4. Otherwise: Use the first read-only parameter

This should solve the first case and address the most common scenario for the second case. A possible future enhancement, if the need arises, would be to add a way for the user to ignore built-ins. 

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -343,7 +343,7 @@ namespace Revit.Elements
             //
             var allParams =
             InternalElement.Parameters.Cast<Autodesk.Revit.DB.Parameter>()
-                .Where(x => x.Definition.Name == parameterName)
+                .Where(x => string.CompareOrdinal(x.Definition.Name, parameterName) == 0)
                 .OrderBy(x => x.Id.IntegerValue);
 
             var param = allParams.FirstOrDefault(x => x.IsReadOnly == false) ?? allParams.FirstOrDefault();
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -318,6 +318,29 @@ namespace Revit.Elements
         /// <returns></returns>
         private Autodesk.Revit.DB.Parameter GetParameterByName(string parameterName)
         {
+            //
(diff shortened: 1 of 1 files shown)


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -310,22 +310,55 @@ public void Tessellate(IRenderPackage package, TessellationParameters parameters
...
+            //
+            // Currently, we try to resolve this and get consistent results by
+            // 1. Get all parameters for the given name
+            // 2. Sort parameters by ElementId - This will give us built-in parameters first (ID's for built-ins are always < -1)
+            // 3. If it exist: Use the first writable parameter
+            // 4. Otherwise: Use the first read-only parameter
+            //
+            var allParams =
+            InternalElement.Parameters.Cast<Autodesk.Revit.DB.Parameter>()
+                .Where(x => x.Definition.Name == parameterName)
```
mjkkirschner: how does this work with localization? Should this be ordinal comparison or is == good enough?

sm6srw: Well, I am not sure. I would say it wouldn't hurt to use CompareOrdinal(). I see that we have used it for similar names in other places. I will change.

sm6srw:
Yes, all node tests succeeded. I had some problems with some of the system tests because of a Revit process that refused to exit but I will rerun those tonight.

PR #1601, opened 2017-03-06
Fix failing test cases

Randy-Ma:
### Purpose

Recently there are three test cases which are failing:
- Level
- RoomProperties
- DividedCurve

The two test cases - Level and DividedCurve are failing because the change in the number node. (Refer details in the pull request: https://github.com/DynamoDS/Dynamo/pull/7582). @ikeough, the test cases are failing because "0..1...1" and "0..20..2" are not supported in the new number node any more. I have just updated the related test cases to fix the issue.

The other test case - RoomProperties is a newly added test case. I do not think it is proper to just compare two doubles 12.869999999999999d and 12.87 without considering tolerance. I am updating the test case to compare in a tolerance of 1e-9.

### Declarations

### FYIs

@ikeough @kronz


Randy-Ma:
@ikeough @kronz 
I am merging this given this is breaking the build. If you have any comments, I will address them later.

PR #1624, opened 2017-03-16
Viewschedule1.32018

mjkkirschner:
Purpose

This PR cherry picks multiple PRs from Ksobons view schedule work to 1.3 2018 branch.

@QilongTang please take a look...


src/Libraries/RevitNodesUI/Properties/Resources.resx
```diff
@@ -372,13 +372,14 @@
     <value>The type of the fill pattern.</value>
```
QilongTang: This is actually interesting, some of the resource changes are already cherry-picked.. I think I'll have to build to find out but everything else looks good

Issue #1652, opened 2017-04-21
Family Instance.Type problem


rbosborne:
Using Dynamo 1.2.2 and 1.3.0

Revit 2016

See issue discussed here on Forum
[https://forum.dynamobim.com/t/familyinstance-type-problems/11197/4](url)

basically - using FamilyInstance.Type fails in an odd manner due to inherited Revit Types
see attached files and images

https://www.dropbox.com/s/uejjspljikyuuw2/FamilyInstanceProblem.zip?dl=0
![capture](https://cloud.githubusercontent.com/assets/12659052/25280249/40b5c338-2677-11e7-818c-4620628008c7.PNG)

andydandy74:
Issue in 2017 as well.
Also, all of those nodes (including AdaptiveComponent.Type) should just be one node using the API method @rbosborne is using here. At least there should not be two nodes for ```FamilyInstance``` that do exactly the same.

ksobon:
@mjkkirschner @kronz this is a perfect example of something that should be deprecated in the next release of Dynamo. Is there a way for us to mark these nodes deprecated for the next release, replace them all with just one, Element.Type node and then two releases from now, just delete them? There is another node in the View class that should be deleted, but @kronz stated that it should stay in the repo for the time being. There needs to be a way for us to purge the code base from time to time.

ksobon:
@mjkkirschner  got it. `[Obsolete()]` attribute is just what we need. Nice. So this can be "fixed" temporarily like so: 

![untitled-1](https://cloud.githubusercontent.com/assets/529781/26022248/3a71ff34-376e-11e7-93bd-fd2d50cc314c.jpg)

The nodes are yellow because I marked all of them Obsolete. We would delete any and all `GetType` and `Type` properties. All of these nodes inherit from `Element` class which means `Element.ElementType` will work for all of them. That makes more sense.

smangarole:
@mjkkirschner Ping.

johnpierson:
Closing this out as part of a backlog cleanup. The thread went quiet years ago and the versions involved are well behind current.

If this still reproduces on a current Dynamo and Revit, please open a fresh issue with the graph and a sample model attached.

PR #1676, opened 2017-05-13
fix to failing familyInstance.getType node

ksobon:
### Purpose

This addresses this issue: https://github.com/DynamoDS/DynamoRevit/issues/1652

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

I had to add an extra obsolete class to fix something else, but it's just temporary before next release when we can just delete them all. Right?

### DIFF
--- a/src/Libraries/RevitNodes/Elements/AbstractFamilyInstance.cs
+++ b/src/Libraries/RevitNodes/Elements/AbstractFamilyInstance.cs
@@ -1,8 +1,5 @@
 ﻿using Autodesk.DesignScript.Runtime;
-using Autodesk.Revit.DB;
-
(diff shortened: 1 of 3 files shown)


kronz:
@ksobon this one is going to need some more extensive review. @sm6srw can you say a little about why this one is a little more complicated?

ksobon:
@kronz why? All I did was clean up and make consistent the type retrieval methods. For some reason they weren't working consistently.

mjkkirschner:
I know one issue is that there are c# 6.0 features in this PR which might cause problems with older versions of revit - we need to check with addin team.

sm6srw:
New PR is #1780.

Issue #1674, opened 2017-05-12
All Elements of Type fails for RoomTagType


andydandy74:
## Dynamo version

1.3 stable

## Revit version

2017.1

## Operating system

Win 7

## What did you do? 

![roomtagtypecollector](https://cloud.githubusercontent.com/assets/3014437/25995439/897a9d5e-3712-11e7-9dc4-d85850e5d848.PNG)

## What did you expect to see?

RoomTagTypes

## What did you see instead?

```
Warnung:ElementQueries.OfElementType fehlgeschlagen.
Input type(Autodesk.Revit.DB.Architecture.RoomTagType) is of an element type that exists in the API, but not in Revit's native object model. Try using Autodesk.Revit.DB.FamilySymbol instead, and then postprocessing the results to find the elements of interest.
Parameter name: type
```

jnealb:
@kronz @riteshchandawar @Racel Ping

ksobon:
@andydandy74 it will also fail for other element types since this node uses `ElementClassFilter()` to retrieve them. Here's documentation bit from Revit API SDK: 

![image](https://cloud.githubusercontent.com/assets/529781/26041476/18249214-38fb-11e7-94be-292c898b4fdc.png)

Here's how to fix it: 

![image](https://cloud.githubusercontent.com/assets/529781/26041612/1a84e850-38fc-11e7-8d7a-97104e138c7f.png)

Code:

```
private static readonly Dictionary<Type, string> ClassFilterExceptions = new Dictionary<Type, string>
        {
            { typeof(Autodesk.Revit.DB.Material), "Material"},
            { typeof(Autodesk.Revit.DB.CurveElement), "CurveElement"},
            { typeof(Autodesk.Revit.DB.ConnectorElement), "ConnectorElement"},
            { typeof(Autodesk.Revit.DB.HostedSweep), "HostedSweep"},
            { typeof(Autodesk.Revit.DB.Architecture.Room), "Room"},
            { typeof(Autodesk.Revit.DB.Mechanical.Space), "Space"},
            { typeof(Autodesk.Revit.DB.Area), "Area"},
            { typeof(Autodesk.Revit.DB.Architecture.RoomTag), "RoomTag"},
            { typeof(Autodesk.Revit.DB.Mechanical.SpaceTag), "SpaceTag"},
            { typeof(Autodesk.Revit.DB.AreaTag), "AreaTag"},
[...]

PR #1677, opened 2017-05-15
fix to ElementClassFilter handling in All Elements of Type node

ksobon:
### Purpose

It addresses an issue listed here: 

https://github.com/DynamoDS/DynamoRevit/issues/1674

Here's the fixed example: 

image

### Declarations

### Reviewers

@mjkkirschner @jnealb @riteshchandawar @kronz @Racel 

### FYIs

I also made a small change to the way that previous collector was called. There was no need to call `ToElements()` and then `x.Id.IntegerValue` when we can call `ToElementIds()` directly. I believe that's a little faster as well.

### DIFF
--- a/src/Libraries/RevitNodes/Elements/InternalUtilities/ElementQueries.cs
+++ b/src/Libraries/RevitNodes/Elements/InternalUtilities/ElementQueries.cs
@@ -10,6 +10,31 @@ namespace Revit.Elements.InternalUtilities
     [IsVisibleInDynamoLibrary(false)]
     public static class ElementQueries
     {
+        private static readonly Dictionary<Type, string> ClassFilterExceptions = new Dictionary<Type, string>
+        {
+            { typeof(Autodesk.Revit.DB.Material), "Material"},
+            { typeof(Autodesk.Revit.DB.CurveElement), "CurveElement"},
+            { typeof(Autodesk.Revit.DB.ConnectorElement), "ConnectorElement"},
+            { typeof(Autodesk.Revit.DB.HostedSweep), "HostedSweep"},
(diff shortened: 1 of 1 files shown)


src/Libraries/RevitNodes/Elements/InternalUtilities/ElementQueries.cs
```diff
@@ -10,6 +10,31 @@ namespace Revit.Elements.InternalUtilities
     [IsVisibleInDynamoLibrary(false)]
     public static class ElementQueries
     {
+        private static readonly Dictionary<Type, string> ClassFilterExceptions = new Dictionary<Type, string>
```
mjkkirschner: I believe what you actually want here is a hashet<t> as your stored value is not really used right?

ksobon: Correct. It's only stored for lookup purposes, but `HashSet` vs `Dictionary` contains tests are both equally fast. There is no advantage to using one over the other. http://stackoverflow.com/questions/2728500/hashsett-versus-dictionaryk-v-w-r-t-searching-time-to-find-if-an-item-exist

mjkkirschner: yep, but the hashset's purpose is more explicit for what you are doing.

ksobon: Yes, it would be. I will get that fixed.

Racel:
Fix looks good after testing. It is in master and will be part of next release.

Issue #1406, opened 2016-11-27
Issue with Parameter Type Node.


dhourihane:
## Dynamo version

1.2.1.0
Core 1.2.0.2690
Revit 1.2.0.2690

## Revit version

2016
Build: 	16.0.490.0
		20150714_1515(x64)
		Service Pack 2

## Operating system

Windows 10

## What did you do? 

I connected the 3 nodes seen in the Image, selected an Element (a Floor) in Revit, and pressed 'Run'. I had originally run with just the first 2 Nodes connected to get the list of available Parameters.

## What did you expect to see?

I expected the 3rd Node to return a Parameter Type (it is a TEXT Parameter).

## What did you see instead?

See error message on image (screenshot).
![bug screenshot 20161127](https://cloud.githubusercontent.com/assets/23583286/20645027/aa1ea7e4-b445-11e6-9034-edbb818ba82c.png)

ksobon:
@dhourihane The issue here is that the GetFamilyParameter node returns a String name of the Parameter Definition not the actual Parameter object. What you can do for now is this: 

![image](https://cloud.githubusercontent.com/assets/529781/20651121/e4f95dc0-b4ac-11e6-97cc-b9c8aa195066.png)

Yes, I agree this is strange and unexpected. I would either recommend that we change the output from the component to return `Revit.Elements.Parameter` or change the name of the port to `paramName` or something like that to prevent any confusion. 

cc: @mjkkirschner @kronz @moethu

dhourihane:
Thanks @ksobon. Much appreciated.

ksobon:
I would not close it yet. I think we have raised some valid issues here...

ksobon:
cc @kronz and @mjkkirschner please see above. What do you think?

ksobon:
@kronz do you agree that the `Get Family Parameter` node should actually be returning a parameter? Here's an example of what that would look like: 

![image](https://cloud.githubusercontent.com/assets/529781/26337695/99485e22-3f48-11e7-839c-253636b04e75.png)

It would require changes to both the dropdown node and the `Parameter` class. I think it's worth the hassle. It should be consistent like that.

kronz:
@ksobon I think the Get Family Parameter node has typically been used in a Get and Set Parameter workflow, and returning a Parameter would break all of these.  It might be appropriate to rename it to Get Family Parameter Name, I suppose, but not to change it's function.
[...]

PR #1688, opened 2017-05-23
fix to get family parameter node

ksobon:
### Purpose

It addresses this issue: https://github.com/DynamoDS/DynamoRevit/issues/1406

### Declarations

### Reviewers

@mjkkirschner

### FYIs

@kronz


kronz:
@ksobon I think the Get Family Parameter node has typically been used in a Get and Set Parameter workflow, and returning a Parameter would break all of these. It might be appropriate to rename it to Get Family Parameter Name, I suppose, but not to change it's function.

ksobon:
Then it's totally misnamed, as both the node name and output port suggest something else. Should we be adjusting that?

Issue #1691, opened 2017-05-29
Element.MoveByVector is not wrapped in transaction


ksobon:
This happens: 

https://forum.dynamobim.com/t/warning-message-meaning/12095

![image](https://cloud.githubusercontent.com/assets/529781/26560952/5e16b37c-4488-11e7-9b9d-d5f24c6e8980.png)

PR #1692, opened 2017-05-29
fix to element.movebyvector that throws transaction errors

ksobon:
### Purpose

This addresses this issue: https://github.com/DynamoDS/DynamoRevit/issues/1691

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -718,12 +718,16 @@ namespace Revit.Elements
         /// Move Revit Element by Vector
         /// </summary>
         /// <param name="vector">Translation Vector</param>
-        public void MoveByVector(Vector vector)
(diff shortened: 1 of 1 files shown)


ksobon:
While I was at it, I made that method return the element that it was supposed to move. That's to prevent these nodes from becoming dead ends. I don't see why it should return `void`.

smangarole:
@sm6srw Can you please take a look at this commit when you get a chance? 
@QilongTang FYI.

Racel:
We are concerned that returning Element might cause an infinite loop. We ran into something similar when we were implementing Rotate. We will test on our machines before merging.

Or, if you have time, please let us know if you noticed any strange behavior.

Racel:
@ksobon - As per our discussion, we are closing this task. Please feel free to reopen at a later date.

PR #1694, opened 2017-05-30
fix to structuralFraming.location

ksobon:
### Purpose

This addresses the issue of a StructuralFraming.Location returning an Exception for vertical columns. They are point based columns, so the Location is a LocationPoint type object and we were casting to LocationCurve. That was fixed along with a Unit Test. 

Before: 

image

Now: 

image

### Declarations

### Reviewers

@ikeough @mjkkirschner


QilongTang:
LGT us, waiting on RTF results

ksobon:
@QilongTang any word on the RTF results?

test/Libraries/RevitNodesTests/Elements/StructuralFramingTests.cs
```diff
@@ -66,7 +66,16 @@ public void TestLocationOfStructuralFraming()
 
             var beam = e as StructuralFraming;
             Assert.NotNull(beam);
-            var curve = beam.Location;
+            var loc = beam.Location;
+
+            var pt = loc as Point;
+            if (pt != null)
+            {
+                Assert.NotNull(pt);
```
sharadkjaiswal: This Assert is not useful because of IF condition. I think you want to Assert.IsNull(pt) and Assert.IsNotNull(curve)

src/Libraries/RevitNodes/Elements/StucturalFraming.cs
```diff
@@ -247,17 +247,28 @@ private void InternalSetCurve(Autodesk.Revit.DB.Curve crv)
         /// <summary>
         /// Gets curve geometry from location of the specified structural element
         /// </summary>
-        new public Autodesk.DesignScript.Geometry.Curve Location
+        public new Autodesk.DesignScript.Geometry.Geometry Location
```
sharadkjaiswal: I am wondering why the base class AbstractFamilyInstance don't have an abstract property to return Geometry rather than Point? Not sure how new property behaves in VM, FFI.

test/Libraries/RevitNodesTests/Elements/StructuralFramingTests.cs
ksobon: well, yes I agree that the `Assert.NotNull(pt)` inside of `pt!=null` is a bit redundant, but at the same time the idea here is that this PR allows for location to return either `Point` or `Curve`. I guess that's what I was thinking here. It makes no sense to account for both in a Unit Test since we are specifically collecting a Beam element, and it will have guaranteed `Curve` Local point. I will remove the point part all together.

src/Libraries/RevitNodes/Elements/StucturalFraming.cs
ksobon: Mhmmm...good point. Should we update the `AbstractFamilyInstance` base implementation of the `Location` property instead? That will cover this example, but will open up a few more fixes in other classes that inherit from it...possibly.

Racel:
@mjkkirschner @QilongTang - Can you take a look and provide feedback to @ksobon about AbstractFamilyInstance? Would like to move forward with this PR. Also, added a task to the backlog for this: https://jira.autodesk.com/browse/QNTM-1183

mjkkirschner:
There are a bunch of files with conflict markers in this branch.
I agree with @sharadkjaiswal it's probably easier to improve the base class since we're not sure how the New keyword will effect the VM. @ke-yu any data on the c# new keyword?

ke-yu:
`new` is not handled by the VM.

mjkkirschner:
@ksobon seems like it would be better and less risky to modify the base class and avoid the new keyword since the impact on the VM is kind of unkown - does that sound okay?

mjkkirschner:
@ksobon there are 15,000 changes in this PR and there are conflict markers committed all over the place - I think something has gone wrong. 😢

Racel:
@ksobon - Any updates on this? If you do not have time to make the fixes, can you please close and reopen when they can be addressed?

PR #1701, opened 2017-06-01
Update test files for testing element binding with custom nodes

aparajit-pratap:
### Purpose

The element binding test, `MultipleCustomNodeInstance` in `ElementBindingTests.cs` had regressed on all Revit branches. Refer to [defect](https://jira.autodesk.com/browse/DYN-873) for details.

The reason for this was that the custom nodes could not bind to the Revit wall elements they had produced in the existing Revit file and were therefore creating new walls instead. It was found that the element binding fails because of the way the callsite ID's are generated. A callsite ID is generated for each element bound node (method call) and is serialized to the DYN file. When the file is reopened, a callsite ID is generated at runtime and if it matches with the one that is serialized with the node, the node can bind to its respective element successfully. In this case, the callsite ID generated in the latest master builds was found to be different from the one in the DYN file, resulting in the failure.

The reason is that the callsite ID is currently generated based on the class and function index of the function being called, which can vary depending on various factors like library loading sequence, etc. Since they are likely to vary over time the possibility of element binding to succeed is quite uncertain and random. We obviously need a better way of uniquely generating the callsite ID's that are independent of external factors such as library loading sequence, local Dynamo environment, etc so that they are always generated consistently across builds.

For the time being I am updating the test case to make it pass by regenerating new callsite ID's and trace data for the DYN workspace.

### Declarations

### FYIs

@ke-yu @riteshchandawar


QilongTang:
@aparajit-pratap Do you still recall the graph results? It is now different from when @ke-yu wrote the test, but I could not find a baseline.

image

aparajit-pratap:
@QilongTang the graph shown above indicates these nodes have shortest lacing but in the original XML based .DYN file the lacing is set to "auto". Could you check if the output is 8 walls when setting these nodes to `auto-lacing`. If so it could be that when you opened the XML DYN, the JSON deserialization didn't work as expected and switched auto-lacing to shortest? Have you tested these scenarios with JSON deserialization changes? @mjkkirschner?

mjkkirschner:
opening the XML DYN should still invoke the xml deserialization and migrations @QilongTang - I thought you recently fixed this bug with custom nodes though where custom nodes were not having migration applied to their nodes - but what about to the custom nodes themselves?

mjkkirschner:
@QilongTang also worth considering if RTF uses some other method to open graphs which does not invoke xml deserialization the same way the open command does.

QilongTang:
@aparajit-pratap @mjkkirschner I have already tried messing lacing around but all get same result which are 4 walls, that's why I am not sure what is the original expected behavior

PR #1707, opened 2017-06-05
Add the option to reuse the existing workspace model if the file we want to work with is already loaded

BogdanZavu:
### Purpose
Sometimes we want to reuse the existing workspace : performance related or other reasons.
In our case when we want to run a script , we want to perform a model check first and only if certain conditions are met we allow a script to run. But it's app to the third party app to decide on those conditions and that's why they can choose to open and then check the model on their end and after that post a run request (and they will probably choose to reuse existing workspace in this case).

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -153,8 +153,9 @@ namespace Dynamo.Applications
 
         /// <summary>
         /// The journal file can specify if a check should be performed to see if the
-        /// existing model already points to the Dynamo file we want to run or perform
-        /// other tasks. If that's the case, we want to use the existing model.
+        /// current workspaceModel already points to the Dynamo file we want to 
+        /// run (or perform other tasks). If that's the case, we want to use the
+        /// current workspaceModel.
         /// </summary>
         public const string DynPathCheckExisting = "dynPathCheckExisting";
 
@@ -571,7 +572,7 @@ namespace Dynamo.Applications
                     }
                 }
 
-                if (!useExistingWorkspace)
+                if (!useExistingWorkspace) //if use existing is false, open the specified workspace
                 {
                     if (ModelState == RevitDynamoModelState.StartedUIless)
                     {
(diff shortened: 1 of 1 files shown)


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -152,6 +152,13 @@ public class JournalKeys
         public const string ModelShutDownKey = "dynModelShutDown";
 
         /// <summary>
+        /// The journal file can specify if a check should be performed to see if the
+        /// existing model already points to the Dynamo file we want to run or perform
+        /// other tasks. If that's the case, we want to use the existing model.
```
mjkkirschner: can you clarify this comment a bit, when you refer to Model, please specify workspaceModel vs DynamoModel

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -551,16 +558,30 @@ private static void TryOpenAndExecuteWorkspaceInCommandData(DynamoRevitCommandDa
...
+                    WorkspaceModel currentWorkspace = revitDynamoModel.CurrentWorkspace;
+                    if (currentWorkspace.FileName.Equals(commandData.JournalData[JournalKeys.DynPathKey], 
+                        StringComparison.OrdinalIgnoreCase))
+                    {
+                        useExistingWorkspace = true;
+                    }
                 }
-                else
+
+                if (!useExistingWorkspace)
```
mjkkirschner: can you add a quick comment to this block - 
something like,  "if use existing is false, open the specified workspace"

mjkkirschner:
@BogdanZavu - looks good to me, only two small requests for comments.
Is there any reason we can't also add this to 2017 and 2017 1.3.1 incase users of those branches want to use this new journal key?

src/DynamoRevit/DynamoRevit.cs
BogdanZavu: done

src/DynamoRevit/DynamoRevit.cs
BogdanZavu: done

BogdanZavu:
@mjkkirschner I was wondering about this , technically no , but we want to make 2018 stuff in general more appealing so that we have more reason to move to subscription.

PR #1716, opened 2017-06-18
element pin/unpin and selection methods

ksobon:
### Purpose

This adds ability to select elements by their Id or Unique Id. It also adds ability to Pin/Unpin elements. 

image

### Declarations

### Reviewers

@mjkkirschner @jnealb 

### FYIs

@AndrewHeumann This seems to be part of the Wombat, Archi-lab and probably many more so instead of all having duplicate functionality let's add it to source.

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -173,6 +173,11 @@ namespace Revit.Elements
         }
 
         /// <summary>
+        /// Check if element is Pinned.
+        /// </summary>
+        public bool IsPinned => InternalElement.Pinned;
+
+        /// <summary>
         /// Returns the FamilyType for this Element. Returns null if the Element cannot have a FamilyType assigned.
         /// </summary>
         /// <returns name="ElementType">Element Type or Null.</returns>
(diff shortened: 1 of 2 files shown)


brianringley:
totally

kronz:
@ksobon The Element.ById and Element.ByUniqueId seem unambigiously awesome and we should have them out of the box.  The pin/unpin interaction is a little weird, can you break that out into a seperate PR and we can have a little discussion about what the right interaction should be?

ksobon:
@kronz it's a small PR. Let's keep it here and I can take care of both without added overhead of splitting it. What's weird about the Pin/Unpin node? It's really just one method so by using a boolean toggle you can either Pin or unpin the element.

ksobon:
@kronz so according to the newly discovered "rules", the pin/unpin should be two nodes where one is called `IsPinned` and another maybe `Pin`. What you think?

kronz:
@ksobon That makes sense to me . . . it's a bit more cumbersome in some ways, but it makes a little more sense in that it doesn' introduce a new interaction ("action").  Is there a use case you can think of where the pin/unpin action would need to be parametric? A stream of things coming in, and some get pinned and some not?  There's always edge cases, but generally seems like no, and in any case you could use FilterByBoolMask or a conditional.

Racel:
@ksobon @kronz - What is the state of this PR? Are we still waiting on changes from @ksobon? Are we splitting up this PR?

kronz:
I believe we agreed "the pin/unpin should be two nodes where one is called Pin another Unpin and then of course we have IsPinned that returns a boolean."  @ksobon ?

Racel:
@ksobon - Can you let us know what the state of this PR is? If you do not have time to address the issues above? Can you please close this PR and re-open at a later date?

PR #1717, opened 2017-06-18
Level.ByName selector

ksobon:
### Purpose

Adds a new level selector using Level Name. It's seems to be a very popular node amongst custom packages. 

image

### Declarations

### Reviewers

@mjkkirschner @jnealb 

### FYIs

@AndrewHeumann similar to Pin/Unpin functionality, it appears to me that this particular node must be a feature in more than just Wombat and Archi-lab. I know for a fact that Rhythm has one too. It's time to roll that into the source.


kronz:
@ksobon this is an interesting one, in that it points out a weird inconsistency in our naming.  "By" should be reserved for creation, so your name here indicates that it should be creating a Level, not finding an existing one.  Now, I know, I know, we use it ALL the time in existing nodes (eg FamilyType.ByFamilyAndName), that doesn't mean it wasn't a mistake.  But is there a better way we can express this action of finding something by a piece of information.  Level.GetByName?  Are you thinking this is something that lives in Selection?

ksobon:
I have no problem with `GetByName`

andydandy74:
@kronz I think it would be great to have all of those nodes in Selection and also to name them {ENTITY}.GetByName(). 

Browsing through the Revit nodes, I found these:
- Category.ByName
- CurtainSystemType.ByName
- Family.ByName
- FamilyType.ByName / FamilyType.ByFamilyAndName / FamilyType.ByFamilyNameAndTypeName (that's at least one node too many, by the way)
- _FillPatternElement.GetByName_ (using the naming scheme you suggested)
- FilledRegionType.ByName
- FloorType.ByName
- GlobalParameter.FindByName (close, but no cigar)
- _LinePatternElement.GetByName_ (using the naming scheme you suggested)
- Material.ByName
- ModelTextType.ByName
- RoofType.ByName
- TextNoteType.ByName
- WallType.ByName

And selecting by ID should live in the Selection category as well:
- Element.ById & Element.ByUniqueId (both in #1716) 
- PerformanceAdviserRule.ById (wouldn't that be obsolete with @ksobon 's new nodes?)

ksobon:
@kronz I don't think it should be in the Selection section of the nodes. I think it should stay where it is, under Level class. I consider selection section to be reserved for general purpose selection methods. This one is specific to Level elements. This is also how the current library is organized.

kronz:
@ksobon OK, let's do GetByName.  @Racel any objection?  Agreed that specific selection ("retrieve this thing I am naming from the document") should stay in the class, keeping the general implementations in Selection (although here we also have inconsistencies with UI nodes for specific slections for Wall Types, etc)

Racel:
@kronz @ksobon - How about "GetWithName"? In our naming convention spec, we specifically call out the word "By" as a signifier for a constructor.

ksobon:
@Racel It's all right as is. Let's get this merged if we can. Also I haven't yet seen "With", so I am not entirely sure why we need to introduce that...

Racel:
@ksobon - For 2.0 and moving forward, we have been working towards a cleaner, more expressive language. With that being said, I think that we should try and address the naming consistencies that this PR magnified before introducing new nodes. We are doing a considerable amount of node cleanup for 2.0

andydandy74:
@Racel no constructors in my list, only selection nodes.

kronz:
OK, I'm going to back up a little here and say we should take this PR as is.  This submission will only introduce inconsistency UNLESS we use the old, bad method.  That is, if we submit this function as Level.GetWithName, or .GetByName, or .FromName, it will be the odd exception.  Improving the general naming and interaction is a bigger, 2.0-ish project.  Level.ByName is a valuable incremental improvement that completely follows the existing (if confusing) model, and will be picked up for change and migration in any kind of future sweeping cleanup. 

Future task:  
Problem
As @Racel indicates, this is an ongoing issue that we should make cleaner with every submission, not muddier.  The problem, along with what the name is, is that the current approach doesn't really seem scalable: a different "selection" for every kind of element.  Are we going to make a .ByName for every kind of system family? This particular issue is very much an indicator of a broader issue, so it makes sense to discuss this as a model for later cleanup and corrections.  Finding a Revit element from string inputs, as opposed to creating or querying an element for properties, is a general Revit interaction for which we should establish a better model.  
Possible Solution:
Isn't this functionality really a query on the Document?  How about obsoleting all the individual .ByName selection nodes in favor of 2 selection nodes taking 2 inputs called something like FromTypeName and FromCategoryName to cover everything? All the individual .ByName nodes are essentially filtered outputs of our "All Elements of" selection nodes, right?

ksobon:
@kronz that's right! Why didn't we think of that approach earlier? One node to replace them all: 

image

Racel:
@kronz - I completely agree with the approach suggested. Much better solution and we can eventually get rid of all the individual type name  and category name nodes.

@ksobon - New nodes look good. Are you planning on changing this PR to include these new suggested nodes instead? Should we wait for those before review? 

Pinging @mjkkirschner @sharadkjaiswal  for their thoughts on the nodemodel vs zerotouch node approach. I don't have a strong opinion on either approach.

kronz:
@ksobon, we do want automatic update, so if only one approach gets that, let's use it

Racel:
@ksobon - Based off our last discussion, I believe you were going to make a new PR in Revit2018. Is this correct? If so, can you please close this and resubmit?

PR #1727, opened 2017-06-28
Add null check for Revit geometries that fail to convert to Dynamo geometries in Element.Geometry

aparajit-pratap:
### Purpose

Added null check for Revit geometries that fail to convert to Dynamo geometries in `Element.Geometry` node. This issue was pointed out by Dieter Vermeulen. 

The issue reported was when a door family when loaded into a project environment and selected using `Select Model Element` node was passed into `Element.Geometry`. `Element.Geometry` simply failed with a `null` value. The reason for this was that out of all the solids in the family (about 7), only one (the doorknob) failed to convert to Dynamo geometry. As a result a null reference exception was thrown and the entire family failed to convert. 

The fix is to percolate the exceptions for Revit geometry that fail to convert (display errors on the node), return `null`, and continue converting the rest of the geometries. Note that the resulting list of geometries will contain only those geometries that have converted successfully. The `null` values are filtered out.

### Declarations

### FYIs

@mjkkirschner


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -452,6 +452,9 @@ public object[] Geometry()
                     if (solid != null)
                     {
                         var geomObjs = solid.ConvertToMany();
```
mjkkirschner: curious if this fails, does it leave behind any ASM geometry in memory which is not disposed?

mjkkirschner:
@aparajit-pratap LGTM with one question.

aparajit-pratap:
Thanks @mjkkirschner for pointing it out. I've made the change.

PR #1729, opened 2017-06-29
Update library view customization with layout spec of Revit nodes

sharadkjaiswal:
### Purpose

This PR integrates Revit layout spec with new library UI in Dynamo. When DynamoView is loaded the DynamoRevit command tries to obtain the `ILibraryViewCustomization` service instance and updates it with a layout spec for Revit nodes. This also calls OnAppShutdown on the customization service when Revit application is being closed, so that CEF can be shut down properly else Revit doesn't close clean.

image

### Declarations

### Reviewers

@Benglin


sharadkjaiswal:
This PR can be merged only if LibraryReorg branch is merged to master.

PR #1768, opened 2017-08-28
Revit element selection nodes should serialize and deserialize properly in Json

alfarok:
### Purpose

[QNTM-1284](https://jira.autodesk.com/browse/QNTM-1284)

The purpose of this PR is to enable the serialization and deserialization of Revit element selection nodes in Json.  Previously Revit/Dynamo would freeze when attempting save a graph containing Revit selection nodes and the Json file is never written.  The selection node should also maintain the Revit element id that is referenced so it will automatically be recognized when reopening a graph.

Supplemental to [Dynamo PR #8131](https://github.com/DynamoDS/Dynamo/pull/8131)

giphy

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] Snapshot of UI changes, if any.

### Reviewers

(FILL ME IN) Reviewer 1  (If possible, assign the Reviewer for the PR)

PR #1774, opened 2017-09-06
add Newtonsoft reference to RevitNodesUI.csproj

alfarok:
### Purpose

Forgot to commit this reference in [1768](https://github.com/DynamoDS/DynamoRevit/pull/1768)

### Declarations

### Reviewers
@mjkkirschner @QilongTang

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitNodesUI.csproj
+++ b/src/Libraries/RevitNodesUI/RevitNodesUI.csproj
@@ -39,6 +39,10 @@
       <HintPath>$(PACKAGESPATH)\DynamoVisualProgramming.Core\lib\net45\DynamoUtilities.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.8.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
+    </Reference>
     <Reference Include="PresentationCore" />
     <Reference Include="PresentationFramework" />
     <Reference Include="ProtoGeometry">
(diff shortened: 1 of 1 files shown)


src/Libraries/RevitNodesUI/RevitNodesUI.csproj
```diff
@@ -39,6 +39,10 @@
       <HintPath>$(PACKAGESPATH)\DynamoVisualProgramming.Core\lib\net45\DynamoUtilities.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.8.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
```
mjkkirschner: you don't want to include the version number here - it should just be
`Newtonsoft.Json`

QilongTang: @mjkkirschner I think the path need to be specific, otherwise dll will not be found. Keith was referring to other project we already have I guess, like DynamoCore. You can check DynamoCore.csproj as well

mjkkirschner: look in your packages folder:
screen shot 2017-09-06 at 2 39 50 pm

the whole point of this wacky packages sym link thing is so that the folders do not have version numbers in them.

QilongTang: Yup you are right. DynamoCore is not equivalent to DynamoRevit project in this case.

mjkkirschner: another option here is that we depend on the version of nuget that dynamo core pulls instead of adding another direct reference to it. @alfarok thoughts on that?

alfarok: @QilongTang for some reason I have 2 versions in my package folder and assumed because the batch file pulled 8.0.3 that was the version I should reference

jnealb:
@alfarok  @mjkkirschner @QilongTang @smangarole @Racel Is this discussion relevant to the currently broken state of our internal builds?

QilongTang:
@jnealb Yes, part of it. This LGTM now

alfarok:
@jnealb I think it is several factors, there is also a language change PR that needs to be updated or reverted influencing core.  D4R references the beta NuGet packages to build so if there are any issues with core build D4R won't build properly either

PR #1826, opened 2017-10-26
Serialization RTF test to Re-Save all Json files

QilongTang:
Please Note:
1. Before submitting the PR, please review [How to Contribute to Dynamo](https://github.com/DynamoDS/Dynamo/blob/master/CONTRIBUTING.md)
2. Dynamo Team will meet 1x a month to review PRs found on Github (Issues will be handled separately)
3. PRs will be reviewed from oldest to newest
4. If a reviewed PR requires changes by the owner, the owner of the PR has 30 days to respond. If the PR has seen no activity by the next session, it will be either closed by the team or depending on its utility will be taken over by someone on the team
5. PRs should use either Dynamo's default PR template or [one of these other template options](https://github.com/DynamoDS/Dynamo/wiki/Choosing-a-Pull-Request-Template) in order to be considered for review.
6. PRs that do not have one of the Dynamo PR templates completely filled out with all declarations satisfied will not be reviewed by the Dynamo team.
7. PRs made to the `DynamoRevit` repo will need to be cherry-picked into all the DynamoRevit Release branches that Dynamo supports. Contributors will be responsible for cherry-picking their reviewed commits to the other branches after a `LGTM` label is added to the PR.

### Purpose

[QNTM-2235](https://jira.autodesk.com/browse/QNTM-2235) All RTF tests should be run against both xml and json file formats

PR to add a test case under SystemTests to open and resave all RTF XML DYN and DYFs into Json format.

New test can be run by RTF but timeout need to be increased from 12000 to 16000 because it is taking some time to open and re-save the ~260 Dyns. I also failed to use TestSource so I am doing everything in a single test
image

Jsons are created under [SystemTempFolder]/DynamoRevitJsons:
image

### Declarations

### Reviewers

@mjkkirschner @ramramps @ColinDayOrg 

### FYIs

@alfarok @smangarole

### DIFF
--- a/test/Libraries/RevitIntegrationTests/SerializationTests.cs
+++ b/test/Libraries/RevitIntegrationTests/SerializationTests.cs
@@ -38,7 +38,6 @@ namespace RevitSystemTests
                 // Open XMLs in new Dynamo Core which should take care of
                 // Migration but element binding might be lost
                 ViewModel.OpenCommand.Execute(filePath);
-                AssertNoDummyNodes();
 
                 // Get new file path under temp folder
                 var fileNameSameStructure = filePath.Split(new string[] { "\\test" }, StringSplitOptions.None).Last();
--- a/test/Libraries/RevitIntegrationTests/SerializationTests.cs
+++ b/test/Libraries/RevitIntegrationTests/SerializationTests.cs
@@ -12,36 +12,19 @@ namespace RevitSystemTests
     [TestFixture]
     class SerializationTests : RevitSystemTestBase
     {
-        /// <summary>
-        /// This function gathers all dyn and dyf in 
-        /// workingDirectory into paths array.
-        /// </summary>
-        /// <returns>Test files paths array</returns>
-        public object[] FindWorkspaces()
-        {
-            var config = RevitTestConfiguration.LoadConfiguration();
+        private string JsonFolderName = "DynamoRevitJsons";
 
-            //get the test path
-            workingDirectory = config.WorkingDirectory;
+        [Test, TestModel(@".\empty.rfa")]
+        public void SaveXMLTestFilesToJson()
+        {
+            // Get all old XML DYNs
             var di = new DirectoryInfo(workingDirectory);
             var fis = new string[] { "*.dyn", "*.dyf" }
             .SelectMany(i => di.GetFiles(i, SearchOption.AllDirectories))
             .ToArray();
+            string[] XmlFilePaths = fis.Select(fi => fi.FullName).ToArray();
 
-            //di.GetFiles("*.dyn, *.dyf", SearchOption.AllDirectories);
-            return fis.Select(fi => fi.FullName).ToArray();
-        }
-
-        protected string JsonFolderName = "DynamoRevitJsons";
-
-        [Test, TestModel(@".\empty.rfa"), TestCaseSource("FindWorkspaces")]
-        public void SaveXMLTestFilesToJson(string filePath)
-        {
-            ViewModel.OpenCommand.Execute(filePath);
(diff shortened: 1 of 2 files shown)


test/Libraries/RevitIntegrationTests/SerializationTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            foreach (var filePath in XmlFilePaths)
+            {
+                // Open XMLs in new Dynamo Core which should take care of
+                // Migration but element binding might be lost
+                ViewModel.OpenCommand.Execute(filePath);
+                AssertNoDummyNodes();
+
+                // Get new file path under temp folder
+                var fileNameSameStructure = filePath.Split(new string[] { "\\test" }, StringSplitOptions.None).Last();
```
mjkkirschner: this seems fine but I would use the system io path methods like combine if there is one available that suits your needs.

test/Libraries/RevitIntegrationTests/SerializationTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                System.IO.Directory.CreateDirectory(jsonFolder);
+            }
+
+            foreach (var filePath in XmlFilePaths)
+            {
+                // Open XMLs in new Dynamo Core which should take care of
+                // Migration but element binding might be lost
+                ViewModel.OpenCommand.Execute(filePath);
+                AssertNoDummyNodes();
```
mjkkirschner: does this fail for any tests?

test/Libraries/RevitIntegrationTests/SerializationTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                {
+                    File.Delete(jsonPath);
+                }
+
+                // Re-save dyn, if folder does not exist, create it
+                FileInfo fileInfo = new FileInfo(jsonPath);
+                if (!fileInfo.Exists)
+                    Directory.CreateDirectory(fileInfo.Directory.FullName);
+
+                ViewModel.SaveAsCommand.Execute(jsonPath);
```
mjkkirschner: does it make sense to verify that the file actually got saved here and some sanity check like it includes a some nodes and node views?

test/Libraries/RevitIntegrationTests/SerializationTests.cs
QilongTang: Yes, I think I might just remove it and let those fail since graph should not give same result when there are dummy nodes

test/Libraries/RevitIntegrationTests/SerializationTests.cs
QilongTang: Considering this is a single test handling re-save for all DYNs, I would not bail out here. Also did not want to introduce NewtonSoft dependency to D4R

test/Libraries/RevitIntegrationTests/SerializationTests.cs
QilongTang: Yeah, I did couple search and seems there is only way to get parent directory, `Path` class itself does not support split according to https://msdn.microsoft.com/en-us/library/system.io.path(v=vs.110).aspx. In this case, it can be nested pretty heavily so I chose to use `Split`

test/Libraries/RevitIntegrationTests/SerializationTests.cs
mjkkirschner: it's already there - we need a dep on json.net to add json constructors etc. But no need to check here if it doesn't make sense.

mjkkirschner:
LGTM

PR #1827, opened 2017-10-26
stop tessellating textElements and curveElements 2017

mjkkirschner:
### Purpose
https://jira.autodesk.com/browse/DYN-906

This PR removes some functionality from DynamoRevit. TextElements (tags, textNotes) and Curve Elements (DetailCurves,ModelCurves,CurveByPoints,ElementCurveReference) will no longer tessellate in the Background or Revit Geometry preview automatically - instead one will need to use a element.geometry or curves or curve node to extract the curves which will render them.

Any of these will work for curve elements, **but** to get the text from TextElements users will need to use TextNote.Curves as this type is not really a geometry type. Note in the image below that the Element.Curves node will return an empty list.

- [ ] I still need to investigate if there are any broken tests

screen shot 2017-10-26 at 2 54 29 pm

### Declarations

### Reviewers

@QilongTang @sm6srw 

### FYIs

@Racel @kronz

### DIFF
--- a/src/Libraries/RevitNodes/Elements/CurveElement.cs
+++ b/src/Libraries/RevitNodes/Elements/CurveElement.cs
@@ -123,19 +123,18 @@ namespace Revit.Elements
 
         #endregion
 
+        /// <summary>
+        /// Empty method - we don't want to tessellate curves automatically
+        /// but it seems we need this method to correctly import this library into Dynamo.
+        /// See description in base class.
+        /// </summary>
+        /// <param name="package"></param>
+        /// <param name="parameters"></param>
         [IsVisibleInDynamoLibrary(false)]
         public new void Tessellate(IRenderPackage package, TessellationParameters parameters)
         {
-            //Ensure that the object is still alive
-            if (!IsAlive) return;
-
-            this.Curve.Tessellate(package, parameters);
-
-            if (package.LineVertexCount > 0)
-            {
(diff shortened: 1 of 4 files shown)


src/Libraries/RevitNodes/Elements/CurveElement.cs
```diff
@@ -123,19 +123,18 @@ protected static void setCurveMethod(Autodesk.Revit.DB.CurveElement ce, Curve c)
 
         #endregion
 
+        /// <summary>
+        /// Empty method - we don't want to tessellate curves automatically
+        /// but it seems we need this method to correctly import this library into Dynamo.
```
QilongTang: It's already empty in the `abstract class Element`, is it because it's abstract method there so child has to implement?

mjkkirschner: I'm not sure, compilation is fine, but when you attempt to import the child class into the VM, it fails without this method implemented, might be the VM is more strict than c# regarding abstract classes?

QilongTang: Could be. Just feeling strange to have a public method not doing anything but also not abstract..

QilongTang:
Please make sure if text note related RTF tests updated. then LGTM

QilongTang:
Also we may want to release note or blog post this. So a previous and after graph difference snapshot on forum would be really nice.

mjkkirschner:
@QilongTang I ran through most rtf system and node tests, no failures so far, even fixed a test! - will cherry pick and then commit next week.

PR #1848, opened 2017-11-09
use fromexisting constructor for creating family from name via document

mjkkirschner:
### Purpose

when we create families using the byBame node this implies that the elements already exist in the document and are revit owned - use the from existing constructor to set the `RevitOwned` bool to false.

add a test which checks the total number of families before and after removing this node from the graph.

### Declarations

### Reviewers

@ColinDayOrg 

### FYIs

@QilongTang

### DIFF
--- a/test/Libraries/RevitIntegrationTests/FamilyTests.cs
+++ b/test/Libraries/RevitIntegrationTests/FamilyTests.cs
@@ -52,7 +52,7 @@ namespace RevitSystemTests
 
             RunCurrentModel();
 
-            //assert number of loaded families in document:
+            //get the number of loaded families in document:
             var fec = new Autodesk.Revit.DB.FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument);
             fec.OfClass(typeof(Autodesk.Revit.DB.Family));
 
@@ -66,7 +66,7 @@ namespace RevitSystemTests
             //now delete the node
             model.CurrentWorkspace.RemoveAndDisposeNode(node);
 
-            //now assert the families count is the same.
+            //get the number of loaded families in document after removing the family.byname node from the graph:
             var fec2 = new Autodesk.Revit.DB.FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument);
             fec2.OfClass(typeof(Autodesk.Revit.DB.Family));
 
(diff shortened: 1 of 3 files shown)


test/Libraries/RevitIntegrationTests/FamilyTests.cs
```diff
@@ -37,6 +40,45 @@ public void GetFamilyInstancesByType()
...
+            var model = ViewModel.Model;
+
+            string samplePath = Path.Combine(workingDirectory, @".\Family\GetFamilyInstancesByName.dyn");
+            string testPath = Path.GetFullPath(samplePath);
+
+            ViewModel.OpenCommand.Execute(testPath);
+
+            RunCurrentModel();
+
+            //assert number of loaded families in document:
```
ColinDayOrg: This comment says "assert", but there is no assert here. Not a big deal, but somewhat confusing. The same for the other comment below about asserting.

test/Libraries/RevitIntegrationTests/FamilyTests.cs
```diff
@@ -37,6 +40,45 @@ public void GetFamilyInstancesByType()
...
+
+            //now assert the families count is the same.
+            var fec2 = new Autodesk.Revit.DB.FilteredElementCollector(DocumentManager.Instance.CurrentDBDocument);
+            fec2.OfClass(typeof(Autodesk.Revit.DB.Family));
+
+            // obtain the family type with the provided name
+            var families2 = fec2.Cast<Autodesk.Revit.DB.Family>();
+            var newcount = families.Count();
+
+            Assert.AreEqual(oldcount, newcount);
```
ColinDayOrg: The comment about asserting should probably be here for clarity.

ColinDayOrg:
A couple of comments about the comments in the code (minor, change them if you agree, but I'm ok if it stays as it is), but other than that LGTM.

test/Libraries/RevitIntegrationTests/FamilyTests.cs
QilongTang: Agree

test/Libraries/RevitIntegrationTests/FamilyTests.cs
QilongTang: Can we add a summery of the test somewhere, e.g. sth like `This test check to ensure that deleting the family.byName node does not delete the family in Revit host document`

PR #1887, opened 2017-12-17
small fix to restore packages batch script to handle spaces in paths 2018

ksobon:
### Purpose

This slightly changes the restorepackages.bat file to handle spaces in file paths. 

### Declarations

### Reviewers

@mjkkirschner 

### FYIs

There was a [PR](https://github.com/DynamoDS/DynamoRevit/pull/1447) for this before. Somehow it regressed again.

### DIFF
--- a/src/restorepackages.bat
+++ b/src/restorepackages.bat
@@ -1,4 +1,4 @@
-@echo off
+﻿@echo off
 REM ***********************************************************************
 REM This bat file uses Aget: https://git.autodesk.com/Dynamo/Aget to 
 REM download the nuget packages from NuGet server or closest Artifactory server for 
@@ -13,6 +13,7 @@ setlocal EnableExtensions
 
 REM 1. set variable values
 set DynamoPackages=%~dp0\packages\_packages
+echo %DynamoPackages%
 set CurrentDir=%~dp0
 if %CurrentDir:~-1%==\ (
     set CurrentDir=%CurrentDir:~0,-1%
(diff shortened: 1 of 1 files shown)


src/restorepackages.bat
```diff
@@ -27,16 +28,17 @@ set NugetConfig=%ConfigDir%\dynamo-nuget.config
...
 REM    and save the replaced file as packages.aget
-set versionQuery=%NugetExe% list DynamoVisualProgramming.Core -prerelease -config "%NugetConfig%"
-for /F "tokens=2 delims= " %%F in ( '%versionQuery%' ) do (set LatestBeta=%%F)
-echo Latest pre-release version of "DynamoVisualProgramming.Core" package is "%LatestBeta%"
-
+set versionQuery=list DynamoVisualProgramming.Core -prerelease -config "%NugetConfig%"
+for /F "tokens=2 delims= " %%F in ( 'call "%NugetExe%" %versionQuery%' ) do (
+    set LatestBeta=%%F
+)
+echo Latest pre-release version of "DynamoVisualProgramming.Core" package is %LatestBeta%
```
QilongTang: Great, now since we are all here. @ksobon would you explain this change a bit and share your wisdom here with the Dynamo Team 😃

ksobon: @ellensi fixed this before, but I pulled the repo fresh this past weekend and was getting the same errors as before. I am keeping the repo on my google drive so the path to source file `%~dp0` needs to be able to handle spaces since mine looks like this: C:\Users\user\Google Drive\DynamoRevit....I didn't change much, but rather re-instated whatever @ellensi did before which was adding "" marks around variables inside of the loop so `"%NugetExe%"`. That was getting broken because we use `delims = "` which I believe was delineating the path variable at a space and causing issues....

QilongTang: @ksobon Got you. I looked at the history of this file and seems the fix did get in for Revit2017 branch but not the others, after we merging this. Would you mind cherry-pick this change to Revit 2019 branches as well? Let me know if you would like to do it. LGTM

PR #1888, opened 2017-12-18
Material wrapper

ksobon:
### Purpose

There is a wrapper class for the Material element but proper call was never added to ElementWrappper.cs utility and all materials get wrapped as UnknownElement if I am not mistaken. 

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
+++ b/src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
@@ -5,7 +5,7 @@ using View3D = Revit.Elements.Views.View3D;
 
 namespace Revit.Elements
 {
-    /// <summary>
+    /// <sumry>
     /// Element wrapper supplies tools for wrapping Autodesk.Revit.DB.Element types
     /// in their associated Revit.Elements.Element wrapper
     /// </summary>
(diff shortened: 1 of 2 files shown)


ksobon:
ok, let's try this again. Last time I added a ton of C#6 formatting so you guys rejected it. This time I am only adding what was necessary to make the material class wrap properly.

src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Revit.Elements
 {
-    /// <summary>
+    /// <sumry>
```
mjkkirschner: spelling :)

src/restorepackages.bat
```diff
@@ -1,4 +1,4 @@
-@echo off
+﻿@echo off
```
mjkkirschner: are these the same changes as in your other PR?

ksobon: This branch was created from the other one, since I couldn't get it to build without it. :-)

src/Libraries/RevitNodes/Elements/InternalUtilities/ElementWrapper.cs
ksobon: I didn't mean to touch that. My bad.

src/restorepackages.bat
ksobon: I guess @QilongTang approved the other fix, so once it's merged, it will just override this. We should be good.

QilongTang: I have just merged the other PR, maybe merging latest R2018 branch into this branch will make this PR cleaner?

ksobon: will do. thanks!

ksobon:
@mjkkirschner this is good to go.

mjkkirschner:
LGTM, can you add cherry pick branches?

PR #1904, opened 2018-01-08
Update failing Revit tests 

aparajit-pratap:
### Purpose

Updated 2 test cases post Dictionary changes in Dynamo:
1. The multiple custom node test baseline was [updated](https://github.com/DynamoDS/DynamoRevit/pull/1841) to return the correct element binding results but due to reasons mentioned in https://github.com/DynamoDS/DynamoRevit/pull/1701 and the library loading sequence being changed again (due to the addition of `Builtin.dll`), after the Dictionary changes, this test has regressed yet again. This PR restores the file back to its original version that was passing in 1.3, which is found to pass once again. This is a temporary fix until the bigger problem of Trace and callsite ID's isn't fixed.
2. The other DYN file was failing for a single value (being returned as `null`) for `List.Combine` with `Noise.Generate` function input from `SimplexNoise.dll`. This test has been marked as "Failure" for the time being as it is a very strange edge case that has a remote chance of being reproduced, which also makes for a hard fix currently.

### Declarations

### FYIs

@QilongTang @smangarole @Racel


mjkkirschner:
@aparajit-pratap is there a reason you did not keep this file as a JSON file?

aparajit-pratap:
@mjkkirschner there is no reason why this couldn't be kept as a JSON file. The only reason why I reverted the file to its original version was because I wanted the old trace data. It wasn't related to format reasons.

PR #1907, opened 2018-01-09
Update the master branch

ZiyunShang:
This PR would merge the branch RC1.3.3_Revit2019 to master branch.


mjkkirschner:
This does not seem correct- what changes do you want to bing to master from -1.3?

QilongTang:
@mjkkirschner The Revit team would like to use master branch in this repo to always align with Revit master branch, I feel like PR is too messy and does not mean too much. Should we simply delete master branch and create it based on RC1.3.3, and later for 2.0 we can do the same. This way it is easy and clean.

What do you think? @ZiyunShang @mjkkirschner

ZiyunShang:
@QilongTang 
I think your method is very well, if there is no other problem.

QilongTang:
@ZiyunShang Thanks, please close this PR and you can go ahead replace the master branch since you should have write access now

PR #1912, opened 2018-01-10
Fix tests

aparajit-pratap:
### Purpose

Correction to the previous PR #1904. The test updates are for similar reasons as mentioned in #1904. I'm marking one test as "Failure" as before and the other multiple custom node test has the baseline updated (both RVT and DYN files) since the callsite ID's had changed due to the addition of a new preloaded library, namely "Builtin.dll". Please refer to explanation in https://github.com/DynamoDS/DynamoRevit/pull/1701. This is probably the same reason why @lillismith's Recharge demo for element binding didn't work with custom nodes, but I could be wrong.

### Declarations

### Reviewers

@QilongTang 
@mjkkirschner note that the new DYN is a JSON file

### FYIs

@smangarole @pboyer


QilongTang:
Thanks, Looks Good

PR #1921, opened 2018-01-24
T shcha/cicd

tshcha:
### Purpose

Source code for CICD.
I'd like it to be included into RC1.3.3_Revit2009 as soon as possible because now I can't run all stages of Jenkinsfile script together to test it.
I need to run only first stage to clone repo, then copy .bat files, CS.prop, .env, .version there manually,  checkout .version file before run other 3 stages.
After Jenkinsfile, 3 .bat scripts, updated CS.props, .env and .version files are cloned from repo it will be possible to run all 4 stages together.

### Declarations

### Reviewers

Phil XIa

Aaron Tang


.version
```diff
@@ -0,0 +1 @@
+0.0.3
```
philxia1: according to current version in artifactory, we need to reset it to 0.0.5

.env
```diff
@@ -0,0 +1 @@
+BRANCH_NAME='master'
```
philxia1: why we need this file?

Jenkinsfile
```diff
@@ -0,0 +1,60 @@
...
+        bat batFolder + '\\' + 'BuildSolutions.bat'
+      } catch(err) {
+        echo "Dynamo or DynamoRevit Build Failed: $err"
+        currentBuild.result = 'FAILURE'
+        throw err
+      }
+    }
+
+    if (branch == 'master' || branch.indexOf('release') == 0) {
+      stage('Deployment to Artifactory') {
```
philxia1: please remove the "|| branch.indexOf('release') == 0" as we only support the master branch right now.

.env
tshcha: I've completed according to revit-model-browser. Branch name is set in external file there.
I'll avoid using of this file for d4r and set branch name in Jenkinsfile directly.

src/scripts/ci/Deploy.bat
```diff
@@ -0,0 +1,30 @@
...
+    IF EXIST %folder% (
+        rmdir /s /q %folder%
+    ) 
+
+    call mkdir %folder%
+    call xcopy /y /i /s "../Dynamo/bin/AnyCPU/release" "./%folder%/release"
+    call xcopy /y /i /s "../Dynamo/bin/AnyCPU/debug" "./%folder%/debug"
+    call xcopy /y /i /s "../Dynamo/bin/AnyCPU/debug" "./%folder%/slow_debug"
+    call "C:\Program Files\7-Zip\7z.exe" a -tzip "%folder%.zip" ".\%folder%\release" ".\%folder%\debug" ".\%folder%\slow_debug"
+    call "C:\DeployTool\art.exe" upload "%folder%.zip" "team-revit-generic/revit-d4r/" --url=https://art-bobcat.autodesk.com/artifactory --user= --password=
```
philxia1: why we still need to call art.exe here? I thought we need to remove it.

PR #1926, opened 2018-01-31
update DynamoRevit for Revit2019 to Revit 2020

ZiyunShang:
The Revit version is updated to Revit 2020, so that the DynamoRevit should change the output from 2019 to 2020. Some functions are deprecated in Revit 2019 and should be replaced in the new version.

### DIFF
--- a/src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
+++ b/src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
@@ -66,6 +66,12 @@ namespace Revit.Filter
         {
             Autodesk.Revit.DB.OverrideGraphicSettings filterSettings = new Autodesk.Revit.DB.OverrideGraphicSettings();
 
+            // the old functions SetCutFillColor, SetCutFillPatternId,
+            // SetProjectionFillColor and SetProjectionFillPatternId,
+            // are obsoleted and suggested by the documentation that will be removed and 
+            // replaced by SetCutForegroundPatternColor, SetCutForegroundPatternId,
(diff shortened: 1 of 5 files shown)


src/Libraries/RevitNodes/Filter/OverrideGraphicSettings.cs
```diff
@@ -67,8 +67,8 @@ internal OverrideGraphicSettings(Autodesk.Revit.DB.OverrideGraphicSettings inter
             Autodesk.Revit.DB.OverrideGraphicSettings filterSettings = new Autodesk.Revit.DB.OverrideGraphicSettings();
 
             // Apply Colors
-            if (cutFillColor != null) filterSettings.SetCutFillColor(ToRevitColor(cutFillColor));
-            if (projectionFillColor != null) filterSettings.SetProjectionFillColor(ToRevitColor(projectionFillColor));
+            if (cutFillColor != null) filterSettings.SetCutForegroundPatternColor(ToRevitColor(cutFillColor));
+            if (projectionFillColor != null) filterSettings.SetSurfaceForegroundPatternColor(ToRevitColor(projectionFillColor));
             if (cutLineColor != null) filterSettings.SetCutLineColor(ToRevitColor(cutLineColor));
```
philxia1: please add comments why we did these changes? like the old methods are obsoleted and the new methods are suggested by the documentation. Did you test them - manually or via RTFRevit?

ZiyunShang: @philxia1 I have added some comments to the replaced.

PR #1942, opened 2018-02-16
Update RevitWatchHandler with changes to IWatchHandler

aparajit-pratap:
### Purpose
Updates to RevitWatchHandler to match changes in IWatchHandler. Refer to: https://github.com/DynamoDS/Dynamo/pull/8487

### Declarations 

### Reviewers
@alfarok 

### FYIs

@jnealb


src/DynamoRevit/ViewModel/RevitWatchHandler.cs
```diff
@@ -46,28 +47,28 @@ private WatchViewModel ProcessThing(Element element, ProtoCore.RuntimeCore runti
...
             catch (Exception)
             {
-                return callback(data.Data, runtimeCore, tag, showRawData);
+                return callback(data.Data, preferredDictionaryOrdering, runtimeCore, tag, showRawData);
             }
         }
 
-        public WatchViewModel Process(dynamic value, ProtoCore.RuntimeCore runtimeCore, string tag, bool showRawData, WatchHandlerCallback callback)
+        public WatchViewModel Process(dynamic value, IEnumerable<string> preferredDictionaryOrdering, ProtoCore.RuntimeCore runtimeCore, string tag, bool showRawData, WatchHandlerCallback callback)
         {
```
alfarok: just curious why you pass an `IEnumerable` here and call `toList()` while just passing `List` elsewhere?

alfarok:
@aparajit-pratap just one question but otherwise LGTM

src/DynamoRevit/ViewModel/RevitWatchHandler.cs
aparajit-pratap: It's generally better practice to use `IEnumerable` (interfaces) in public facing API's so that if the concrete implementation changes from List to say Dictionary or something else, then client code need not be updated. However in this case it's not so useful. Usually private methods have concrete types like List, etc.

PR #1966, opened 2018-03-20
upgrade some obsolete properties and methods

ZiyunShang:
Some class properties and methods is obsolete. They are deprecated in Revit 2019 and removed in the Revit 2020, so that some test cases can't run successfully.


src/Libraries/RevitNodes/RevitNodes.csproj
```diff
@@ -57,11 +57,7 @@
     <Reference Include="ProtoGeometry">
       <HintPath>$(DYNAMOAPI)\ProtoGeometry.dll</HintPath>
       <Private>False</Private>
-    </Reference>
-    <Reference Include="RevitAPI">
-      <HintPath>$(REVITAPI)\RevitAPI.dll</HintPath>
-      <Private>False</Private>
-    </Reference>
+    </Reference>    
     <Reference Include="System" />
```
philxia1: why we change this project?

philxia1:
@AndyDu1985 please review the api changed in this PR. Thanks.

src/Libraries/RevitNodes/RevitNodes.csproj
ZiyunShang: There is two RevitAPI.dll reference in the same csproj, I think it could be delete one.

QilongTang: Sorry where is the other reference you mentioned? Which line in this file?

ZiyunShang: @QilongTang line 49-52 is the same as the line 61-64, this is repeated.

PR #2129, opened 2018-09-26
DO NOT MERGE DynamoRevit should load more specific version of libG using major minor build

mjkkirschner:
### Purpose

related to:
https://github.com/mjkkirschner/DynamoRevit.git

Two big changes are made:
* when we preload ASM pick the libG folder by using major minor and build version
* when we ask Dynamo for the geometryFactoryPath we should be able to specify the same version we preloaded.
* when revitSystemTestBase is setup it will preloadASM like normal DynamoRevit startup 
* when revitNodeTest creates a testConfig it will inject the revit version - or it will use the config file with some better fallback rules (want feedback on this first)

### TODO:
- [ ] This PR should not be merged until new nugets exist and this branch targets a new branch of dynamoCore. (2.02, but preferably 2.1)
- [x] cherry pick to 2017, 2018 , 2019
- [ ] cherry pick to 1.3.4 - 2016, 2017, 2018, 2019
- [x] cherry pick to 2.02 - 2017 2018 2019
- [x] verify RTF tests still load asm correctly
- [x] verify unit tests still load asm correctly 
- [x] update testConfiguration xml file with new variable (`RequestedLibraryVersion2`)

### Declarations

### Reviewers

Not ready for full review

### FYIs

@aparajit-pratap 
@kronz 
@philxia1

### DIFF
--- a/test/Libraries/RevitTestServices/TestServices.dll.config
+++ b/test/Libraries/RevitTestServices/TestServices.dll.config
@@ -2,6 +2,6 @@
 <configuration>
   <appSettings>
     <add key="DynamoBasePath" value=""/>
-    <add key="RequestedLibraryVersion2" value="224.4.0"/>
+    <add key="RequestedLibraryVersion2" value="hostdefault"/>
   </appSettings>
 </configuration>
--- a/test/Libraries/RevitTestServices/TestServices.dll.config
+++ b/test/Libraries/RevitTestServices/TestServices.dll.config
@@ -2,6 +2,6 @@
 <configuration>
   <appSettings>
     <add key="DynamoBasePath" value=""/>
-    <add key="RequestedLibraryVersion" value="Version223"/>
+    <add key="RequestedLibraryVersion2" value="224.4.0"/>
   </appSettings>
 </configuration>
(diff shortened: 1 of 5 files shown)


aparajit-pratap:
@mjkkirschner looks good.

mjkkirschner:
@aparajit-pratap @QilongTang another solution to this problem is to simply ignore the testConfig file during revit testing and always use the version of ASM which we find bundled with the current revit - that will be simpler and will require no dynamo core changes (can close those PRs) but it's less flexible, and I imagine there was some reason the test config exists?

test/Libraries/RevitTestServices/TestServices.dll.config
```diff
@@ -2,6 +2,6 @@
 <configuration>
   <appSettings>
     <add key="DynamoBasePath" value=""/>
-    <add key="RequestedLibraryVersion" value="Version223"/>
+    <add key="RequestedLibraryVersion2" value="hostdefault"/>
```
aparajit-pratap: Why has the key been renamed? Where is `hostdefault` defined?

mjkkirschner: The key has been added not renamed - the old key is parsed as a LibGVersion - while the new one is usually a `System.Version` hostdefault is a special token for letting the host decide what version to load- I've noted it in the summary for the `RequestedLibraryVersion2` property.  

That said I'm not really happy with this design, let's try to find some time tomorrow to talk. I think we can come up with something simpler.

aparajit-pratap:
@mjkkirschner what's the status of this PR?

mjkkirschner:
@aparajit-pratap this shouldn't get merged until this branch is brought up to 2.0.2 or 2.1. It's missing commits from the other branches.

PR #2131, opened 2018-10-02
Rc2.02r19 libGLoading with specific version

mjkkirschner:
### Purpose
RC2.02 2019
Cherry picks:
https://github.com/DynamoDS/DynamoRevit/pull/2129

- [x] This PR should not be merged until new 2.02 nugets exist.

### Declarations

### Reviewers
@aparajit-pratap
@qilongtang

### FYIs

@ziyunshang

### DIFF
--- a/test/Libraries/RevitTestServices/RevitNodeTestBase.cs
+++ b/test/Libraries/RevitTestServices/RevitNodeTestBase.cs
@@ -1,6 +1,7 @@
+using System;
 using System.IO;
 using System.Reflection;
-
+using Dynamo.Applications;
 using DynamoUnits;
 
 using NUnit.Framework;
@@ -43,7 +44,8 @@ namespace RevitTestServices
 
         protected override TestSessionConfiguration GetTestSessionConfiguration()
         {
-            return new TestSessionConfiguration(Dynamo.Applications.DynamoRevitApp.DynamoCorePath);
+            var asmLocation = AppDomain.CurrentDomain.BaseDirectory;
+            return new TestSessionConfiguration(Dynamo.Applications.DynamoRevitApp.DynamoCorePath, DynamoRevit.findRevitASMVersion(asmLocation));
         }
 
         private static void SetupTransactionManager()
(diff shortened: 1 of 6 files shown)


test/Libraries/RevitTestServices/RevitNodeTestBase.cs
```diff
@@ -43,7 +44,8 @@ public override void Setup()
 
         protected override TestSessionConfiguration GetTestSessionConfiguration()
         {
-            return new TestSessionConfiguration(Dynamo.Applications.DynamoRevitApp.DynamoCorePath);
+            var asmLocation = AppDomain.CurrentDomain.BaseDirectory;
+            return new TestSessionConfiguration(Dynamo.Applications.DynamoRevitApp.DynamoCorePath, DynamoRevit.findRevitASMVersion(asmLocation));
```
mjkkirschner: @QilongTang @aparajit-pratap this change is not decided on yet - please let me know what you think. This change makes it so all RevitNodeTests ignore the config file and always loads the host libG version (or should testing this takes a while as nugets must be generated).

I've also modified dynamo in my latest PR so that a simple alternative is to provide the config file but leave the version set to anything but a valid version - this results in the config scanning all installed asm products - on a test machine this will work in most cases.

The crux of the issue is between flexibility and robustness I think.

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -207,11 +207,14 @@ protected override void StartDynamo(TestSessionConfiguration testConfig)
                 var revitTestPathResolver = new RevitTestPathResolver();
                 revitTestPathResolver.InitializePreloadedLibraries();
 
+                //preload ASM and instruct dynamo to load that version of libG.
+                var requestedLibGVersion = DynamoRevit.PreloadAsmFromRevit();
```
mjkkirschner: during systemTests, just load the same libG as Revit

ZiyunShang:
Hi @mjkkirschner , does this need cherry pick to master branch after master upgrades to Dcore 2.0.2?

mjkkirschner:
Hi @ZiyunShang yes we'll need to cherry pick this or something very similar to the master branch once it is updated to use 2.0.2 or 2.1 - I'm hoping that if we target 2.0.2 on the master(2020) branch that even when we move to 2.1 as the DyanmoCore version we won't need to do anything else on the DyanmoRevit branch except for perhaps recompile as the apis should be very similar or the same....

I'm not sure 100% on that though.

There are a few blockers before we can move forward.
we can go over them at the meeting Wed.

PR #2203, opened 2018-10-14
upate common test files to 2018 2.0.2

mjkkirschner:
in an effort to avoid timeouts and speed up tests try updating common sample files (these are used in many RTF tests)


mjkkirschner:
@smangarole

PR #2224, opened 2018-10-18
update adaptive component by face test

mjkkirschner:
update test and test file which was found to have an issue in previous versions as well, the test was still passing but timing out - make the test more precise by checking the resulting geometry.

screen shot 2018-10-18 at 5 04 00 pm

### DIFF
--- a/test/Libraries/RevitIntegrationTests/AdaptiveComponentTests.cs
+++ b/test/Libraries/RevitIntegrationTests/AdaptiveComponentTests.cs
@@ -14,6 +14,7 @@ using DoubleSlider = CoreNodeModels.Input.DoubleSlider;
 using Revit.Elements;
 using Dynamo.Graph.Nodes;
 using RevitServices.Persistence;
+using System;
 
 namespace RevitSystemTests
 {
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@alfarok @smangarole PTAL

test/Libraries/RevitIntegrationTests/AdaptiveComponentTests.cs
```diff
@@ -181,6 +182,10 @@ public void CreateAdaptiveComponentsByPointsOnFace()
                 var adapValue = GetPreviewValueAtIndex(adapID, i) as AdaptiveComponent;
                 Assert.IsNotNull(adapValue);
             }
+            var volumeSumId = "f76b55af-52f0-451e-9819-2a51ce8a1afc";
+            var volume = GetPreviewValue(volumeSumId);
+            Assert.IsTrue(Math.Abs((double)volume - 13.39) < 0.1);
```
alfarok: Hey @mjkkirschner, can you elaborate a little on this assertion?  Also does `AdaptiveComponentsByPointsOnFace.dyn` get used for any other tests that may be influenced by the changes made to the graph?

alfarok:
@mjkkirschner Just a couple questions, otherwise LGTM

mjkkirschner:
@alfarok 1 .only used in this test.
2. the assertion checks that the volume of the generated adaptiveComponents is about 13 units^2. This is just a simple way of checking that valid geometry was generated by looking at the output of a single node's preview.

PR #2282, opened 2018-12-07
Revert "Merge pull request #1999 from ksobon/3dview-wrapper-2018"

ZiyunShang:
Reverts DynamoDS/DynamoRevit#2114

An API breaking change is made, the main 3d view classes have their base class changed. Any code that relied on the base class now breaks.
DynamoRaas functionality also now is broken because those nodes expected view3d types and now all 3d views have a base class of Abstract3dView.


AndyDu1985:
This fix is OK, but it is an API breaking change. We can file a new PR for this fix and merge it when new major, such as version 3.x, is ready.

ZiyunShang:
We would merge the 3dview-wrapper commits after version 3.x. Track it with link: https://github.com/DynamoDS/DynamoRevit/pull/2290 .

PR #2349, opened 2019-02-22
Judge whether Item have no elements or SelectedIndex is invalid and r…

ZiyunShang:
…eturn a null node

### Purpose

The nodes like "Wall Location" which implement abstract class "CustomGenericEnumerationDropDown", run error when they have no selected value. It should return a null value in that case.

### Declarations

### Reviewers

@AndyDu1985 @QilongTang 

### FYIs

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitDropDown.cs
+++ b/src/Libraries/RevitNodesUI/RevitDropDown.cs
@@ -63,13 +63,13 @@ namespace DSRevitNodesUI
             base.Dispose();
         }
 
-        public Boolean IsItemOrSelectedIndexValid(string itemValue = null, string selectedValue = null)
+        public Boolean CanBuildOutputAst(string itemValue = null, string selectedValue = null)
         {
             if(Items.Count == 0 || SelectedIndex < 0)
                 return false;
-            if (string.IsNullOrEmpty(itemValue) && Items[0].Name == itemValue) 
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -62,6 +62,17 @@ public override void Dispose()
             DynamoRevitApp.EventHandlerProxy.DocumentOpened -= Controller_RevitDocumentChanged;
             base.Dispose();
         }
+
+        public Boolean CanBuildOutputAst(string itemValue = null, string selectedValue = null)
```
AndyDu1985: Rename itemValue => itemValueToIgnore, selectedValue => selectedValueToIgnore, and add comments for this method.

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -110,13 +121,9 @@ protected override SelectionState PopulateItemsCore(string currentSelection)
 
         public override IEnumerable<AssociativeNode> BuildOutputAst(List<AssociativeNode> inputAstNodes)
         {
-            if (Items.Count == 0 ||
-                Items[0].Name == NO_FAMILY_TYPES ||
-                SelectedIndex == -1)
-            {
+            if(!CanBuildOutputAst(NO_FAMILY_TYPES,null))
```
AndyDu1985: The 'null' seems can be removed.

PR #2360, opened 2019-03-08
Upgrade revit system test base

ZiyunShang:
### Purpose

Update RevitSystemTestBase to use local config file Users\[UserName]\AppData\Roaming\Dynamo\Dynamo Revit\X.X\DynamoSettings.xml when run RTF Tests for DynamoRevit.

### Declarations

### Reviewers

@AndyDu1985 

### FYIs

@QilongTang @mjkkirschner

### DIFF
--- a/test/Libraries/RevitTestServices/RevitSystemTestBase.cs
+++ b/test/Libraries/RevitTestServices/RevitSystemTestBase.cs
@@ -23,10 +23,11 @@ using RevitServices.Transactions;
 using RTF.Applications;
 using SystemTestServices;
 using TestServices;
+using Dynamo.Configuration;
 
 namespace RevitTestServices
 {
-    public class RevitTestConfiguration
+   public class RevitTestConfiguration
     {
         /// <summary>
         /// Directory where test files are kept
@@ -133,7 +134,7 @@ namespace RevitTestServices
 
         }
     }
-
+   
     [TestFixture]
     public class RevitSystemTestBase : SystemTestBase
     {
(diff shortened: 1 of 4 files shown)


test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -229,9 +230,25 @@ protected override void StartDynamo(TestSessionConfiguration testConfig)
...
+                  Environment.SpecialFolder.CommonApplicationData),
+                  "Autodesk", "RVT " + commandData.Application.Application.VersionNumber, "Dynamo");
+
+                // Set Path Resolver's user data folder and common data folder with DynamoRevit runtime.
+                var pathResolverParams = new TestPathResolverParams()
+                {
+                   UserDataRootFolder = userDataFolder,
+                   CommonDataRootFolder = commonDataFolder
+                };
+                RevitTestPathResolver revitTestPathResolver = new RevitTestPathResolver(pathResolverParams);
```
AndyDu1985: Is it possible that direct get pathResolverParams inside the constructor of RevitTestPathResolver since that RevitTestPathResolver already has key word 'Revit'?

test/Libraries/RevitTestServices/RevitTestPathResolver.cs
```diff
@@ -6,6 +6,12 @@ namespace RevitTestServices
 {
     public class RevitTestPathResolver : TestPathResolver
     {
+        public RevitTestPathResolver() { }
```
AndyDu1985: There still has one place use the constructor 'RevitTestPathResolver' without a parameter, could you check if we can replace it with our new constructor and remove the constructor without a parameter?

ZiyunShang: It is not necessary to set this parameter in another place https://github.com/DynamoDS/DynamoRevit/blob/04ae617597d553a7e774eefad479a0060ea57f91/test/Libraries/RevitIntegrationTests/RegressionTests.cs#L48 . This is a DynamoRevit Node test, and I think it has no need to load some other local nodes.

PR #2361, opened 2019-03-14
Add two new ImportInstance nodes because of RevitAPI changes

ZiyunShang:
### Purpose

Because of Revit API changes, the ImportInstance should have another param to import Geometries into correct position.

### Declarations

### Reviewers

@AndyDu1985 

### FYIs

@QilongTang @mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodes/Elements/ImportInstance.cs
+++ b/src/Libraries/RevitNodes/Elements/ImportInstance.cs
@@ -34,7 +34,7 @@ namespace Revit.Elements
         }
 
         #region Private constructor
-      
+
         /// <summary>
         /// Constructor for ImportInstance
         /// </summary>
@@ -56,7 +56,7 @@ namespace Revit.Elements
         }
 
         #endregion
-              
+
         /// <summary>
         /// Initialize an ImportInstance element
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Libraries/RevitNodes/Elements/ImportInstance.cs
```diff
@@ -95,6 +106,46 @@ private void InitImportInstance(string satPath, XYZ translation = null)
             ElementBinder.SetElementForTrace(importInstance);
         }
 
+        /// <summary>
+        /// Initialize an ImportInstance element
+        /// </summary>
+        /// <param name="satPath"></param>
+        /// <param name="translation"></param>
+        /// <param name="view"></param>
+        private void InitImportInstance(string satPath, Revit.Elements.Views.View view, XYZ translation = null)
```
AndyDu1985: The code of private void InitImportInstance(string satPath, XYZ translation = null) and private void InitImportInstance(string satPath, Revit.Elements.Views.View view, XYZ translation = null) are similiar, could you simplify them?  InitImportInstance(string satPath, XYZ translation = null) can just call InitImportInstance(string satPath, Revit.Elements.Views.View view, XYZ translation = null) with a specific view parameter.
You can also simplify of the constructor of ImportInstance.

PR #2367, opened 2019-04-01
Update version to1 1

ZiyunShang:
### Purpose

Update version to 0.1.1 and create a new d4r package to Revit which fixed some bugs and improvements.

### Declarations

### Reviewers

@AndyDu1985 

### FYIs

@QilongTang @mjkkirschner


ZiyunShang:
Hi @QilongTang @mjkkirschner , is the function **AddNodeDirectory** available here? I didn't find any similar use case in Dynamo Tests, what do I need to pay attention to when using it?

Mikhinja:
Would this work with custom nodes, i.e. *.dyf files? Although we would categorize those as content for our package, it would still be nice to be able to test them.

QilongTang:
@ZiyunShang In Dynamo tests, we use `GetLibrariesToPreload()` to preload dlls, for dyfs most of test case could simply open it to get them in library. Are you using it right now for preloading revit nodes developed by other Revit team? If this method works for you, then feel free to use it.

ZiyunShang:
@QilongTang Thanks to your reply.
@Mikhinja I think this only work with dll files, but you can open dyf files in your test code to get them in library as Aaron said.

PR #2398, opened 2019-06-12
Resolve new dynamo core version & Refactory some HardCoded String

ZiyunShang:
### Purpose
1. Resolve DynamoCore new version 2.2.1.5204
2. Refactory some hardcoded string

### Declarations

### Reviewers

@AndyDu1985 

### FYIs

@QilongTang @mjkkirschner


QilongTang:
I don't think we need to do this.

QilongTang:
Wrong comment, I did not realize it is the Dynamo Reference you are updating, thanks for doing these

src/Libraries/RevitNodes/Elements/AbstractFamilyInstance.cs
```diff
@@ -92,9 +92,9 @@ public Point Location
         public override string ToString()
         {
             if (InternalFamilyInstance != null && InternalFamilyInstance.IsValidObject)
-                return string.Format("Family={0}, Type={1}", InternalFamilyInstance.Symbol.Name, InternalFamilyInstance.Name);
+                return string.Format(Properties.Resources.Family + "={0}, " + Properties.Resources.Type + "={1}", InternalFamilyInstance.Symbol.Name, InternalFamilyInstance.Name);
```
mjkkirschner: I worry a little about localizing the `toString()`'s of elements - this means that if someone's code is comparing toString() between revit languages - it will now fail, when it worked before...
⚠️

mjkkirschner: maybe not important, but definitely should be considered.

mjkkirschner: @Amoursol any idea if people do this?

ZiyunShang: toString() is used to show the results of this node, as the picture showed
image
I don't know if this needs improvement or not, what do you think of this?

mjkkirschner: Yep, I understand that the `toString()`  is used to show the output in preview and watch - but I consider this a part of the evaluated code since I know that users sometimes serialize their data using `stringFromObject` and `stringFromArray`, and not a part of the UI - though really I guess it is both.

I believe this design was intentional - the same way that the node names themselves are not localized - only descriptions and tooltips are  - to facilitate data and code staying static between different languages. 

@kronz any thoughts about this?

That's my opinion any way 😄 but up to you guys of course.

QilongTang: @ZiyunShang Thinking of the impact of all the `ToString()` changes as well as other changes in this PR. They will make a subset values in preview bubble localized comparing to the rest of the cases. This would make Dynamo node preview bubble working in a in consistent way, please consider testing your changes as a localized build in a localized Revit env and you should see users will get preview bubbles in mixed languages which might not be so great. I suggest we hold on committing to these as part of 2.2.1.

ZiyunShang: @QilongTang Yep, it seems like this commit would have a big impact on Dynamo, not only **ToString()**, but also the **private const string** like https://github.com/DynamoDS/DynamoRevit/blob/8dd8ee352097b348a8df259ded648731dddf6077/src/Libraries/RevitNodesUI/Selection.cs#L607 , this const string would be used by class SelectionBase<TSelection, TResult> in DynamoCore.
I will talk to the team about this commit, and now it seems that it is not a good time to make this change.

aparajit-pratap:
I agree with Mike and Aaron. Node preview strings should not be localized.

kronz:
Late to the party here, and sorry, but I'm having a little trouble following the thread, so sorry for any misunderstanding. Yes, we were very conservative on what was going to be localized, addressing as many features as possible as "code" and therefore not localized. Descriptions, errors, and menu items are basically all we localized. User created names in Revit would be preserved as they are written, but I think the feature you are looking at are hard coded Revit parameter names which are localized in Revit?  Is the question "should we get and set parameter names based on the Revit localized names or on the 'internal' API names?" If the answer is to change the existing behavior, I would suggest a) being very sure about it and doing some user testing and b) not do it mid-release but wait for a major update

src/Libraries/RevitNodes/Elements/AbstractFamilyInstance.cs
Amoursol: > @Amoursol any idea if people do this?

I've not personally used, nor been exposed to this way of working beyond one singular use case of a South Korean Revit file containing Korean characters. However, I've only worked in English speaking locations so it could very well be problematic for those working in a different language and collaborating globally, albeit these kind of collaborations typically set English as the working language.

We should be careful to make meaningful differences between Dynamo based on localization.

PR #2408, opened 2019-07-11
Add new Select node "All Elements of Category In View"

awbushnell:
### Purpose

Scripts are in the process of being internally developed to leverage the existing Path of Travel Nodes to generate Paths of Travel for various purposes. One of these scripts analyzes data that takes in exit points for a floor plan and the set of rooms in the plan. Then analyzes all possible paths from center points and corners to those exit points. Once paths are calculated, based on the start points and path elements, generated, it calculates what would be the longest distance to the closest exit, removing all other elements that don't satisfy the criteria.

In order to be able to ship this in the box, we require a built in node that allows for querying for the list of elements of a specific category AND contained within a specific view. 

The Node:

- Takes a View and Category as input
- Returns the list of elements of the specified category in the specified view. Empty list if none found.

Additionally, this pr consolidates all the sln files to a single one to facilitate moving from one version of Visual Studio to the next:

- Deleted all but "DynamoRevit.2017.sln.*"
- Renamed "DynamoRevit.2017.sln" to "DynamoRevit.All.sln" as well as associated DotSettings File. 
- Updated Readme.md

### Declarations

### Reviewers

@AndyDu1985
@QilongTang 
@ZiyunShang 

### FYIs

@jrichter2019


QilongTang:
@awbushnell We have similar challenge in Dynamo Repo but I personally have been using the same solution file in VS2019, from you experience,any downside doing that?

awbushnell:
@QilongTang Honestly, solution files are compatible from VS2012 on so I see no reason not to just have a single solution file, but I just rolled with the current setup...    Especially in repos that are solutions of all just cs projects, then it really makes little sense to have a bunch of version specific solution files.. a single sln is usable in VS2015 and beyond...

src/Libraries/RevitNodes/Elements/Views/View.cs
```diff
@@ -245,14 +246,42 @@ public Revit.Filter.OverrideGraphicSettings FilterOverrides(Revit.Filter.Paramet
...
-        #endregion
-
-        #region View Templates
-
         /// <summary>
-        ///     Checks if View is a View Template.
+        ///     Get the elements contained in the view of the specified category. 
         /// </summary>
-        /// <returns></returns>
+        /// <param name="category">Element Category</param>
```
AndyDu1985: There is an existing node called "All Elements of Category" under Revit->Selection, if we want to add the node, it should also under Revit->Selection for consistance, maybe just add another "All Elements of Category" node which take a view and a category as parameters.
And there is a package, archi-lab.net, which also provide a node Select.ByCategoryAndView. 
@QilongTang, do you have any suggestions?

test/Libraries/RevitNodesTests/Elements/Views/ViewTests.cs
```diff
@@ -36,6 +38,31 @@ public void ExportAsImage_BadArgs()
...
+            Assert.Greater(elements.Count, 1);
+        }
+
+        [Test, TestModel(@".\Empty.rvt")]
+        public void ElementsByCategory_EmptyList()
+        {
+            var testView = CreateTestView();
+            var cat = Category.ById((int)Autodesk.Revit.DB.BuiltInCategory.OST_Rooms);
+            var elements = testView.ElementsByCategory(cat);
+            Assert.Equals(elements.Count, 0);
```
ZiyunShang: Here Assert.Equals should not be used for Assertions. Should replace it with Assert.AreEqual like
Assert.AreEqual(0, elements.Count);

src/Libraries/RevitNodes/Elements/Views/View.cs
awbushnell: @AndyDu1985  OK, after posting this it came to my attention the possibility of what you suggest. FYI, this PR's purpose is to provide an internal alternative to the ARchilab node as our team needs it internal to ship a script inside the box of revit.. I will go forwarard with your proposed changes.

test/Libraries/RevitNodesTests/Elements/Views/ViewTests.cs
awbushnell: @ZiyunShang OK, just for my own clarification, is this a new requirement for tests? I based my added tests of existing.

awbushnell:
@QilongTang  I consolidated the sln files as we discussed. @ZiyunShang  can you verify CI builds know to look for the new name. The JenkinsFile did not reference in anyway.

QilongTang:
@awbushnell Very nice, this actually aligns with how Dynamo repo manage sln file. I am also open for fixing the CPU architecture setting in both repo if the current one make less sense.

awbushnell:
@QilongTang With regards to the CPU setting.. Currently DynamoRevit builds as "AnyCPU" however, Revit builds specificaly for "X64".   It seems/feels to me that since DynamoRevit build against/for Revit that it should probably be X64 as well. This would avoid warnings and I think makes sense to be consistent. 

@AndyDu1985  Related to this, I notice the DynamoForRevit packages contains at least 1 x86 folder which seems to make no sense as part of Revit or in general... could we look at cleaning up the package download etc. to reduce footprint in revit?

test/Libraries/RevitNodesTests/Elements/Views/ViewTests.cs
ZiyunShang: This Assert.Equals throws exception “Assert.Equals should not be used for assertions”.
It would be successful that use Assert.AreEqual instead.

awbushnell: @ZiyunShang Correct, the node was moved to a different node and thus the other node removed and these tests no longer were valid.

AndyDu1985:
@awbushnell Yes, I will look at cleaning up.  Did you add unit tests for your added node?

AndyDu1985:
You can use RevitTestFramework to run your unit tests, https://github.com/DynamoDS/RevitTestFramework

awbushnell:
@AndyDu1985  thanks, I will add unit tests and run. Once run successfully, I will push to this PR and check off the "tested" item in the checklist.. I am waiting on the new icons before I post screenshots and check off UI changes.

awbushnell:
@QilongTang @AndyDu1985 @ZiyunShang @mjkkirschner I have uploaded a system test for this as well as updated the solution naming etc. to be consistent with Dynamo conventions..

PR #2412, opened 2019-07-18
Add Element Delete Node

awbushnell:
### Purpose

Scripts are in the process of being internally developed to leverage the existing Path of Travel Nodes to generate Paths of Travel for various purposes. One of these scripts analyzes data that takes in exit points for a floor plan and the set of rooms in the plan. Then analyzes all possible paths from center points and corners to those exit points. Once paths are calculated, based on the start points and path elements, generated, it calculates what would be the longest distance to the closest exit, removing all other elements that don't satisfy the criteria. 

In order to be able to ship this in the box, we require a built in node be added that allows deletion of an element from the model and this submission is adding that built in module for use. 

The Node:

- Will take the element to delete as input
- Will return the list of element ids' deleted.

### Declarations

- [] Snapshot of UI changes, if any.

### Reviewers

@AndyDu1985 
@ZiyunShang 

### FYIs

@jrichter2019


QilongTang:
@awbushnell Curious about the interaction. Graph containing this node only going to run once right because second time running, the element is gone?

test/Libraries/RevitNodesTests/Elements/ElementTests.cs
```diff
@@ -46,6 +46,21 @@ public void SetParameterByName_Element_CanSuccessfullySetMaterialByElement()
...
+        {
+            // Get element from document
+            var wall = ElementSelector.ByElementId(184176, true);
+            Assert.IsNotNull(wall);
+
+            // Delete element, confirming deletion success
+            Assert.IsTrue(wall.Delete());
+
+            // Verify element really deleted.
+            Assert.IsFalse(wall.Delete());
```
ZiyunShang: This Double call of Delete() would throw an exception before the Assert.IsFalse called, then this test failed.
I think you can use RTF(https://github.com/DynamoDS/RevitTestFramework) to run the tests to make sure your tests is reliable. You can install the RTF or build by yourself.

ZiyunShang:
I just found this node also appeared in the archi-lab.net package, I am still confused about its boolean output - after run, the elements was deleted, and this output always keeps empty List. It looks strange to me..

awbushnell:
@ZiyunShang  @QilongTang I have updated the purpose to this PR to hopefully clarify why we need this BUILT IN node to delete an element.  We are looking to ship a script but to do so, it cannot depend on any 3rd party nodes, our script generates a bunch of Paths Of travel then after generation of all those elements, based on some criteria, we delete the ones that don't match allowing for the script to give the designer information for the longest path to the closest exit.. Thus the delete in question would be on elements that each run of the script generates... and yes, archilab has something like this, but we cannot depend on it and it seems like a basic functionality that should be built in anyway.. IMHO.

test/Libraries/RevitNodesTests/Elements/ElementTests.cs
awbushnell: @ZiyunShang Good points. I am going to update the node api as follows for consistency with the internal revit api and some "safety" regarding the element getting deleted:

- I am going to make it a "static" routine so that there upon return from the delete, you are not in a member method and thus potential for issues accessing object data...
- match the internal api and return a list of the delete element Id's, or empty list upon failure to remove.

QilongTang:
@awbushnell I have very less concern about adding such built-in node, I think this should be a great plus. The fact archi-lab include this node indicates usage is there for sure. Moving away from boolean output makes sense to me, can you also show maybe an example how you would delete a list of elements?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -308,6 +308,40 @@ public void Tessellate(IRenderPackage package, TessellationParameters parameters
...
+            
+            // Start the transaction
+            TransactionManager.Instance.EnsureInTransaction(document);
+
+            // Delete the element
+            try
+            {
+                document.Delete(this.InternalElementId);
+            }
+            catch(Autodesk.Revit.Exceptions.ArgumentException)
```
mjkkirschner: is it worth also rethrowing these exceptions so users see the exception messages on the nodes?

QilongTang: @mjkkirschner Can you give an example, or it is as simple as setting the node warning?

mjkkirschner: https://github.com/DynamoDS/DynamoRevit/blob/master/src/Libraries/RevitNodes/Elements/DirectShape.cs#L369

see this example for directshape - any exception that gets throw (in general) will be used to set the warning state on a node - it will turn yellow and the error popup will display the exception type and inner message.

awbushnell: @mjkkirschner OK, so for example, if I just let the exceptions propagate out? (i.e. I don;'t need to catch.. I just need to add finally for transaction work for example, but if the exceptions occur, they will propagate out?

mjkkirschner:
@awbushnell I think ideally this PR would also include a system test that includes a .dyn file using the node, this is very useful for users who are trying to understand how to use the node or when other team members who are not familar with the functionality in Revit need to see how to use it.

heres an example:
https://github.com/DynamoDS/DynamoRevit/blob/master/test/Libraries/RevitIntegrationTests/DirectShapeTests.cs#L45

note how it opens the dynamo file as well as the revit file.

src/Libraries/RevitNodes/Elements/Element.cs
mjkkirschner: on second thought - let me show another example - it seems to me you want to return a valid output - this can also be done while still putting the node into a warning state -with a different API. one sec.

mjkkirschner: here is an example of using the `OnLogWarningMessage` method to put the node into a warning state while still returning a valid output:

https://github.com/DynamoDS/Dynamo/blob/ec10f936824152e7dd7d6d019efdcda0d78a5264/src/Libraries/DSOffice/Excel.cs#L427

not required, just something that is possible if you want both. This method is part of the DynamoServices.dll which DynamoRevit is likely already referencing.

mjkkirschner: when you throw an exception from a zero touch node,  it will propagate up to Dynamo and get caught, and the node will go into an error state and return `null` - if you want to return a valid output - you don't want to throw, and can continue the pattern you are using :).

If you also want to let the user know why the element was not deleted you can set the node in an error state but still return your valid output of`false`using that logWarning method.

awbushnell: @mjkkirschner thank you. In hindsight, the pattern of returning an empty list in this probably not ideal in this case. The delete should delete and if it doesn't it is best to throw rather than have the script continue along without having done so, regardless of an empty list being returned etc.

awbushnell:
@mjkkirschner I have seen the examples of system tests with rvt and dyn files etc. Still getting my head around dyn files as a whole, but I face issues trying to create rvt files etc. as I work on Revit proper and I don't install revit so files generated by my builds would not be anything I would want to submit as I am not clear how up gradable they would be etc. It seems to me most of the rvt files present now 2016 based so up gradable from there forward.

mjkkirschner:
@awbushnell - yep, makes sense, I think you would want to submit 2020 files if possible, though you might get away with just resuing one of the existing .rvt files already in the repo, and use your node to delete some element from there, then assert it's gone.

awbushnell:
@QilongTang @mjkkirschner Updates pushed based on feedback. Still working on sample dyn and running of test framework, but I believe the changes pushed map to our discussions.

awbushnell:
> @QilongTang @mjkkirschner Updates pushed based on feedback. Still working on sample dyn and running of test framework, but I believe the changes pushed map to our discussions.

Hold on.. ugg, my git client did a commit all.. and added some local build files.. I am removing those.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -308,6 +308,46 @@ public void Tessellate(IRenderPackage package, TessellationParameters parameters
             // or transactions and which must necessarily be threaded in a specific way.
         }
 
+        /// <summary>
+        /// Delete the element and any elements that are totally dependent upon the element.
```
QilongTang: Thank you for the more detailed summary

awbushnell:
@mjkkirschner  Pardon if this should be obvious, but I am having problems figuring out to test my generate .dyn file for Delete Element. As suggested basing it off of DirectShape, but I notice that the tests there all GetPreviewValeue(..).. So where do I get the "guids" for those test nodes?  FWIW, the .dyn I have is one I simply ran dynamo with a sample project open and dropped in nodes to call delete..

ZiyunShang:
@awbushnell I think you can get the "guids" from the dyn file - Open the file with text editor, in 2.x version, it would be 
      "ConcreteType": "Dynamo.Graph.Nodes.ZeroTouch.DSFunction, DynamoCore",
      "NodeType": "FunctionNode",
      "FunctionSignature": "Autodesk.DesignScript.Geometry.Point.ByCoordinates@double,double,double",
      "Id": "0258597ad1b343bf94bce495604e0667",
      ...
The "Id" is what you want.
In previous version, it may be 
https://github.com/DynamoDS/DynamoRevit/blob/fdb7311bdface067a2741b3910c93efab8e493f8/test/System/AdaptiveComponent/AdaptiveComponent.dyn#L4

AndyDu1985:
@awbushnell "some internal scripts are being worked on to ship with Revit that require this functionality", could you share those internal scripts with us?

awbushnell:
@ZiyunShang @mjkkirschner @AndyDu1985 I have added both a system test and node test that I ran successfully via the test framework...

johnpierson:
Also wanted to add that `Spring Nodes` has this functionality as well. It is called `Springs.Doc.DeleteElements`. From what I remember, package authors added this early on as people wanted to be able to delete junk in a Revit file. At the time the team's stance was "We do not want to allow this because it might be giving people a dangerous node that they do not understand completely".

PR #2439, opened 2019-10-16
Preload Steel Connections for Dynamo package from Revit

vpavel04:
Preload Steel Connections for Dynamo package from Revit

### DIFF
--- a/src/DynamoRevit/RevitPathResolver.cs
+++ b/src/DynamoRevit/RevitPathResolver.cs
@@ -55,6 +55,19 @@ namespace Dynamo.Applications
             // Add the Revit_20xx folder for assembly resolution
             additionalResolutionPaths = new List<string> { currentAssemblyDir };
 
+			      // Add Steel nodes to preload
+			      var steelNodesDirectory = Path.Combine(currentAssemblyDir, "nodesSteelConnections", "steel-pkg", "bin");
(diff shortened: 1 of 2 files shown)


vpavel04:
I believe that Dynamo4Revit should have an Addin or an Extension mechanism to preload additional nodes and to extend LayoutSpecs.json properly. For the moment, I believe this is the best solution until Dynamo4Revit will be able to dynamically discover nodes to preload

src/DynamoRevit/RevitPathResolver.cs
```diff
@@ -55,6 +55,19 @@ internal RevitPathResolver(string userDataFolder, string commonDataFolder)
             // Add the Revit_20xx folder for assembly resolution
             additionalResolutionPaths = new List<string> { currentAssemblyDir };
 
+			// Add Steel nodes to preload
+			var steelNodesDirectory = Path.Combine(currentAssemblyDir, "nodesSteelConnections", "steel-pkg", "bin");
```
QilongTang: Is the format here broken or it's just a github display issue?

vpavel04: i've had tabs instead of spaces. I've fixed it

QilongTang:
@AndyDu1985 LGTM, what do you think?

vpavel04:
i will come back with the generic mechanism

PR #2441, opened 2019-10-17
add functionality to dynamically preload internal packages

vpavel04:
### Purpose

We need to preload Steel Connections for Dynamo package from Revit. This new functionality will load all internal nodes placed in subfolders from [RevitInstallerfolder]\Addins\DynamoForRevit\Revit\packages\. 

Every internal package needs to create an extension.xml file where it will specify the path to the node and the path to the layoutspecs.json file.

Here is a sample of extension.xml:
```xml
<ExtensionInfo>
  <NodePath>.\bin\AdvanceSteelConnAutoNodes.dll</NodePath>
  <LayoutSpecsPath>.\LayoutSpecs.json</LayoutSpecsPath>
</ExtensionInfo>
```

### Declarations

### Reviewers

@AndyDu1985
@QilongTang

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -913,15 +913,25 @@ namespace Dynamo.Applications
         }
     }
 
+    /// <summary>
+    /// Defines parameters used for loading internal Dynamo Revit packages
+    /// </summary>
     [Serializable()]
-    public class ExtensionInfo
+    public class InternalPackage
     {
+        /// <summary>
+        /// keeps the path to the node file
+        /// </summary>
         public string NodePath { get; set; }
+
+        /// <summary>
+        /// keeps the path to the layoutSpecs.json file
+        /// </summary>
         public string LayoutSpecsPath { get; set; }
     }
     internal static class DynamoRevitInternalNodes
     {
-        private static IEnumerable<string> GetAllExtensionFiles()
+        private static IEnumerable<string> GetAllInternalPackageFiles()
         {
             string currentAssemblyPath = Assembly.GetExecutingAssembly().Location;
             string currentAssemblyDir = Path.GetDirectoryName(currentAssemblyPath);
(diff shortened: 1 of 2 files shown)


QilongTang:
This looks like a flexible idea in general, I wonder if we want to make it part of DynamoCore FYI: @mjkkirschner

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -890,4 +912,85 @@ public override IEnumerable<string> GetDynamoUserDataLocations()
             return paths;
         }
     }
+
+    [Serializable()]
+    public class ExtensionInfo
```
mjkkirschner: this name seems to conflict with the already existing `Extension` and `ViewExtension` mechanisms that dynamo provides for loading application level plugins.

* Also, does this need to be public?
* public properties should have /// summary comments that describe their use etc.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -890,4 +912,85 @@ public override IEnumerable<string> GetDynamoUserDataLocations()
...
+        public string LayoutSpecsPath { get; set; }
+    }
+    internal static class DynamoRevitInternalNodes
+    {
+        private static IEnumerable<string> GetAllExtensionFiles()
+        {
+            string currentAssemblyPath = Assembly.GetExecutingAssembly().Location;
+            string currentAssemblyDir = Path.GetDirectoryName(currentAssemblyPath);
+
+            string internalNodesDir = Path.Combine(currentAssemblyDir, "packages");
```
mjkkirschner: so this assumes a packages folder inside of Dynamo For Revit? or wherever DynamoRevit.dll is loading from?

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -890,4 +912,85 @@ public override IEnumerable<string> GetDynamoUserDataLocations()
...
+            {
+                return new List<string>();
+            }
+
+            string[] internalNodesFolders = Directory.GetDirectories(internalNodesDir);
+
+            List<string> extensionFiles = new List<string>();
+            foreach (string dir in internalNodesFolders)
+            {
+                string extensionFile = Path.Combine(dir, "extension.xml");
```
mjkkirschner: again, extension seems confusing given the existing and completely orthogonal extension functionality.

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -890,4 +912,85 @@ public override IEnumerable<string> GetDynamoUserDataLocations()
...
+                        {
+                            extInfo.LayoutSpecsPath = Path.Combine(extensionFileDir, extInfo.LayoutSpecsPath);
+                        }
+
+                        extensionInfos.Add(extInfo);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Console.WriteLine(string.Format("Exception while trying to parse extension file {0}", extensionFile));
```
mjkkirschner: if this will be shown to users it should live in a .resx - if it's just for internal devs, than this should be fine.

src/DynamoRevit/DynamoRevit.cs
QilongTang: To be consistent I would suggest putting them under `Revit\nodes` instead of under Revit folder, also can we get rid of the `packages` folder since it is now no longer a package?

mjkkirschner:
Hi All,

some thoughts on this approach:

* when the new library was designed it was intentional that there would be a split between built in nodes and add-ons (packages, imported libraries). I see here we are trying to address the more subtle use case of internal packages.... This seems tricky, how do we maintain this functionality for internal tools but didscrourage its use for external packages?
* Let's make sure to add tests.
* Extension is now a very overloaded name, we should not re-use it for a completely different functionality.
* Are packages the best mechanism to attach this additional layout spec to?

src/DynamoRevit/DynamoRevit.cs
AndyDu1985: Is the folder "[Revit]\Addins\DynamoForRevit\Revit"?

src/DynamoRevit/DynamoRevit.cs
AndyDu1985: Extension seems too generic, I think what we are doing here is to pre load customized packages.

src/DynamoRevit/DynamoRevit.cs
vpavel04: i updated it to use "[Revit]\Addins\DynamoForRevit\Revit\nodes

src/DynamoRevit/DynamoRevit.cs
vpavel04: It needs to be public for XmlSerializer serializer = new XmlSerializer(typeof(InternalPackage));.
I renamed it to InternalPackage and added some comments

src/DynamoRevit/DynamoRevit.cs
vpavel04: this is only for internal devs

vpavel04:
Hi all,

Thank you for your valuable feedback. I agree to you that the Extension name was to overloaded. 
I've renamed it to InternalPackage wich is more related to the problem that we want to solve. It is used to load the internal Revit nodes and include them in Revit UI category.

src/DynamoRevit/DynamoRevit.cs
vpavel04: renamed it

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -890,4 +912,95 @@ public override IEnumerable<string> GetDynamoUserDataLocations()
...
+        public string LayoutSpecsPath { get; set; }
+    }
+    internal static class DynamoRevitInternalNodes
+    {
+        private static IEnumerable<string> GetAllInternalPackageFiles()
+        {
+            string currentAssemblyPath = Assembly.GetExecutingAssembly().Location;
+            string currentAssemblyDir = Path.GetDirectoryName(currentAssemblyPath);
+
+            string internalNodesDir = Path.Combine(currentAssemblyDir, "nodes");
```
QilongTang: Would you split `nodes` out to be a const string?

QilongTang:
After looking at the solution again, this seems a really specific solution to D4R which is great. After rethinking a little bit, I believe although DynamoCore may also need similar functionality in the future, it is less likely any host specific packages will be shipped with Dynamo nodes together for now. So other than providing a common interface in Dynamo which we can always do later, the current solution looks good. What do you think @mjkkirschner @AndyDu1985

src/DynamoRevit/DynamoRevit.cs
vpavel04: done

AndyDu1985:
@mjkkirschner @QilongTang @ZiyunShang Do you have any more comments?

QilongTang:
@AndyDu1985 Nope, LGTM

PR #2450, opened 2019-11-07
Element.Geometry should provide geometry for FamilySymbols with no instances

Dewb:
### Purpose

Element.Geometry node returns no result for FamilySymbol elements if the FamilySymbol is not considered "activated" (generally, when it has no instances in the document.) Graph authors have to implement workarounds with temporary elements to guarantee they can get geometry. 

The [Revit API guidance on this issue](https://www.revitapidocs.com/2015/69706e35-87cc-a6d9-68fe-90a41c1c48db.htm) is fairly straightforward, and Element.Geometry already triggers a document regen, so this is something that likely can be safely handled without the user becoming aware of it. (However, there will likely be performance impacts when activating large numbers of unplaced FamilySymbols.)

Before:
d4r-familysymbol-before

After:
d4r-familysymbol-after

### Declarations

- [X] Snapshot of UI changes, if any.

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -532,6 +532,19 @@ namespace Revit.Elements
                 }
             }
 
+            // FamilySymbol is a special case: the symbol must be activated before geometry is available, unless an instance is already present in the model
+            else if ((thisElement is FamilySymbol) && (geomElement == null || !geomElement.Any()))
+            {
+                var fs = (FamilySymbol)thisElement;
+                if (!fs.IsActive)
+                {
+                    fs.Activate();
+                    DocumentManager.Regenerate();
+                    geomElement = fs.get_Geometry(goptions0);
+                }
+
+            }
+
             return CollectConcreteGeometry(geomElement, useSymbolGeometry);
         }
 


Dewb:
Have not run tests yet, but wanted to get the discussion started.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -532,6 +532,19 @@ internal IEnumerable<Autodesk.Revit.DB.GeometryObject> InternalGeometry(bool use
...
             }
 
+            // FamilySymbol is a special case: the symbol must be activated before geometry is available, unless an instance is already present in the model
+            else if ((thisElement is FamilySymbol) && (geomElement == null || !geomElement.Any()))
+            {
+                var fs = (FamilySymbol)thisElement;
+                if (!fs.IsActive)
+                {
+                    fs.Activate();
+                    DocumentManager.Regenerate();
```
Dewb: The documentation at https://www.revitapidocs.com/2015/69706e35-87cc-a6d9-68fe-90a41c1c48db.htm does not mention that regeneration is required, but Scott C suggested it. Looking into whether this is strictly necessary or whether we need to update the Revit API docs.

Dewb: Scott C reminded me that that site is not official, the official docs do mention the regeneration requirement.

mjkkirschner: Will this kick off document updates? If so that might end up re-executing the graph as some nodes (and in some cases DynamoRevitModel) are subscribed to document updated events.

AndyDu1985:
Did you run RevitTestFramework to see if all tests can pass?

src/Libraries/RevitNodes/Elements/Element.cs
AndyDu1985: There is a `DocumentManager.Regenerate()` at the beginning of this method, could we activate this `FamilySymbol` also at the beginning?

AndyDu1985:
@ZiyunShang Could you look into it?

src/Libraries/RevitNodes/Elements/Element.cs
AndyDu1985: Yes, I think it will. But the second time it won't call `DocumentManager.Regenerate()` again for the same family symbol, as fs.IsActive is true.

AndyDu1985: Seems this will be one time action, I think it can just be here.

ZiyunShang:
I have run the regression test in Revit on farm, all tests passed.

PR #2452, opened 2019-11-21
Element nodes 1/3

SHKnudsen:
### Purpose

This is the first of 3 PR's for the Element category. The PR contains 5 nodes related to Revit.Elements.Element. 
image

#### Unit Test
image

#### System Test
image

### Declarations

### Reviewers

@AndyDu1985 
@radumg 

### FYIs

#### Element.IsPinned
image

#### Element.SetPinnedStatus
image

#### Element.HostedElements
image

#### Element.AreJoined
image

#### Element.GetJoinedElements
image


ZiyunShang:
It looks good to me.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -501,6 +527,56 @@ public object[] Geometry()
...
+        public List<Element> GetHostedElements(
+            bool includeOpenings = false,
+            bool includeShadows = false,
+            bool includeEmbeddedWalls = false,
+            bool includeSharedEmbeddedInserts = false)
+        {
+
+            HostObject hostObject = this.InternalElement as HostObject;
+            if (hostObject == null)
+                throw new NullReferenceException("Element is not a Host Element");
```
mjkkirschner: this should be in a .resx file.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -692,9 +768,20 @@ protected bool IsAlive
                 return !ElementIDLifecycleManager<int>.GetInstance().IsRevitDeleted(InternalElementId.IntegerValue);
             }
         }
+        /// <summary>
+        /// Finds the elements that are joined with the given element.
+        /// </summary>
+        /// <returns>All elements joined to the given element</returns>
+        public Element[] GetJoinedElements()
```
mjkkirschner: In general it is best to use interface types when defining apis - IEnumerable<Element> rather than array unless there is a reason to use array.

radumg: Yep, agree - didn't check for this as most of the nodes in that same category are returning arrays. @SHKnudsen fixed this now though.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -772,14 +790,81 @@ protected bool IsAlive
...
+        /// </summary>
+        /// <param name="elements">List of elements to unjoin</param>
+        /// <returns>Elements that have been unjoined</returns>
+        public static List<Element> UnjoinGeometry(List<Element> elements)
+        {
+            List<Element> modifiedElements = new List<Element>();
+            for (int i = 0; i < elements.Count; i++)
+            {
+                Element[] joinedElements = elements[i].GetJoinedElements();
+                if (joinedElements.Length > 0)
```
radumg: invert `if` and bail early (with `continue;` for example) if we can't proceed.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -772,14 +790,81 @@ protected bool IsAlive
...
+                        JoinGeometryUtils.UnjoinGeometry(
+                            Document,
+                            elements[i].InternalElement,
+                            joinedElements[j].InternalElement);
+
+                        // check if the unjoined element is already in the modifiedElements list 
+                        // if not we add it here
+                        if (!modifiedElements.Any(item => item.Id == joinedElements[j].Id))
+                        {
+                            modifiedElements.Add(joinedElements[j]);
```
radumg: should we return a list of lists ? that way users can use the output as it's grouped by elements.
if so, we would create an intermediate list for the inner `for` loop and add that to the master list
it would also reduce the number of times we're checking ids are already added to once.

test/Libraries/RevitIntegrationTests/ElementTests.cs
```diff
@@ -40,7 +40,36 @@ public void CanDeleteElement()
             // query count node to verify 1 item deleted as a result of the wall deletion. 
             Assert.AreEqual(1, GetPreviewValue("ccd8a5ba37fd4b1297def564392ccf54"));
          }
-   
+
+        [Test]
+        [TestModel(@".\Element\elementJoin.rvt")]
+        public void CanCheckIfTwoElementsAreJoined()
+        {
+            #region Arange
```
radumg: no regions inside tests

test/System/Element/canCheckIfElmentsAreJoined.dyn
```diff
@@ -0,0 +1,458 @@
+{
+  "Uuid": "3c9d0464-8643-5ffe-96e5-ab1769818209",
+  "IsCustomNode": false,
+  "Description": "",
+  "Name": "canCheckIfElmentsAreJoined",
```
radumg: file name (and the json prop as a consequence) has a typo : `Elments`

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -151,6 +151,24 @@ public string UniqueId
...
 
+        /// <summary>
+        /// Checks if two elements are joined
+        /// </summary>
+        /// <param name="otherElement">Element to check</param>
+        /// <returns>True if the two elements are joined, False otherwise</returns>
+        public bool IsJoined(Element otherElement)
+        {
+            if (this.InternalElement == null)
+                throw new Exception(nameof(this.InternalElement));
```
radumg: did we still need these `null` checks ? thought the base takes care of it.
if we do, need to make them `ArgumentNullException`s

radumg:
branch `DemoMeeting` got accidentally merged here.

PR #2460, opened 2019-12-12
Element nodes third batch

SHKnudsen:
### Purpose

This is the third of 3 PR's for the Element category. The PR contains 3 nodes related to Revit.Elements.Element.

AllNodes

All of the nodes have related unit and system tests, which have been run in RTF before submitting.

#### Unit Test
Element3of3_UnitTest_121219

#### System Test
Elements3of3_systemtest_121219

### Declarations

### Reviewers
@AndyDu1985
@radumg
@mjkkirschner 
@ZiyunShang 

### FYIs
#### Element.GetChildElements
Element GetChildElements

#### Element.GetParentElements
Element GetParentElement

#### Element.Transform
Element Transform


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,252 @@ protected bool IsAlive
...
+                t.CancelTransaction();
+                throw new Exception(ex.Message);
+            }         
+            TransactionManager.Instance.TransactionTaskDone();
+            return this;
+        }
+
+        private void RotateElementFromCS(Element element, CoordinateSystem fromCS, CoordinateSystem contextCS)
+        {
+            var familyInstance = element as FamilyInstance;
```
ZiyunShang: It seems that the transform operation can only be performed when the element can be as a FamilyInstance.
Is this node suitable for Element? I use other Element for transform operation and only get "Element.Transform operation failed.
element" information, which does not help people use it properly

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,252 @@ protected bool IsAlive
...
+        public IEnumerable<Element> GetChildElements()
+        {
+            return GetElementChildElements(this.InternalElement);
+            
+        }
+
+        private IEnumerable<Element> GetElementChildElements(Autodesk.Revit.DB.Element element)
+        {
+            List<Element> components = new List<Element>();
+            BuiltInCategory builtInCategory = (BuiltInCategory)System.Enum.Parse(typeof(BuiltInCategory),
```
mjkkirschner: what if this fails - is it possible the category is not a built in category?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,252 @@ protected bool IsAlive
...
+                    break;
+            }
+            return components;
+        }
+
+        private static List<Element> GetChildElementsFromFamilyInstance(Autodesk.Revit.DB.Element element)
+        {
+            var familyInstance = element as Autodesk.Revit.DB.FamilyInstance;
+            if (familyInstance == null)
+                throw new NullReferenceException(Properties.Resources.NoChildElements);
```
mjkkirschner: throwing and catching exceptions is quite slow - are we sure we don't just want to return an empty list?

src/Libraries/RevitNodes/Elements/Element.cs
mjkkirschner: yeah agree, this either needs better docs/error message, a more constrained type input, or both.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,252 @@ protected bool IsAlive
...
+            TransformUtils.ExtractEularAnglesFromTransform(newTransform, out newRotationAngles);
+            // Convert Eular angle to degrees
+            var rotationDegrees = (newRotationAngles.FirstOrDefault() / (2 * Math.PI)) * 360;
+            var newRotationAngle = rotationDegrees * Math.PI / 180;
+
+            if (!oldRotationAngles[0].AlmostEquals(newRotationAngle, 1.0e-6))
+            {
+                var rotateAngle = newRotationAngle - oldRotationAngles[0];
+                var axis = Autodesk.Revit.DB.Line.CreateUnbound(oldTransform.Origin, oldTransform.BasisZ);
+                Autodesk.Revit.DB.ElementTransformUtils.RotateElement(Document, new Autodesk.Revit.DB.ElementId(Id), axis, -rotateAngle);
```
mjkkirschner: hmm, will this always be negative?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,252 @@ protected bool IsAlive
...
+        /// Gets the child Elements of the current Element.
+        /// </summary>
+        /// <returns>Child Elements.</returns>
+        public IEnumerable<Element> GetChildElements()
+        {
+            return GetElementChildElements(this.InternalElement);
+            
+        }
+
+        private IEnumerable<Element> GetElementChildElements(Autodesk.Revit.DB.Element element)
```
QilongTang: To be consistent, can we call it `GetChildElementsFromElement`?

src/Libraries/RevitNodes/Elements/Element.cs
QilongTang: @mjkkirschner I guess we are trying to distinguish the case that an element comes with zero child elements V.S. an element is simply not supported for such node operation?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,252 @@ protected bool IsAlive
...
+            return beamSystemComponentElements;
+        }
+
+        private static List<Element> GetChildElementsFromStairs(Autodesk.Revit.DB.Element element)
+        {
+            var stairElement = element as Autodesk.Revit.DB.Architecture.Stairs;
+            List<ElementId> stairComponentIds = new List<ElementId>();
+            stairComponentIds.AddRange(stairElement.GetStairsLandings().ToList());
+            stairComponentIds.AddRange(stairElement.GetStairsRuns().ToList());
+            stairComponentIds.AddRange(stairElement.GetStairsSupports().ToList());
```
QilongTang: Can we add more comments here indicating why these are counted as child elements or point to an online reference? I feel like there is some hidden knowledge here for a Revit newbie like me

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,252 @@ protected bool IsAlive
...
+                                                           .ToList();
+            return componentElements;
+        }
+
+        private static List<Element> GetChildElementsFromRailings(Autodesk.Revit.DB.Element element)
+        {
+            var railingElement = element as Autodesk.Revit.DB.Architecture.Railing;
+            List<ElementId> railingComponentIds = new List<ElementId>();
+            railingComponentIds.AddRange(railingElement.GetHandRails().ToList());
+            railingComponentIds.Add(railingElement.TopRail);
```
QilongTang: Same here, comments for these two lines please incase these need to be updated in the future

src/Libraries/RevitNodes/Elements/Element.cs
SHKnudsen: I haven't been able to find a way to get the Transform of an Element, which is why I am casting it to FamilyInstance as this has a GetTransform(). Happy to hear if you guys know of a way I can get the Transform of an Element.

Maybe this should be a node under FamilyInstance instead to avoid the confusion?

src/Libraries/RevitNodes/Elements/Element.cs
radumg: or better yet `GetElementChildren` ?

src/Libraries/RevitNodes/Elements/Element.cs
radumg: agree, @SHKnudsen let's add some notes with the knowledge/info we gathered with Revit team

src/Libraries/RevitNodes/Elements/Element.cs
SHKnudsen: as far as I know it cant? @AndyDu1985 @ZiyunShang any thoughts on this?

src/Libraries/RevitNodes/Elements/Element.cs
SHKnudsen: our thought on throwing is to provide more information to the user, might be confusing to just get an empty list as it might not be if something was done wrong

src/Libraries/RevitNodes/Elements/Element.cs
SHKnudsen: using the code from the original SeRotation method which always has the `rotateAngle` as negative.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,205 @@ protected bool IsAlive
...
+            {
+                case BuiltInCategory.OST_StairsLandings:
+                    parent = GetParentComponentFromStairsLandings(element);
+                    break;
+
+                case BuiltInCategory.OST_StructuralFraming:
+                    parent = GetParentElementFromStructuralFraming(element);
+                    break;
+
+                case BuiltInCategory.OST_Railings:
```
ZiyunShang: From GetChildElements, OST_Railings can get HandRails and TopRail as its child elements. But here shows, it can't get the parent element if the element is hand rails or top rails. This makes these two methods irreversible. I think it's better to add these two cases: 
case BuiltInCategory.OST_RailingHandRail:
case BuiltInCategory.OST_RailingTopRail:
What do you think of this?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -769,6 +771,205 @@ protected bool IsAlive
...
+        {
+            Autodesk.Revit.DB.Element parent;
+            BuiltInCategory builtInCategory = (BuiltInCategory)System.Enum.Parse(typeof(BuiltInCategory),
+                                                                                 element.Category.Id.ToString());
+
+            // By default we use the SuperComponent on the elements FamilyInstance to get the Parent Element, 
+            // for now the node also handles special cases including Stairs, StructuralFramingSystems and Railings 
+            switch (builtInCategory)
+            {
+                case BuiltInCategory.OST_StairsLandings:
```
ZiyunShang: In GetChildElements, OST_Stairs will get its landings, runs and supports. landings and runs all have GetStairs method. But for supports, I can't find any Revit API to get their stairs, @AndyDu1985 , do you have any suggestions for this?

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
```diff
@@ -543,6 +544,41 @@ public FamilyInstance SetRotation(double degree)
             return this;
         }
 
+        ///// <summary>
+        ///// Transforms this FamilyInstance from a source CoordinateSystem to a new context CoordinateSystem.
+        ///// </summary>
+        ///// <param name="fromCoordinateSystem">Source CoordinatSystem</param>
+        ///// <param name="contextCoordinateSystem">Context CordinateSystem</param>
+        ///// <returns>Transformed Element</returns>
+        public FamilyInstance Transform(CoordinateSystem fromCoordinateSystem, CoordinateSystem contextCoordinateSystem)
```
AndyDu1985: Seems this implementation can be applied to all kinds of model elements, right? If yes, I think it can be moved into Element Category.

SHKnudsen: As mentioned above, the reason i am using FamilyInstance is because i need to check the elements Transform, i cannont seem to find a way to do this if the element is not a FamilyInstance? Maybe you have some ideas?

AndyDu1985: I use below code to do move and rotate, works for Wall, Stairs, FamilyInstance. From the document of ElementTransformUtils Class, "A collection of utilities allowing transformation of elements (e.g. move, rotate, mirror and copy). ", I think it works for all kinds of element.
`
public void MoveElement()
{
	Reference refer = Application.ActiveUIDocument.Selection.PickObject(ObjectType.Element);
	Transaction tran = new Transaction(Document, "Move Element");
	tran.Start();
	ElementTransformUtils.MoveElement(Document, refer.ElementId, new XYZ(10, 10, 10));
	tran.Commit();
}

public void rotateElement()
{
	Reference refer = Application.ActiveUIDocument.Selection.PickObject(ObjectType.Element);
	Line zAxis = Line.CreateBound(new XYZ(0, 0, 0), new XYZ(0, 0, 1));
	Transaction tran = new Transaction(Document, "Move Element");
	tran.Start();
	ElementTransformUtils.RotateElement(Document, refer.ElementId, zAxis, 90);
	tran.Commit();
}
`

SHKnudsen: What i need is to check the old `Transform` of the element before applying any new `Transform` to it. For example the `SetRotation` node in `FamilyInstance` does this

    var oldTransform = InternalGetTransform();
    double[] oldRotationAngles;
    TransformUtils.ExtractEularAnglesFromTransform(oldTransform, out oldRotationAngles);
    var newRotationAngle = degree * Math.PI / 180;

    if (!oldRotationAngles[0].AlmostEquals(newRotationAngle, 1.0e-6))

it gets the old transformation by doing `InternalGetTransformation()` which uses the `GetTransform`. Im guessing that the `SetRotation` node only works with `FamilyInstance` because of this? Which is also why i had to move this to the `FamilyInstance`

AndyDu1985: OK, I see. And I also find Element.SetLocation, which use with FamilyInstance.SetLocation, can get the same effect as FamilyInstance.Transform.

src/Libraries/RevitNodes/Elements/Element.cs
AndyDu1985: For element.Category, I think they should be builtInCategory. And user may add sub category to a category which may not be a builtInCategory, I'm not sure about it. 
I suggest to add a check to see if `builtInCategory` is `INVALID` .

src/Libraries/RevitNodes/Elements/FamilyInstance.cs
AndyDu1985: I suggest to remove the new node `FamilyInstance.Transform`.

src/Libraries/RevitNodes/Elements/Element.cs
AndyDu1985: Any updates?

PR #2468, opened 2020-01-06
Element type nodes

SHKnudsen:
### Purpose

This PR introduces an ElementType category to the Dynamo library, all other Dynamo types (WallType, FloorType, etc.) now inherits from ElementType which means a lot of duplicated code has been removed.

The following new nodes have been added:
#### ElementType nodes
ElementType

#### WallType nodes
WallType

#### FloorType nodes
FloorType

#### RoofType nodes
RoofType

#### Unit Test
ElementType_UnitTest_030120

#### System Test
ElementType_SysytemTest_030120

### Declarations

### Reviewers
@AndyDu1985
@radumg
@mjkkirschner 
@ZiyunShang 

### FYIs
#### ElementType Properties test
CanGetElementTypeProperties

#### ElementType by name test
CanGetElementTypeByName

#### Duplicate ElementType test
CanDuplicateElementType

#### ElementType get preview image test
GetElementTypePreviewImage

#### WallType Properties test
CanGetWallTypeProperties

#### WallType Thermal Properties test
CanGetWallTypeThermalProperties

#### Check if FloorType is foundation slab test
CanCheckIfFloorTypeIsFoundationSlab

#### get FloorType structural material test
CanGetFloorTypeStructuralMaterial

#### FloorType thermal properties test
CanGetFloorTypeThermalProperties

#### RoofType thermal properties test
CanGetRoofTypeThermalProperties

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Element.cs
+++ b/src/Libraries/RevitNodes/Elements/Element.cs
@@ -197,23 +197,19 @@ namespace Revit.Elements
         }
 
         /// <summary>
-        /// Returns the FamilyType for this Element. Returns null if the Element cannot have a FamilyType assigned.
+        /// Returns the ElementType for this Element. Returns null if the Element cannot have an ElementType assigned.
         /// </summary>
         /// <returns name="ElementType">Element Type or Null.</returns>
-        public Element ElementType
+        public ElementType ElementType
         {
             get
             {
                 var typeId = this.InternalElement.GetTypeId();
                 if (typeId == ElementId.InvalidElementId)
-                {
                     return null;
-                }
-                else
-                {
-                    var doc = DocumentManager.Instance.CurrentDBDocument;
-                    return doc.GetElement(typeId).ToDSType(true);
-                }
+                
+                var doc = DocumentManager.Instance.CurrentDBDocument;
+                return doc.GetElement(typeId).ToDSType(true) as ElementType;
             }
         }
 
--- a/src/Libraries/RevitNodes/Elements/RoofType.cs
+++ b/src/Libraries/RevitNodes/Elements/RoofType.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
+using Autodesk.DesignScript.Runtime;
+using Autodesk.Revit.DB;
 using RevitServices.Persistence;
 
 namespace Revit.Elements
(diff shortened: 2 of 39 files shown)


mjkkirschner:
This PR has API breaking implications. You must be very careful when refactoring like this - Code which previously was written on top of the DynamoRevit node classes now has the potential to break.

It's difficult to say what will happen unless it is throughly tested. But I know for certain that this breaks a few public APIs and as such should not be released as a Dynamo 2.x change.

In general you should not change the type or signature of public members until you are ready to break an API.

https://stackoverflow.com/questions/1456785/a-definitive-guide-to-api-breaking-changes-in-net

for example, this similar PR broke lots of code:
https://github.com/DynamoDS/DynamoRevit/pull/2282

If you intend to get this into 2.x then I would act as conservatively as possible, do not refactor these classes, but add additional methods to `Element` using a new interface, and obsolete old properties.

If you intend to get this into 3.x then leave this PR open and maintain it (merging master) until DynamoRevit intentionally decides to break API compatibility- that could be today! - but thats for @ZiyunShang @AndyDu1985 and Minjie to decide think.

radumg:
Thanks for the notes @mjkkirschner , it's an interesting issue! I hadn't originally flagged this as an issue in our internal review as the signature of the classes will _essentially_ resolve fine at run-time, but now realise I hadn't considered whether Dynamo uses fully qualified names when compiling nodes/etc.

Would be great if you could give us a few examples of the APIs that you know will break, would love to test those a bit more. If it's still an issue, we can take a more conservative approach, no probs.

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -197,23 +197,19 @@ public Category GetCategory
         }
 
         /// <summary>
-        /// Returns the FamilyType for this Element. Returns null if the Element cannot have a FamilyType assigned.
+        /// Returns the ElementType for this Element. Returns null if the Element cannot have an ElementType assigned.
         /// </summary>
         /// <returns name="ElementType">Element Type or Null.</returns>
-        public Element ElementType
+        public ElementType ElementType
```
mjkkirschner: @radumg here is an example of a simple API break - this class no longer has as property called `Element`.

src/Libraries/RevitNodes/Elements/RoofType.cs
```diff
@@ -1,31 +1,29 @@
...
+using Autodesk.Revit.DB;
 using RevitServices.Persistence;
 
 namespace Revit.Elements
 {
     /// <summary>
     /// A Revit RoofType
     /// </summary>
-    public class RoofType : Element
+    public class RoofType : ElementType
```
mjkkirschner: for any of these types which now have new base classes, it might be fine, or it might not, it's worth considering a package which used this type as the input to a node or even just internally, will this code need to be recompiled? Will it need to have changes made and then recompiled? 

Can you prove it won't?

mjkkirschner: also, it looks like you did not have to modify any existing tests? That is usually a good sign for these type of refactors.

I think the best that can be done, if you want to move forward with a refactor like this is to try this change out on a package that was compiled against an old version of DynamoRevit and see what happens. This can be run as an automated test.

also, please see the one change which looks to be a genuine break.

AndyDu1985:
@SHKnudsen This PR cause build failure, I've revert it, could you resend a new PR? Thanks!

PR #2490, opened 2020-02-03
Family document nodes

SHKnudsen:
### Purpose

This PR adds a new FamilyDocument class with 10 nodes + a FamilyParameter wrapper class

AllNodes

#### Unit Test
FamilyDocumentNodes_UnitTest_230120

#### System Test
FamilyDocumentNodes_SystemTest_230120

### Declarations

### Reviewers
@AndyDu1985
@radumg
@mjkkirschner 
@ZiyunShang 

### FYIs
#### CanAddParameterToFamilyDocument
CanAddParameterToFamilyDocument

#### CanCreateFamilyDocumentByDocument
CanCreateFamilyDocumentByDocument

#### CanDeleteParameterFromFamilyDocument
CanDeleteParameterFromFamilyDocument

#### CanGetFamilyDocumentCategory
CanGetFamilyDocumentCategory

#### CanGetFamilyDocumentFormula
CanGetFamilyDocumentFormula

#### CanGetFamilyDocumentParameters
CanGetFamilyDocumentParameters

#### CanGetFamilyDocumentValue
CanGetFamilyDocumentValue

#### CanSetFamilyDocumentCategory
CanSetFamilyDocumentCategory

#### CanSetFamilyDocumentFormula
CanSetFamilyDocumentFormula

#### CanSetFamilyDocumentValue
CanSetFamilyDocumentValue

#### CanGetFamilyDocumentTypes
CanGetFamilyDocumentTypes


erfajo:
Looks like yet another copy of nodes I have had for years... However, it is a kind of acknowledgement that it seems to be so. When do you plan to have the entire Orchid package replicated?

andydandy74:
@SHKnudsen - Nice to see so many PRs with new Revit nodes. I see quite a bit of overlap with Clockwork and I'm looking forward to retire some nodes once these PRs get released into the wild ;-)
With this particular PR I do have some questions, though:
Why should I need to convert documents (exist in the Revit API) to family documents (don't exist in the Revit API) here and why would I even have to tell the node explicitly which is the current document? Barely none of the existing Revit nodes have a document input. Why can't these nodes act the same way?
Having an **optional** document input for background open operations would be nice, though. But I have a feeling that's not what it's for.
Also, I am curious about the SetValue node. Shouldn't that have another input for the family type? Right now, does it set the value for all types or the current type?

SHKnudsen:
@andydandy74 - Thank your for the constructive comments!

First of all, good point about being able to choose which family type to Get/Set values or formulas on, right now it simply uses the current type. Will have look at adding this to the nodes.

There is few reasons for introducing the `FamilyDocument` class, as you say the Revit API doesn't distinguish between a `Document` and a `FamilyDocument`, it's all just `Documents`. We added the `FamilyDocument` so the nodes explicitly tells what it needs, if it just requires a `Document` it might not be obvious what to input. Might be worth getting @AndyDu1985 and @ZiyunShang opinion on this.

Regarding opening "FamilyDocuments" in the background, this is in fact possible with this approach. We have another node that i will make a PR on soon as well, that can get the `FamilyDocument` from a `Family`. Below picture sets the formula on a parameter in the `FamilyDocument` without opening the family. (might be good to get @AndyDu1985 and @ZiyunShang opinion on this as well)
image

ZiyunShang:
Hi @SHKnudsen , could you resolve conflicts here?

amirmoeini65:
@SHKnudsen 
Hi, Thank you for sharing this amazing post. I have a question. I was trying to use these nodes to set a new value for a parameter in a nested family. The code runs without any warning and the new value is set to that parameter, but when I check the Revit model, there's no update in the value of that parameter. Should I use another node at the end?

image

PR #2507, opened 2020-03-02
Update system tests

ZiyunShang:
### Purpose
Update some System Tests - some tests may failure due to element ID is unstable.

### Declarations

### Reviewers

@AndyDu1985 

### FYIs

@SHKnudsen

### DIFF
--- a/test/Libraries/RevitIntegrationTests/ElementTests.cs
+++ b/test/Libraries/RevitIntegrationTests/ElementTests.cs
@@ -236,7 +236,7 @@ namespace RevitSystemTests
 
         }
 
-        [Test]
+        [Test, Ignore, Category("Failure")]
         [TestModel(@".\Element\elementTransform.rvt")]
         public void CanTransformElement()
         {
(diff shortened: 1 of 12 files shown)


test/Libraries/RevitIntegrationTests/ElementTests.cs
```diff
@@ -236,7 +236,7 @@ public void CanUnjoinListOfElements()
 
         }
 
-        [Test]
+        [Test, Ignore, Category("Failure")]
```
ZiyunShang: This test is failed because the original location is changed after transform in the Revit document. Original location is the same as transform location.

test/Libraries/RevitIntegrationTests/ElevationMarkerTests.cs
```diff
@@ -18,7 +18,7 @@ namespace RevitSystemTests
     class ElevationMarkerTests : RevitSystemTestBase
     {
         
-        [Test]
+        [Test, Ignore, Category("Failure")]
```
ZiyunShang: This test failed, in the CanCreateElevationMarker.dyn, a Node which is contained in third party package 'Rhythm', was used to get ViewFamilyType.

PR #2532, opened 2020-04-01
Fix SystemTests failure

ZiyunShang:
### Purpose

Update some system tests which category is failure.
Some tests was commented out because of obsolete nodes.
Some tests was updated on codes and dyn scripts.

### Declarations

### Reviewers

@AndyDu1985 

### FYIs

@QilongTang @mjkkirschner


test/Libraries/RevitIntegrationTests/MigrationTest.cs
```diff
@@ -68,7 +68,7 @@ public void TestMigration_Analyze_Measure()
         [Test, Category("Failure")]
         [TestModel(@".\empty.rfa")]
         public void TestMigration_Analyze_Render()
-        {
+        {   //There are some obsolete node in dyn script.
```
ZiyunShang, line 71: The three tests of Migration tests have some obsolete node, but I don't know if this is what this test is expecting.

PR #2537, opened 2020-04-05
Longest of shortest exit paths node

ahmedidoumohmed:
### Purpose

LongestOfShortestExitPaths node: Regression test model RoomsAndExits.rvt is now issued from Revit 2021. Touch tests accordingly as the model was regenerated from scratch using Revit 2021.

### Declarations

- [x ] The code base is in a better state after this PR
- [x ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [x ] The level of testing this PR includes is appropriate

### Reviewers

@lukechurch
@awbushnell
@bykovsm
@t-mayrf
@aan.du@autodesk.com
@ziyun.shang@autodesk

### FYIs

@awbushnell
@shafirb
@wangmin


src/Libraries/RevitNodes/Elements/Views/FloorPlanView.cs
```diff
@@ -60,6 +52,18 @@ private void InitFloorPlanView(Autodesk.Revit.DB.Level level)
 
         #endregion
 
+        #region Public constructors
+
+        /// <summary>
+        /// Create a Revit Floor Plan from Autodesk View Plan
+        /// </summary>
+        public  FloorPlanView(Autodesk.Revit.DB.ViewPlan view)
```
mjkkirschner, line 60: this likely should not be public - or it should at least be hidden from the node library.

ahmedidoumohmed: When I did the merge, I didn't pay attention to ZiyunShang commit ( revert public constructor to private (#2523)) where the fix for using default view in tests without using the private constructor of FloorPlanView was implemented.
Changes reverted back to ZiyunShang  commit.
Thanks and sorry.

PR #2560, opened 2020-05-12
Improve d4r drop down nodes

ZiyunShang:
### Purpose

For DynamoRevit Dropdown Nodes, a mechanism to monitor Revit Element change is needed so that the nodes can follow the changes of Revit document.

### Declarations

### Reviewers

@AndyDu1985 

### FYIs

@mjkkirschner @QilongTang

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitDropDown.cs
+++ b/src/Libraries/RevitNodesUI/RevitDropDown.cs
@@ -25,6 +25,9 @@ using Parameter = Autodesk.Revit.DB.Parameter;
 using Revit.Application;
 using System.Text.RegularExpressions;
 
+using RevitServices.Elements;
+using RevitServices.Transactions;
+
 namespace DSRevitNodesUI
 {
     public class DropDownItemEqualityComparer : IEqualityComparer<DynamoDropDownItem>
@@ -42,15 +45,16 @@ namespace DSRevitNodesUI
 
     public abstract class RevitDropDownBase : DSDropDownBase
     {
-
         protected RevitDropDownBase(string value) : base(value)
         {
+            RevitServicesUpdater.Instance.ElementsUpdated += Updater_ElementsUpdated;
             DynamoRevitApp.EventHandlerProxy.DocumentOpened += Controller_RevitDocumentChanged;
         }
 
         [JsonConstructor]
         public RevitDropDownBase(string value, IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(value, inPorts, outPorts)
         {
+            RevitServicesUpdater.Instance.ElementsUpdated += Updater_ElementsUpdated;
             DynamoRevitApp.EventHandlerProxy.DocumentOpened += Controller_RevitDocumentChanged;
         }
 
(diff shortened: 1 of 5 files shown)


mjkkirschner:
It’s important to consider if this makes it much easier for users to inadvertently create an infinite loop - and if so how to avoid that.

ZiyunShang:
> It’s important to consider if this makes it much easier for users to inadvertently create an infinite loop - and if so how to avoid that.
> …

@mjkkirschner I also think that your concern is necessary. I used the RevitServicesUpdater.Instance.ElementsUpdated event, referring to the use of several other places like ElementsInView, SiteLocation and Selection. 
I can’t think of any scenarios that can cause loops. Do you have any specific examples?

mjkkirschner:
@ZiyunShang - maybe testing something like a dropdown that accesses all family types in the document, and then add a node downstream which modifies the document, does this cause any issue, I guess another case to be aware of is all the user code which may inherit from DSDropDownBase - it might be hard to guarantee this does not cause issues, maybe it's best to run the full test suite and get it into the hands of users for some extra testing via some targeted beta feedback?

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -59,8 +63,29 @@ void Controller_RevitDocumentChanged(object sender, EventArgs e)
...
         }
 
+        private void Updater_ElementsUpdated(object sender, ElementUpdateEventArgs e)
+        {
+            switch (e.Operation)
+            {
+                case ElementUpdateEventArgs.UpdateType.Added:
+                    break;
+                case ElementUpdateEventArgs.UpdateType.Modified:
+                    bool dynamoTransaction = e.Transactions.Contains(TransactionWrapper.TransactionName);
```
mjkkirschner, line 73: ahh, I see you're using this to stop the loops caused by nodes, so that should be good - in other examples did you also see this check being done for added elements?

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -59,8 +63,29 @@ void Controller_RevitDocumentChanged(object sender, EventArgs e)
             PopulateItems();
         }
 
+        private void Updater_ElementsUpdated(object sender, ElementUpdateEventArgs e)
+        {
+            switch (e.Operation)
+            {
+                case ElementUpdateEventArgs.UpdateType.Added:
```
AndyDu1985, line 70: Expect case  ElementUpdateEventArgs.UpdateType.Modified, other cases can all be the default? Should them be list here?

src/Libraries/RevitNodesUI/RevitDropDown.cs
ZiyunShang: I just found only "Selection" class has this check, https://github.com/DynamoDS/DynamoRevit/blob/433e00f19dd6cf3e3554a4317ca34dfe77e9920e/src/Libraries/RevitNodesUI/Selection.cs#L224 
And in other case, don't have checked transactions.
I investigated on Dynamo Forum, and found that some issues about infinite loop, most of them related to ElementsQueryBase nodes like All Elements of Category and All Elements of Type.

src/Libraries/RevitNodesUI/RevitDropDown.cs
ZiyunShang: Yes, other case may be useless here, I will update soon.

ZiyunShang:
@mjkkirschner @AndyDu1985 I have a new commit for DropDown Nodes Deleted event and Transaction check for some nodes like ElementsQueryBase nodes. Could you help review it?

AndyDu1985:
@ZiyunShang Looks good to me, please make sure no test is broken.

ZiyunShang:
@AndyDu1985 Yes, all the Regression tests have passed with this change.

PR #2565, opened 2020-05-21
AGD-1513 Add SelectModelElementByCategory node to D4R to compliment the new SelectElementsByCategory node

saintentropy:
### Purpose

https://jira.autodesk.com/browse/AGD-1526

SelectModelElementByCategoryNode exists in DynamoRevit that works similar to the plural version but follows the singular selection UX.

Screen Shot 2020-05-21 at 3 28 13 PM
### Declarations

### Reviewers

@SHKnudsen @AndyDu1985 @ZiyunShang 

### FYIs

@radumg @mjkkirschner

### DIFF
--- a/src/Libraries/RevitNodesUI/SelectionCombo.cs
+++ b/src/Libraries/RevitNodesUI/SelectionCombo.cs
@@ -89,6 +89,66 @@ namespace Dynamo.ComboNodes
         }
     }
 
+    [NodeName("Select Model Element By Category"),
+    NodeCategory(Revit.Elements.BuiltinNodeCategories.REVIT_SELECTION),
+    NodeDescription("SelectModelElementByCategoryDescription", typeof(DSRevitNodesUI.Properties.Resources)),
+    IsDesignScriptCompatible]
+    public class DSModelElementByCategorySelection : ElementFilterSelection<Element>
+    {
+        private const string message = "Select Model Element";
+        private const string prefix = "Element";
+
+        internal CategoryElementSelectionFilter<Element> SelectionFilter { get; set; }
+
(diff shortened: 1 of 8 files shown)


src/Libraries/RevitNodesUI/SelectionCombo.cs
```diff
@@ -33,7 +33,7 @@ namespace Dynamo.ComboNodes
     NodeCategory(Revit.Elements.BuiltinNodeCategories.REVIT_SELECTION),
     NodeDescription("SelectModelElementsByCategoryDescription", typeof(DSRevitNodesUI.Properties.Resources)),
     IsDesignScriptCompatible]
-    public class DSModelElementByCategorySelection : ElementFilterSelection<Element>
+    public class DSModelElementsByCategorySelection : ElementFilterSelection<Element>
```
mjkkirschner, line 36: This is an API break.

saintentropy: Shouldn't break if this hasn't been released yet which I will confirm.  We don't have to update this...  It just seemed appropriate to update the class name to plural as this PR adds the actual singular Element selection version of the node.

ZiyunShang:
All tests passed, it looks good to me.

ThomasMahon:
Having the category as an input embedded in the node is bad idea IMO for the simple reason that a user might want to provide multiple categories as an input but cant with what is being proposed. If the category is an input then the user can provide as many categories as they need.

johnpierson:
>  If the category is an input then the user can provide as many categories as they need.

☝️ I agree with this approach completely. It would be great because now you can use the other built-in ways of providing categories to this node.

PR #2600, opened 2020-07-21
Material structural parameters

ZiyunShang:
### Purpose

When use Material.StructuralParameters node to get StructuralParameters from a Material, there are some invalid parameters which just have ID, no name, no type, and can't get access to theirs StorageType and value. So I think it is reasonable to filter out these unavailable Parameters.

https://github.com/DynamoDS/DynamoRevit/issues/2598

### Declarations

### Reviewers

@AndyDu1985 @wangyangshi @QilongTang @mjkkirschner 

### FYIs

@Amoursol

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Material.cs
+++ b/src/Libraries/RevitNodes/Elements/Material.cs
@@ -204,13 +204,7 @@ namespace Revit.Elements
                     Autodesk.Revit.DB.AppearanceAssetElement appearance = document.GetElement(this.InternalMaterial.AppearanceAssetId) as Autodesk.Revit.DB.AppearanceAssetElement;
                     if (appearance != null)
                     {
-                        foreach (var parameter in appearance.Parameters)
-                        {
-                            Parameter p = new Parameter(parameter as Autodesk.Revit.DB.Parameter);
-                            if (!appearances.Contains(p) && p.InternalParameter.Definition != null) appearances.Add(p);
-                            if (!invalidAppearances.Contains(p) && p.InternalParameter.Definition == null)
-                                invalidAppearances.Add(p);
-                        }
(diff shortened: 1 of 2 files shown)


src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -205,11 +207,22 @@ public IEnumerable<Parameter> AppearanceParameters
                         foreach (var parameter in appearance.Parameters)
                         {
                             Parameter p = new Parameter(parameter as Autodesk.Revit.DB.Parameter);
-                            if (!appearances.Contains(p)) appearances.Add(p);
+                            if (!appearances.Contains(p) && p.InternalParameter.Definition != null) appearances.Add(p);
```
AndyDu1985, line 210: Below maybe better:
```
if (p.InternalParameter.Definition == null)
{ 
   if(!invalidAppearances.Contains(p))
      invalidAppearances.Add(p); 
   continue;
}
if (!appearances.Contains(p)) 
   appearances.Add(p);
```

src/Libraries/RevitNodes/Elements/Material.cs
```diff
@@ -232,11 +246,24 @@ public IEnumerable<Parameter> ThermalParameters
                         foreach (var parameter in thermal.Parameters)
                         {
                             Parameter p = new Parameter(parameter as Autodesk.Revit.DB.Parameter);
-                            if (!thermals.Contains(p)) thermals.Add(p);
+                            if (!thermals.Contains(p) && p.InternalParameter.Definition != null) thermals.Add(p);
```
AndyDu1985, line 249: You can have a function to take `Parameters` as input, so you can reuse the code.

PR #2638, opened 2020-09-28
Fix a issue about Mass Family

ZiyunShang:
### Purpose

Fix a issue that it will have an offset when select a face or faces from Mass FamilyInstance into Dynamo.
#### Steps to reproduce the problem.
1- Create a Conceptual Mass family
2- Create a simple form
3- Draw some lines around
4- Load the family into the project
5- Open Dynamo and try the Select Face or Select Faces node and select a face of the loaded mass into project
6- Try using ImportInstance.ByGeometry and see how faces are far far away from where they are supposed to be

### Declarations

### Reviewers

@AndyDu1985 @wangyangshi @QilongTang @mjkkirschner 

### FYIs

@Amoursol

### DIFF
--- a/src/Libraries/RevitNodes/GeometryObjects/GeometryObjectSelector.cs
+++ b/src/Libraries/RevitNodes/GeometryObjects/GeometryObjectSelector.cs
@@ -67,6 +67,8 @@ namespace Revit.GeometryObjects
         /// <returns></returns>
         private static bool RequiresTransform(Autodesk.Revit.DB.FamilyInstance familyInstance)
         {
+            if (familyInstance.Category.Id.IntegerValue == (int)BuiltInCategory.OST_Mass)
+                return false;
             var geom = familyInstance.get_Geometry(new Options());
             return geom.OfType<Autodesk.Revit.DB.GeometryInstance>()
                 .Any(x => x.GetInstanceGeometry().Any());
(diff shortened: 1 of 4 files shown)


ZiyunShang:
Hi All, this issue was strange with the 3rd step - "Draw some lines around". Without these lines, RequiresTransform will return false, and Mass FamilyInstance won't be transformed, and its location will be correct. If there are, RequiresTransform will return true, and the faces' location will be incorrect.
So I think maybe Mass Family has no require to be transformed. BTW, I just try some other families, RequiresTransform will return true for all of them when Select faces or edges from them.
@mjkkirschner @QilongTang , I want to know why this condition is used to determine whether FamilyInstance needs transform?

QilongTang:
@ZiyunShang I'm afraid I am not aware why this is implemented this way. From the code history we can tell these code are written 6 years ago so maybe the assumption then is no longer valid. @mjkkirschner is still on sabbatical so if you found your fix reflect the expected behavior, let's ship it and confirm if he knows later maybe.

src/Libraries/RevitNodes/GeometryObjects/GeometryObjectSelector.cs
```diff
@@ -67,6 +67,8 @@ public static object ByReferenceStableRepresentation(string referenceString)
         /// <returns></returns>
         private static bool RequiresTransform(Autodesk.Revit.DB.FamilyInstance familyInstance)
         {
+            if (familyInstance.Category.Id.IntegerValue == (int)BuiltInCategory.OST_Mass)
+                return false;
```
AndyDu1985, line 71: I think it's safe to limit it to Mass family.

AndyDu1985 (approved):

Issue #2587, opened 2020-07-08
Element.ElementType node and its output port are misnamed


cjmccrone:
## Dynamo version

Dynamo Core 2.5.0.7460
Dynamo Revit 2.5.0.7586

## Revit version

Revit 2021

## Operating system

Windows 10

## What did you do? 

Selected an element in the Furniture category. I used the node Element.ElementType in order to determine an element's FAMILY TYPE. The node works correctly, but the node is called "Element Type" and its output port says "Element Type." That's a problem because the node actually returns a Family Type, so downstream nodes that take an "Element Type" as an input won't work.

## What did you expect to see? / What did you see instead?

![element_type](https://user-images.githubusercontent.com/5391789/86859331-3f11e500-c077-11ea-99a5-63cc5576d04f.png)

ZiyunShang:
Hi @cjmccrone , "ElementType" is a wrapper of "Autodesk.Revit.DB.ElementType", it is an element of DynamoRevit. ElementType is not only FamilyType, some other types also inherit from it, such as FloorType, RoofType, WallType, etc.
“All Element of Type” need a "System.Type" as input, you can get the "Types" with "Element Types" node under Revit->Selection.

@AndyDu1985 , Do you have any suggestions for this kind of naming?

johnpierson:
Closing this out - the question was answered in the thread above and the issue was simply never closed.

If the answer no longer holds on a current build, please open a fresh issue.

PR #2641, opened 2020-10-20
Rename "Element Types" to "Element Classes" distinguished from "Eleme…

ZiyunShang:
…ntType"

### Purpose

Rename Element Types node to make it more meaningful and distinguished from "ElementType"
Related to issue https://github.com/DynamoDS/DynamoRevit/issues/2587 

### Declarations

### Reviewers

@ShengxiZhang @wangyangshi 

### FYIs

@Amoursol @QilongTang @mjkkirschner


wangyangshi (approved):

src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -947,7 +947,7 @@ public class SpacingRuleLayouts : EnumAsInt<SpacingRuleLayout>
         public SpacingRuleLayouts(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(inPorts, outPorts) { }
     }
 
-    [NodeName("Element Types")]
+    [NodeName("Element Classes")]
```
ZiyunShang, line 950: This class is subclass of AllChildrenOfType<Element>, and can't modify its output name.
The output name is set in DynamoCore - https://github.com/DynamoDS/Dynamo/blob/master/src/Libraries/CoreNodeModels/Enum.cs#L65 .
So I think I need submit a update here.

ZiyunShang: ElementClass
Hi @ShengxiZhang , with this PR, the results as the screenshot shown. We can't rename the output name "Type" on DynamoRevit side.

mjkkirschner:
I think it's important that before making this change testing is done on graphs and custom nodes that reference these nodes to make sure they open okay.

ZiyunShang:
@mjkkirschner I just run all DynamoRevit tests and all have passed. I also test the old script with this change, it can run without error or warning.
test

chuongmep:
I think we not necessary change name to class because engineer difficult for read document update .

ZiyunShang:
> I think we not necessary change name to class because engineer difficult for read document update .

But this does mislead some users and confuse them like https://github.com/DynamoDS/DynamoRevit/issues/2587 . So I think this will be helpful for distinguishing Element Class and ElementType.

PR #2659, opened 2020-12-21
Add ElementBinder to AnalysisDisplay Nodes

ZiyunShang:
### Purpose

Add ElementBinder to AnalysisDisplay Nodes avoid these nodes will clear SpatialFieldManager every time.

### Declarations

### Reviewers

@wangyangshi @ShengxiZhang @mjkkirschner @QilongTang 

### FYIs
@Amoursol

### DIFF
--- a/src/Libraries/RevitNodes/AnalysisDisplay/PointAnalysisDisplay.cs
+++ b/src/Libraries/RevitNodes/AnalysisDisplay/PointAnalysisDisplay.cs
@@ -16,43 +16,6 @@ using View = Revit.Elements.Views.View;
 
 namespace Revit.AnalysisDisplay
 {
-    [SupressImportIntoVM]
-    [Serializable]
-    public class SpmPrimitiveIdListPair : ISerializable
-    {
-        public int SpatialFieldManagerID { get; set; }
-        public List<int> PrimitiveIDs { get; set; }
-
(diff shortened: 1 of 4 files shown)


src/Libraries/RevitNodes/AnalysisDisplay/PointAnalysisDisplay.cs
```diff
@@ -16,43 +16,6 @@
 
 namespace Revit.AnalysisDisplay
 {
-    [SupressImportIntoVM]
-    [Serializable]
-    public class SpmPrimitiveIdListPair : ISerializable
```
mjkkirschner, line 21: for your team to consider how you treat semantic versioning, API compatibility etc, but you might want to consider obsoleting this class instead of deleting it.

I'm assuming you're thinking of treating Dynamo4Revit APIs the same way that Revit treats apis? Meaning it's okay to break them between releases, or after a release where the API is obsolete?

Anyway, something to think about.

ZiyunShang: @mjkkirschner I just found that this class has never been used in code, but I think you are right, I will set it obsolete first.

wangyangshi (approved):

PR #2702, opened 2021-05-07
Define input output type of UI nodes

ZiyunShang:
### Purpose

Add input/output type for UI Nodes

### Declarations

### Reviewers

@wangyangshi @ShengxiZhang @QilongTang 

### FYIs

@mjkkirschner @Amoursol

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitTypes.cs
+++ b/src/Libraries/RevitNodesUI/RevitTypes.cs
@@ -24,6 +24,7 @@ namespace DSRevitNodesUI
     [NodeCategory("Revit.Elements.Phase")]
     [NodeDescription("PhaseSelectorDescription", typeof(DSRevitNodesUI.Properties.Resources))]
     [IsDesignScriptCompatible]
+    [OutPortTypes("Revit.Elements.Phase")]
     public class RevitPhases : CustomRevitElementDropDown
     {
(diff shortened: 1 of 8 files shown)


src/Libraries/RevitNodesUI/RevitTypes.cs
```diff
@@ -69,6 +72,7 @@ public FilledRegionTypes(IEnumerable<PortModel> inPorts, IEnumerable<PortModel>
     [NodeCategory("Revit.Filter.RuleType")]
     [NodeDescription("FilterTypeSelectorDescription", typeof(DSRevitNodesUI.Properties.Resources))]
     [IsDesignScriptCompatible]
+    [OutPortTypes("string")]
```
ZiyunShang, line 75: Hi @QilongTang , here are some nodes have "string" as output, I just add this, but don't work. Any suggestions?

wangyangshi (approved):

ShengxiZhang (approved):

src/Libraries/RevitNodesUI/RevitTypes.cs
QilongTang: Right because right now, there are too many nodes with type as string. We will most likely enable the suggestion later with machine learning

PR #2720, opened 2021-06-25
WIP - Add option to disable the TransactionManager in D4R globally.

saintentropy:
### Purpose

[AGD-2201](https://jira.autodesk.com/browse/AGD-2201). This task introduces the ability to disable the TransactionManager globally.  This mode will support disabling nodes or python code utilizing the TransactionManager that creates, modifies, or deletes the Revit elements.  This PR also adjust the SafeInit() function.  SafeInit previously wrapped all initialization functions in a Revit transaction regardless if the operation was a DB query only.  This change will also need to be communicated to third party Revit packages developers.  

### Declarations

### Reviewers

@ZiyunShang  @BogdanZavu 

### FYIs

@QilongTang @ShengxiZhang


src/Libraries/RevitServices/Transactions/TransactionManager.cs
```diff
@@ -96,13 +98,16 @@ private TransactionManager(ITransactionStrategy strategy)
...
+            ReadOnlyMode = false;
         }
 
         /// <summary>
         /// Ensure that there is a valid transaction active
         /// </summary>
         public void EnsureInTransaction(Document document)
         {
+            if (ReadOnlyMode)
+                throw new Exception("Dynamo For Revit is in read-only mode");
```
ZiyunShang, line 110: I I wonder how this exception will behave? Will Node turn yellow and display this information? For a Python node or a Custom node, how will it be displayed?

saintentropy: For a typical node it will turn yellow and display this error.   The python and custom nodes do their typical behavior of suppressing the specific error message.  I will double check was the user facing UX is

saintentropy: For now with this internal method we are thinking that the application is only for running the graph in headless mode where the user would not see the impact.

saintentropy: Also I am going to update the message to be a bit more precise

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -32,6 +32,24 @@ protected void SafeInit(Action init)
...
 
+            SafeInitImpl(init);
+        }
+
+        /// <summary>
+        /// Handling exceptions when calling the initializing function
+        /// </summary>
+        /// <param name="init"></param>
+        /// <param name="fromExistingElement"></param>
+        protected void SafeInit(Action init, bool fromExistingElement)
```
BogdanZavu, line 43: Why fromExistingElement ? I was expecting something like 'ensureInTransaction' Am I missing something ?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -32,6 +32,24 @@ protected void SafeInit(Action init)
         {
             TransactionManager.Instance.EnsureInTransaction(DocumentManager.Instance.CurrentDBDocument);
 
+            SafeInitImpl(init);
+        }
+
+        /// <summary>
```
BogdanZavu, line 38: Shouldn't we handle this the other way around and have a default param set to true to ensureTransaction ?

ZiyunShang:
Well, I just run the Revit's Regression tests on farm, all passed. Any updates about the param issues Bogdan mentioned?
BTW, is it feasible to add some testing of DisableTransactions?

ZiyunShang (approved):

PR #2735, opened 2021-07-29
Improve some node API

ZiyunShang:
### Purpose

1. Add exclusive RevisionNumberingSequence for Revision creation node.
2. Add offset for Floor creation API because of updated Revit API.
3. Add non-null judgment for ExtractSurface of RuledFace.

### Declarations

### Reviewers

@ShengxiZhang @wangyangshi 

### FYIs

@QilongTang @mjkkirschner @Amoursol

### DIFF
--- a/src/Libraries/RevitNodes/Elements/Floor.cs
+++ b/src/Libraries/RevitNodes/Elements/Floor.cs
@@ -35,6 +35,8 @@ namespace Revit.Elements
             get { return InternalFloor; }
         }
 
+        private static readonly double Tolerance = 1e-6;
+
         #endregion
 
         #region Private constructors
@@ -73,12 +75,12 @@ namespace Revit.Elements
         private void InitFloor(List<CurveLoop> profiles, Autodesk.Revit.DB.FloorType floorType, Autodesk.Revit.DB.Level level, double offset = 0)
         {
             TransactionManager.Instance.EnsureInTransaction(Document);
-
+            
             // we assume the floor is not structural here, this may be a bad assumption
             Autodesk.Revit.DB.Floor floor = Autodesk.Revit.DB.Floor.Create(Document, profiles, floorType.Id, level.Id);
             var param = floor.get_Parameter(BuiltInParameter.FLOOR_HEIGHTABOVELEVEL_PARAM);
             
-            if(param !=null && offset != 0)
+            if(param !=null && Math.Abs(offset - 0) > Tolerance)
             {
                 InternalUtilities.ElementUtils.SetParameterValue(param, offset);
             }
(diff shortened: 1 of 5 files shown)


src/Libraries/RevitNodes/Elements/Floor.cs
```diff
@@ -70,13 +70,18 @@ private void InitFloor(Autodesk.Revit.DB.Floor floor)
...
+        private void InitFloor(List<CurveLoop> profiles, Autodesk.Revit.DB.FloorType floorType, Autodesk.Revit.DB.Level level, double offset = 0)
         {
             TransactionManager.Instance.EnsureInTransaction(Document);
 
             // we assume the floor is not structural here, this may be a bad assumption
             Autodesk.Revit.DB.Floor floor = Autodesk.Revit.DB.Floor.Create(Document, profiles, floorType.Id, level.Id);
-
+            var param = floor.get_Parameter(BuiltInParameter.FLOOR_HEIGHTABOVELEVEL_PARAM);
+            var paramName = param.ToString();
+            if(param.AsDouble() == 0 && offset != 0)
```
ShengxiZhang, line 81: Do you mean `if (param.AsDouble() != offset)` ? And where do we use the variable `paramName`?

ZiyunShang: According to the new Revit API, this parameter of new Floor should be 0. May be we should add an exception for it?
the "paramName" used to get and set Element's parameter, but I found that ElementUtils can be used directly. So it should be removed.

src/Libraries/RevitNodes/Elements/Revision.cs
```diff
@@ -167,12 +167,28 @@ private static ElementId GetRevisionNumberingSequence(RevisionNumberType numberT
...
+                switch (numberType)
+                {
+                    case RevisionNumberType.Alphanumeric:
+                        revisionNumberingSequence = RevisionNumberingSequence.CreateAlphanumericSequence(doc, "Dynamo_Alphanumeric", new AlphanumericRevisionSettings());
+                        revisionNumberingSequenceId = revisionNumberingSequence.Id;
+                        break;
+                    case RevisionNumberType.Numeric:
+                        revisionNumberingSequence = RevisionNumberingSequence.CreateNumericSequence(doc, "Dynamo_Numeric", new NumericRevisionSettings());
+                        revisionNumberingSequenceId = revisionNumberingSequence.Id;
+                        break;
```
ShengxiZhang, line 188: no default branch?

ZiyunShang: If not Alphanumeric or Numeric, all return invalidElementId. And revisionNumberingSequenceId was set to invalid element id when it initializes.

ShengxiZhang (approved):

wangyangshi (approved):

anshivashankar (approved):

PR #2739, opened 2021-08-11
do not import CurvePartition into the VM

mjkkirschner:
While working on a documentation generator I noticed that this class was getting imported into Dynamo - it should likely be internal or suppressed from import like the CurvePartitionLoop class above.

@ZiyunShang please take a look.

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/CurveLoopPartition.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/CurveLoopPartition.cs
@@ -125,7 +125,7 @@ namespace Revit.GeometryConversion
         }
     }
 
-    public class CurvePartition
+    internal class CurvePartition
     {
         public List<Curve> OuterCurves;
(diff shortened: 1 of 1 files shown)


ZiyunShang (approved):

wangyangshi (approved):

src/Libraries/RevitNodes/GeometryConversion/CurveLoopPartition.cs
```diff
@@ -125,7 +125,7 @@ private static bool IsContainedIn(this IEnumerable<XYZ> curveLoop, BoundingBox b
         }
     }
 
-    public class CurvePartition
+    internal class CurvePartition
```
ZiyunShang, line 128: Does it need the [SupressImportIntoVM] tag here as well?

mjkkirschner:
Either will stop it from being imported.

ShengxiZhang (approved):

PR #2761, opened 2021-11-02
Keep the revit element with Node when Disable Transaction

ZiyunShang:
### Purpose

After add Transaction control, an issue was found that element binding was broken when disable and enable transaction. So we need to keep the Revit elementbinder with D4R nodes.
According to design, when a node runs a Revit element, then disable transaction, the node should not turn yellow and display warning because it's not user error.

### Declarations

### Reviewers

@ShengxiZhang @wangyangshi @saintentropy 

### FYIs

@QilongTang @mjkkirschner @Amoursol

### DIFF
--- a/src/Libraries/RevitServices/Persistence/ElementIDLifecycleManager.cs
+++ b/src/Libraries/RevitServices/Persistence/ElementIDLifecycleManager.cs
@@ -100,8 +100,13 @@ namespace RevitServices.Persistence
                 //ID already existed, check we're not over adding
                 if (existingWrappers.Contains(wrapper))
                 {
-                    int index = existingWrappers.FindIndex((x) => object.ReferenceEquals(x, wrapper));
-                    existingWrappers.RemoveAt(index);
+                    var removeList = existingWrappers.FindAll((x) => object.ReferenceEquals(x, wrapper));
+                    for (int i = 0; i < removeList.Count; i++)
+                    {
+                        int index = existingWrappers.FindIndex((x) => object.ReferenceEquals(x, wrapper));
+                        existingWrappers.RemoveAt(index);
+                    }
+                    
                     if (existingWrappers.Count == 0)
                     {
                         wrappers.Remove(elementID);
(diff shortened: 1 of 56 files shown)


src/Libraries/RevitServices/Persistence/ElementIDLifecycleManager.cs
```diff
@@ -100,8 +100,13 @@ public int UnRegisterAssociation(T elementID, Object wrapper)
                 //ID already existed, check we're not over adding
                 if (existingWrappers.Contains(wrapper))
                 {
-                    int index = existingWrappers.FindIndex((x) => object.ReferenceEquals(x, wrapper));
-                    existingWrappers.RemoveAt(index);
+                    var removeList = existingWrappers.FindAll((x) => object.ReferenceEquals(x, wrapper));
```
mjkkirschner, line 103: what case triggered this change? Could index ever be -1? (not found?)

ZiyunShang: When I debug disable/enable transaction, I found that existingWrappers contains multiple wrapper, but no one can be referenceEqueals.

ZiyunShang: This may be debugging code, I will solve this, but I think it would be better to add a judgment protection to the index.

ShengxiZhang: Thank you @ZiyunShang for the update.
I still have concern on this implementation. This requires Dynamo developer to keep in mind that the node may run in DisableTransaction mode, and need to deal with the behavior. And it also requires Dynamo developers to be familiar with the API internal implementation, know if the API calls regen or starts Transaction internally.
Moreover, even doing all above, this only works for the built-in nodes. It doesn't work for package nodes, python nodes, custom nodes, which are also in the AC list. 
@saintentropy @QilongTang any ideas?

PR #2770, opened 2021-12-02
ForgeType Refactor

saintentropy:
### Purpose

The purpose of this PR is to refactor the user presentation of the `ForgeTypeID` object found in the RevitAPI for the DynamoRevit context.  Note, these nodes have not be included in a Revit Release yet so all changes are not API breaking.  In general, the idea is to reduce the confusion between nodes which require a `ForgeTypeID` object as an input but in reality need a specific type of `ForegTypeID`.  For example a node like `CreateSharedParameter` requires two `ForgeType` inputs but they need to representing an underlying `SpecType` and `GroupType` respectively.  This can be confusing to a Dynamo user as there is no clear way in the Dynamo type system to understand the distinction.  This issue can be additionally confusing as some specs and groups share the same name like Length or Area.  The other reason for this change is to better reflect the way that Dynamo Core has wrapped the `ForgeUnit` types.  While the underlying data is the same as the `SpecType` and `GroupType` (ie a ForgeType object), the Dynamo Core implementation provides unique types for Unit, Quantity, and Symbol.  This change will better match the core implementation.  Specifically this PR does the follwoing: 

* Add two new unique types.  `SpecType` and `GroupType` based on the `ForgeType` as a baseclass.  
* The `ForgeType` now only has properties (`TypeID` and `IsValidType`).  The constructors are found on the derived classes.
* The constructors for the `GroupType` and `SpecType` objects guard against ForgeTypeID strings and objecst that are not valid for their respective types. 
* All nodes that require inputs of a `ForgeType` now require a `GroupType` or `SpecType`.
* All nodes that return `ForgeType` now return a `GroupType` or `SpecType` as required.

This PR also adjusts the `ToSting()` override used for displaying ForgeType object in the Dynamo preview.  Where possible we show the name of the object vs the `TypeId`.  The `TypeId` is still available as a property of the type but is not as friendly to expose as a name.  So now the preview will show _SpecType(Name = Length)_ vs _autodesk.spec.aec:length-2.0.0_ as an example.

### Declarations

### Reviewers

@ShengxiZhang @ZiyunShang @dbecroft 

### FYIs

@QilongTang @Amoursol @mjkkirschner

### DIFF
--- a/src/DynamoRevit/Resources/LayoutSpecs.json
+++ b/src/DynamoRevit/Resources/LayoutSpecs.json
@@ -185,6 +185,9 @@
                   "path": "RevitNodes.Revit.Elements.Group"
                 },
                 {
+                  "path": "RevitNodes.Revit.Elements.GroupType"
+                },
+                {
                   "path": "RevitNodes.Revit.Elements.ImportInstance"
                 },
                 {
@@ -228,6 +231,9 @@
                 },
                 {
                   "path": "RevitNodes.Revit.Elements.Space"
+                },
+                {
+                  "path": "RevitNodes.Revit.Elements.SpecType"
                 },
                 {
                   "path": "RevitNodes.Revit.Elements.StructuralFraming"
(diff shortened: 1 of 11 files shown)


ZiyunShang:
LGTM, and I think there are some test file should update to this change -
test/System/FamilyDocument/CanAddParameterToFamilyDocument.dyn,
test/System/Parameter/GlobalParameter.dyn

ZiyunShang (approved):

PR #2791, opened 2022-01-19
Add converter for internal Revit unit types not found ForgeUnit SDK

saintentropy:
### Purpose

The purpose of this PR is to add support for the internal Revit Unit types (handwritten) not found in the ForgeUnitSDK. 

These Unit types are: `DegreesMinutes`, `FeetFractionalInches`, `FractionalInches`, `MetersCentimeters`, `SlopeDegrees`, `StationingFeet`, `StationingMeters`, and `StationingSurveyFeet`

This PR introduces a converter function that will convert these to their equivalent ForgeUnitSDK types.

### Declarations

### Reviewers

@ShengxiZhang @ZiyunShang @dbecroft

### FYIs

@nate-peters

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
@@ -24,6 +24,41 @@ namespace Revit.GeometryConversion
             return UnitUtils.ConvertToInternalUnits(1, unitTypeId);
         }
 
+        public static ForgeTypeId ConvertRevitInternalUnitTypeIdToSupportedUnitTypeId(ForgeTypeId forgeTypeId)
+        {
+            //Handle conversion of the Revit Handwritten units to UnitsSDK equivalents
+
+            if (forgeTypeId.Equals(UnitTypeId.FeetFractionalInches)
+                || forgeTypeId.Equals(UnitTypeId.StationingFeet))
+            {
+                return UnitTypeId.Feet;
+            }
+
+            if (forgeTypeId.Equals(UnitTypeId.StationingSurveyFeet))
+            {
+                return UnitTypeId.UsSurveyFeet;
+            }
+
+            if (forgeTypeId.Equals(UnitTypeId.FeetFractionalInches))
+            {
+                return UnitTypeId.Inches;
+            }
+
+            if (forgeTypeId.Equals(UnitTypeId.MetersCentimeters)
+                || forgeTypeId.Equals(UnitTypeId.StationingMeters))
+            {
+                return UnitTypeId.Meters;
+            }
+
+            if (forgeTypeId.Equals(UnitTypeId.DegreesMinutes)
+                || forgeTypeId.Equals(UnitTypeId.SlopeDegrees))
+            {
+                return UnitTypeId.Degrees;
+            }
+
+            return forgeTypeId;
+        }
+
         public static double HostToDynamoFactor(ForgeTypeId specTypeId)
         {
             // Here we invert the conversion factor to return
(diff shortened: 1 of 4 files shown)


saintentropy:
We found this in `Document.UnitTypeBySpecType` node but it also may be found in `Parameter.Unit` and `FamilyParameter.Unit`.

src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
```diff
@@ -24,6 +24,41 @@ public static double DynamoToHostFactor(ForgeTypeId specTypeId)
             return UnitUtils.ConvertToInternalUnits(1, unitTypeId);
         }
 
+        public static ForgeTypeId ConvertRevitInternalUnitTypeIdToSupportedUnitTypeId(ForgeTypeId forgeTypeId)
```
ZiyunShang, line 27: Is this an issue for Revit API? For the following types, is "UnitTypeId.FeetFractionalInches" and "UnitTypeId.StationingFeet" the same as "UnitTypeId.Feet" in Revit?

Another question, is this should cherry pick to Revit 2023?

saintentropy: I think it is not an issue when using the Revit API by itself.  It only becomes an issue when converting between Revit API units and those found in the ForgeUnitSDK schemas.  There seem to be a few special cases but  I am going to confirm with the Omicron team.  

I would say yes to cherry pick once this PR is approved

dbecroft:
Good morning Craig, one other identifier to keep in mind is `UnitTypeId.Custom`. That's Revit's way of saying "I don't know what this unit is", and if you get a call like `Unit.ByTypeID(UnitTypeId.Custom)` it would be right to throw an exception. 

The proposed conversions are sensible. Revit has an unfortunate legacy of mixing number formats into its units of measurement. In Forge Units, a unit represents a magnitude regardless of whether that magnitude is written in fractional notation, fixed-point notation, or something else. Most of these equivalences arise from that. The "stationing" units are also just a special number format. 

Just FYI, `SlopeDegrees` are a little bit different. Swapping them for vanilla `degrees` as in this PR is correct, but `SlopeDegrees` are not just a special number format. In Forge Units, `degrees` are derived from the primitive unit `radians` and can be converted to other units of angle. They cannot be converted directly to units derived from the primitive unit `ratioTo1`. In Revit, `SlopeDegrees` is convertible directly to rise-over-run slope units that are derived from `ratioTo1`. Revit achieves that with the help of some trigonometry that is not included in Forge Units.

dbecroft (approved):

src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
```diff
@@ -24,6 +24,48 @@ public static double DynamoToHostFactor(ForgeTypeId specTypeId)
...
+                return UnitTypeId.Meters;
+            }
+
+            if (forgeTypeId.Equals(UnitTypeId.DegreesMinutes)
+                || forgeTypeId.Equals(UnitTypeId.SlopeDegrees))
+            {
+                return UnitTypeId.Degrees;
+            }
+
+            return forgeTypeId;
```
ShengxiZhang, line 66: Would a `Dictionary` be more readable than multiple `if` blocks?

ShengxiZhang (approved):

ZiyunShang (approved):

ZiyunShang:
Hi @saintentropy , is there any further update here? Or I can merge this?

saintentropy:
@ZiyunShang I am fine merging... I can take a look at the Dictionary later but I don't want to block this

PR #2826, opened 2022-05-31
fix the issue revit-191624

honghuiqin:
### Purpose

This PR is to fix the issue  "https://jira.autodesk.com/browse/REVIT-191642", d4r cannot set parameter value of integer type in Dynamo 2.13 for Revit 2023 and D4R new Warning messages about the ForgeUnits changes.

### Reviewers

@wangyangshi 
@ShengxiZhang

### DIFF
--- a/src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
+++ b/src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
@@ -12,13 +12,24 @@ namespace Revit.GeometryConversion
     {
         public static double DynamoToHostFactor(ForgeTypeId specTypeId)
         {
+<<<<<<< Updated upstream
             var units = DocumentManager.Instance.CurrentDBDocument.GetUnits();
+=======
+>>>>>>> Stashed changes
             if (!UnitUtils.IsMeasurableSpec(specTypeId))
             {
                 return 1;
             }
+<<<<<<< Updated upstream
 
             var unitTypeId = units.GetFormatOptions(specTypeId).GetUnitTypeId();
+=======
+            var unitTypeId =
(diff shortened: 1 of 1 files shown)


src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
```diff
@@ -12,11 +12,13 @@ public static class UnitConverter
...
-                DocumentManager.Instance.CurrentDBDocument.GetUnits()
-                    .GetFormatOptions(specTypeId)
-                    .GetUnitTypeId();
-           
+            var units = DocumentManager.Instance.CurrentDBDocument.GetUnits();
+            if (!UnitUtils.IsMeasurableSpec(specTypeId))
+            {
+                return 1;
+            }
+
```
wangyangshi, line 20: can we just add the condition check first in this function and no change to later codes.

src/Libraries/RevitNodes/GeometryConversion/UnitConverter.cs
```diff
@@ -12,6 +12,10 @@ public static class UnitConverter
     {
         public static double DynamoToHostFactor(ForgeTypeId specTypeId)
         {
+            if (!UnitUtils.IsMeasurableSpec(specTypeId))
+            {
+                return 1;
```
ShengxiZhang, line 17: better to `return 1.0;` since return type is double.

ShengxiZhang: By the way, I think it may be better to move this check to the caller `SetParameter`, and leave this function just getting the factor. But if there are too many place to modify, we can keep it as is for now.

ShengxiZhang (approved):

PR #2879, opened 2022-11-16
add a step of finding RevitAPI

wangyangshi:
### Purpose

description of how to find RevitAPI in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ git clone https://github.com/DynamoDS/DynamoRevit.git
 - Make sure you have the following installed on your computer:
    - [.Net Framework 4.8 SDK](https://dotnet.microsoft.com/download) 
 - Run `restorepackages.bat` from a command prompt with administrative privileges (Located in your _Github\DynamoRevit\src folder_)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,6 +21,7 @@ git clone https://github.com/DynamoDS/DynamoRevit.git
 - Make sure you have the following installed on your computer:
    - [.Net Framework 4.8 SDK](https://dotnet.microsoft.com/download) 
 - Run `restorepackages.bat` from a command prompt with administrative privileges (Located in your _Github\DynamoRevit\src folder_)
+- Copy `RevitAPI.dll`&`RevitAPIUI.dll` to the folder `DynamoRevit\lib\Revit Preview Release\net48`, these 2 dlls are in the folder same with `Revit.exe` installed on your computer
```
QilongTang, line 24: I think we want to list a few cases here:
1. If you have the corresponding Revit installed
2. If you do not have Revit installed, how to get the dll and copy them?
e.g. I do think I had to do the copy step as long as I have Revit installed?

wangyangshi: Generally, the DLLs should be get from installed Revit, 
because if user's local machine doesn't install the corresponding version of Revit, 
then the change in d4r cannot be tested locally.

PR #2880, opened 2022-11-20
Fix issue with localization for SelectByCategory Nodes

saintentropy:
### Purpose

The purpose of this PR is to address an issue observer when switching between Revit localizations with a the SelectByCategory node.  The node does not deserialize to the correct selected filter categroy if when opened in a different localization from where it was saved.

### Declarations

### Reviewers

@BogdanZavu @twastvedt 

### FYIs

### DIFF
--- a/src/Libraries/RevitNodesUI/SelectionCombo.cs
+++ b/src/Libraries/RevitNodesUI/SelectionCombo.cs
@@ -64,6 +64,11 @@ namespace Dynamo.ComboNodes
             set
             {
                 DropDownNodeModel.SelectedString = value;
+
+                if (DropDownNodeModel.SelectedIndex >= 0)
+                {
+                    SelectionFilter.Category = (BuiltInCategory)DropDownNodeModel.Items[SelectedIndex].Item;
+                }
             }
         }
 
@@ -133,6 +138,11 @@ namespace Dynamo.ComboNodes
             set
             {
                 DropDownNodeModel.SelectedString = value;
+
+                if (DropDownNodeModel.SelectedIndex >= 0)
+                {
+                    SelectionFilter.Category = (BuiltInCategory)DropDownNodeModel.Items[SelectedIndex].Item;
+                }
             }
         }
 
--- a/test/Libraries/RevitIntegrationTests/SelectionTests.cs
+++ b/test/Libraries/RevitIntegrationTests/SelectionTests.cs
@@ -619,7 +619,7 @@ namespace RevitSystemTests
         public void SelectModelElementsByCategory()
         {
             OpenAndAssertNoDummyNodes(Path.Combine(workingDirectory, @".\Selection\SelectModelElementsByCategory.dyn"));
-            TestMultipleCategorySelection<Element, Element>();
+            TestMultipleCategorySelection<Element>();
         }
 
         [Test]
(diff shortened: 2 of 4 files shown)


twastvedt (approved):

saintentropy:
@wangyangshi are there any tests that currently test a localization issue like this one?

saintentropy:
@wangyangshi can you confirm that the new and modified tests are running correctly?

wangyangshi:
Yes, the system/unit test can be searched in the project, test calls the node function

src/Libraries/RevitNodesUI/SelectionCombo.cs
```diff
@@ -64,6 +64,11 @@ public string SelectedString
             set
             {
                 DropDownNodeModel.SelectedString = value;
+
+                if (DropDownNodeModel.SelectedIndex >= 0)
```
mjkkirschner, line 68: is the idea here that in different localizations that the index will be more reliable than the string? Doesn't that undo the purpose of using the string to help when the indicies change between Revit versions? FYI @QilongTang 

Maybe I am misunderstanding.

wangyangshi:
Yes, let me run the test today, once I get the results I will update it here.

src/Libraries/RevitNodesUI/SelectionCombo.cs
QilongTang: Yes, this would cause regression for cases that category index change across Revit versions. Also if we want to do this in the setter, I would rather we obsolete `SelectedString`, no point keeping the property if not going to use it.

wangyangshi: Just curious
when changing selected-string can selected-index be changed accordingly,
when changing selected-index can selected-string be changed accordingly? 
is this realized by dynamo core side?

saintentropy: @mjkkirschner I think it is the opposite.  SelectedIndex is not reliable between localization or Revit files.  Strings are the same but order of collection is different.  For the drop down nodes in Revit for deserialization, the `SelectedIndex` is first `Set` if it exists in the collection.  Index alone does not guarantee alignment.  The `SelectedString` is `Set`...  It checks the alignment of strings (which are all English strings) and it will Re-`Set` the selected index.  That step is what I am adding here in this PR.

mjkkirschner: (maybe it makes sense to add a property changed handler to the DropDownNodeModel and watch for selected string and index prop updates to centralize this.

mjkkirschner (approved):

jnealb:
Just an fyi: the Jira ticket for this is https://jira.autodesk.com/browse/AGD-2087

wangyangshi (approved):

ShengxiZhang (approved):

src/Libraries/RevitNodesUI/SelectionCombo.cs
QilongTang: @saintentropy Thank you for the explanation

QilongTang (approved):

test/Libraries/RevitIntegrationTests/SelectionTests.cs
```diff
@@ -896,6 +904,10 @@ private void AssertElementCategoryOnCollection(List<object> collection, string e
                 var elem = elements[i] as Revit.Elements.Element;
                 Assert.AreEqual(expectedCategoryString, elem.GetCategory.Name);
             }
+
+            var selectionFilter = selectNode.Filter as CategoryElementSelectionFilter<Element>;
+            Assert.IsTrue(selectionFilter.Category == BuiltInCategory.OST_Walls);
+
```
wangyangshi, line 910: I tested it today locally, ran the system test, but it failed as there is an assert failed. 
failed at `Assert.NotNull(elements);`, I will check it again.

wangyangshi: > @wangyangshi can you confirm that the new and modified tests are running correctly?


I merged it as want to integrate this fix to Revit2024 although the test failed.
image
from debug, the elements is null, this is unexpected compared with other tests, like `SelectModelElementsByCategory`
image
@saintentropy any ideas about this?

PR #2882, opened 2022-11-28
change int to long in some d4r tests

wangyangshi:
### Purpose

1 some d4r test codes use the ElementId(int) to construct element id, this will be obsoleted.
so change them to long.

2 change `SpatialFieldManagerID` to long, as in revit codes, I found they are regarded as element id. so also change it.


awbushnell (approved):
Seeem good to me. Thank you!

ShengxiZhang (approved):

PR #2885, opened 2022-12-06
do not copy newtonsoft.json to Revit/Nodes folder

mjkkirschner:
### Purpose

unify references to not include version number
do not copy to nodes folder

⚠️ note
If you are testing this change locally I had to manually delete the old copy of newtonsoft.json from the Revit.Nodes folder - a rebuild would not delete it for some reason.

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitNodesUI.csproj
+++ b/src/Libraries/RevitNodesUI/RevitNodesUI.csproj
@@ -39,9 +39,9 @@
       <HintPath>$(PACKAGESPATH)\DynamoVisualProgramming.Core\lib\net48\DynamoUtilities.dll</HintPath>
       <Private>False</Private>
     </Reference>
-    <Reference Include="Newtonsoft.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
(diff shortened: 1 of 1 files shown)


QilongTang (approved):

QilongTang:
@wangyangshi if there will be more hotfix for past D4R releases, you may want to consider hotfix as well.

wangyangshi:
will this fix the steel connection issue in dynamo 2.17?
I am not very clear why we need to change it in earlier D4R releases, because in before versions, it has no issues.
if my understanding not right, please correct me, thank you

QilongTang:
> will this fix the steel connection issue in dynamo 2.17? I am not very clear why we need to change it in earlier D4R releases, because in before versions, it has no issues. if my understanding not right, please correct me, thank you

I would say in short this implementation will easily cause dependency loading conflicts. This could cause problems whenever there is a version mismatch of NewtonSoft.Json dll. The bug reported by AS team is just one instance where other users could run into but did not report it. Since this is a compile-time preference, as @mjkkirschner  mentioned, let's not copy this dll to nodes folder, because NewtonSoft.Json is almost guaranteed to be loaded by Dynamo or Revit already, there is no need to load another hard copy of it.

wangyangshi:
OK, understand it, thank you Aaron for the explanation!

wangyangshi:
after testing, this fixed the regression issue, can it be merged now?

QilongTang:
> after testing, this fixed the regression issue, can it be merged now?

Yeah if you approve the PR. Feel free

PR #2757, opened 2021-10-28
Use OAuth2 provider to communicate with package manager

mjkkirschner:
### Purpose

* use new GregRevitAuth OAuth2 provider
* don't force ssonet to load, it should already be loaded in newer versions of Revit by default.

### TODO:
- [x] Actually update Greg and GregRevitAuth nuget versions when published.

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [x] User facing strings, if any, are extracted into `*.resx` files
- [x] Snapshot of UI changes, if any.

PR #2887, opened 2022-12-12
Greg update

wangyangshi:
### Purpose
https://jira.autodesk.com/browse/REVIT-196808
1 update greg version in Revit 2022
2 newtonsoft references Revit 2022

### Declarations
reference:
https://github.com/DynamoDS/DynamoRevit/pull/2885
https://github.com/DynamoDS/DynamoRevit/pull/2757
https://github.com/DynamoDS/DynamoRevit/pull/2756

### DIFF
--- a/src/Config/packages.aget
+++ b/src/Config/packages.aget
@@ -9,14 +9,14 @@
       "DynamoVisualProgramming.WpfUILibrary": "2.12.1.8246",
       "DynamoVisualProgramming.ZeroTouchLibrary": "2.12.1.8246",
       "DynamoVisualProgramming.DynamoSamples": "2.12.1.8246",
-      "Greg": "2.0.7507.22529",
-      "GregRevitAuth": "1.0.7057.20655",
+      "Greg": "2.1.7972.24350",
+      "GregRevitAuth": "1.1.7972.27909",
       "IronPython": "2.7.9",
       "pythonnet_py38_win": "2.5.1",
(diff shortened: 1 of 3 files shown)


mjkkirschner:
hmm, I think using these version numbers is safe - we know they worked together in 2023.

@wangyangshi I am not sure that updating to use `RevitOAuth2Provider` will work in Revit 2022, 21 etc - I don't think that is actually necessary as we did not remove the `RevitOxygenProvider` from GregRevitAuth.dll

I think either is fine if you are sure that Revit 2022,21, 20's SSONet provides the correct methods to get an oauth2 token etc. 

Will Revit QA have time to test the following?
1. test package upload (can be done against package manager dev by just changing the DynamoPackages.config file to point to `https://dev.dynamopackages.com`
2. retest the original bug - IE load SpaceMaker addin first, start it, then try to start DynamoRevit and confirm it starts correctly.

I think testing in each hotfix version - if there are going to be others to 2021,20 etc would be ideal.

wangyangshi:
@mjkkirschner OK, what about keep `RevitOxygenProvider`, I want to change less, affect less to old releases.

mjkkirschner:
> @mjkkirschner OK, if `RevitOxygenProvider` can still work, what about keep it, I want to change less, affect less to old releases.

I think that makes sense for hot fixes.

mjkkirschner (approved):

PR #2889, opened 2022-12-13
Update 2021 Greg version

wangyangshi:
### Purpose
Update Greg version in 2021,
similar to 
https://github.com/DynamoDS/DynamoRevit/pull/2887


mjkkirschner (approved):

wangyangshi:
@mjkkirschner Can I use below versions?

      "Greg": "1.2.7177.24952",
      "GregRevitAuth": "1.2.8382.31161"

      
I tried to change Greg version to "2.1.7972.24350", but build failed.

mjkkirschner:
Hi @wangyangshi let me look into this - please recall that Greg version is really set from Dynamo - I believe that the Greg.dll is actually delivered from DynamoCoreRuntime - so I will check what version is referenced there.

wangyangshi:
Michael thank you very much for your help, some information:
this branch's dynamo core's version is 2.6.1.8786
the Greg.dll in it version is 2.2.0.1080
image

mjkkirschner:
@wangyangshi I think we should just use that version (and retest!) - in addition perhaps even on the 2022/2020 branch we should revert the changes to the greg.dll and retest there to keep it aligned with the version Dynamo is delivering since it doesn't get delivered anyway.

Also - I have one more concern - why are we modifying these branches which are already released instead of creating new 2022.1.4 etc branches so we can easily see what the released state was?

mjkkirschner:
@wangyangshi - in 2022 - strangely Dynamo 2.12.1 references/delivers an older Greg version 2.1.7962 - in this version the oauth2 interface does not exist, so it was a good choice not to move to oauth2 provider!

definitely lets roll back to `2.1.7962` so they are aligned and we don't try to use APIs that won't be there at runtime.

wangyangshi:
Thank you for your remind, these 3 branches are for new release
Revit 2022.1.4
Revit 2021.1.8
Revit 2020.2.10
I thought the old branches won't be changed as they have been released, so just changed in them, 
But your suggestion is right, I should create new branches to change these version for Revit new releases.
Thank you, I will close these PRs and create new branches and new PRs.

mjkkirschner:
2020 references Greg.2.2.0.1080 as well, so those are aligned between 2021 and 2020.
but 2022 references an older `2.1.7962` as stated above - this is a bit weird, but I don't think it causes a problem - and to avoid making another dynamo core release - I think all that needs to be done is that DynamoRevit should just match that version - Greg 2.1.7962 - that way we don't use any newer APIs that won't exist at runtime. (like the oauth2provider)

wangyangshi:
> @wangyangshi - in 2022 - strangely Dynamo 2.12.1 references/delivers an older Greg version 2.1.7962 - in this version the oauth2 interface does not exist, so it was a good choice not to move to oauth2 provider!
> 
> definitely lets roll back to `2.1.7962` so they are aligned and we don't try to use APIs that won't be there at runtime.

Thank you, so for 3 new branches I just change:
Greg to the version that its Dynamo core really used, 
2022 : `2.1.7962.24801`
2021/2020 : `2.2.0.1080`
And GregRevitAuth to the version
2022 : `1.1.7972.27909`
2021/2020 : `1.2.8382.31161`

mjkkirschner:
@wangyangshi that looks correct to me - let me know when these builds are up and I will try to do extra testing on them.

PR #2890, opened 2022-12-13
Update greg in 2020

wangyangshi:
### Purpose
Update Greg version in 2020,
similar to
https://github.com/DynamoDS/DynamoRevit/pull/2887


mjkkirschner:
Does this build successfully? Since this branch is very old the target framework was .net 4.7 - I believe that was the .net runtime Revit 2020 hosted as well, this should definitely get tested with the 2 manual  tests mentioned in the other PR.

wangyangshi:
Thank you for the remind, 2022 build and test successfully.
I am testing 2021 & 2020 now

wangyangshi:
@mjkkirschner Yes, 2020&2021 both build failed

Package GregRevitAuth 1.1.7972.27909 is not compatible with net47.

Do we need to use the other version of GregRevitAuth that support net47?
or upgrade dynamoRevit to use net48?

mjkkirschner:
hmm I think we could update DynamoRevit to .net 4.8 - but it's not clear if these older versions of Revit support .net 4.8 binaries.

For minimal changes for the hotfix we can try recompiling GregRevitAuth to target .net 4.7 - I'm not *sure*  that will work either - but let's try that first...? (maybe multi targeting?)

I will look into it tomorrow if you don't get to it first.

PR #2894, opened 2022-12-16
Upgrade Greg, GregRevitOAuth, and RestSharp version in 2022.1.4

wangyangshi:
this PR want to fix this issue:
https://jira.autodesk.com/browse/REVIT-196808
related discussions are here:
https://github.com/DynamoDS/DynamoRevit/pull/2889


mjkkirschner:
@wangyangshi it looks good but I have a concern about all branches - these updates use the same DynamoRevit version as before - since these are for hotfixes do you intend to update the version number of DynamoRevit binaries. IMO that is safest so that patch installers actually replace the binaries - there are changes in this PR besides just delivering different binaries - the targets that the code are built against also changed to a different framework.

I don't know if it's important, but it seems like best practice that
1. the branch name would maybe be RC12.1.2_Revit_202.1.4 - I guess the current one is okay if it really refers to the DynamoCore version.
2. That the DynamoRevit binary version is updated.

mjkkirschner (approved):

wangyangshi:
@mjkkirschner 
Yes, these PRs are just for Greg change, next I will cherry-pick the unify Newtonsoft.json to them, and then bump d4r versions in these 3 branches.

wangyangshi:
Yes, they really refer to the DynamoCore version, below are the build scripts for them, you can see the dynamo core version in it.
https://git.autodesk.com/Dynamo/DynamoRevitUtils/blob/RC2.12.1_Revit2022.1.4/BuildSolutions.bat
https://git.autodesk.com/Dynamo/DynamoRevitUtils/blob/RC2.6.2_Revit2021.1.8/BuildSolutions.bat
https://git.autodesk.com/Dynamo/DynamoRevitUtils/blob/RC2.3.2_Revit2020.2.10/BuildSolutions.bat

PR #2913, opened 2023-04-14
Fix DynamoPlayer when splash screen is enabled.

reddyashish:
### Purpose

Shows splash screen in Revit 2024 and fixes Dynamo player.
Task: https://jira.autodesk.com/browse/DYN-5489

When Dynamo is started from Revit, we will show the splash screen, which will load the dynamo window. When Dynamo player is started, we will initialize RevitDynamoModel in UIless mode. This is done by checking the commandData for JournalKeys.ShowUiKey and if it is false, we will start in UIless mode.

dynamo player fix

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] Snapshot of UI changes, if any.

### Reviewers
@QilongTang @mjkkirschner @wangyangshi

PR #2914, opened 2023-04-21
Option to launch Dynamo directly without splash screen.

reddyashish:
### Purpose

Option to launch Dynamo directly without splash screen. This is a follow-up of https://github.com/DynamoDS/DynamoRevit/pull/2913 after enabling splash screen in Revit.
@BogdanZavu reported that player samples were not being edited with the latest changes.

When Dynamo player samples are edited, main dynamo is needed and it should start a example workspace. So we need a option to launch main Dynamo directly in that case.

### Declarations

### Reviewers
@QilongTang @Mikhinja

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -305,6 +305,34 @@ namespace Dynamo.Applications
             //subscribe to the assembly load
             AppDomain.CurrentDomain.AssemblyLoad += AssemblyLoad;
 
+            // Launch main Dynamo directly when ShowUiKey is true.
+            if (CheckJournalForKey(commandData, JournalKeys.ShowUiKey, false))
+            {
+                try
+                {
+                    extCommandData = commandData;
+                    LoadDynamoWithoutSplashScreen();
+                }
+                catch (Exception ex)
+                {
+                    // Notify instrumentation
+                    Analytics.TrackException(ex, true);
+                    MessageBox.Show(ex.ToString());
+
+                    DynamoRevitApp.DynamoButtonEnabled = true;
+
+                    // If for some reason Dynamo has crashed while startup make sure the Dynamo Model is properly shutdown.
+                    if (RevitDynamoModel != null)
+                    {
+                        RevitDynamoModel.ShutDown(false);
+                        RevitDynamoModel = null;
+                    }
+
+                    return Result.Failed;
+                }
+                return Result.Succeeded;
+            }
+
             // Show splash screen when dynamo is started, otherwise run UIless mode when needed.
             if (CheckJournalForKey(commandData, JournalKeys.ShowUiKey, true))
             {
@@ -312,8 +340,9 @@ namespace Dynamo.Applications
                 {
                     LoadDynamoView();
                 }));
-                //Register DynamoExternalEventHandler so Revit will call our callback
-                //we requested with API access.
+
+                /* Register DynamoExternalEventHandler so Revit will call our callback
+                we requested with API access. */
                 SplashScreenExternalEvent = ExternalEvent.Create(ssEventHandler);
 
                 try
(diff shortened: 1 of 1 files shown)


QilongTang:
@mjkkirschner There is already a setting in Dynamo for this. FYI: @reddyashish

reddyashish:
That setting is for static splash screen. Dynamic splash screen is always shown. 
@mjkkirschner you mean an option to load the Dynamic main window and not the dynamic splash screen from Core? 
Yes that would be ideal. I thought, since splash screen is a webview2 component and not inside core, DynamoRevit was linked to splash screen directly.

QilongTang:
> hey @reddyashish @QilongTang -nice that this is possible - but do you think it would be better if this was just an option in DynamoCore instead so that integrators would not have to do anything special to accomplish this? (maybe an improvement for the future)

I feel like I want your elaboration on this

mjkkirschner:
yeah, I don't have a solution in mind, it just feels to me that the splash screen is too complex, and using it has become quite complex for integrators, and we should redesign the API for using it so all they need to think about is - I want to start dynamo with a splash screen or without.

QilongTang:
@mjkkirschner That make sense. @reddyashish It looks like to me, that a setting is needed to totally skip even the static splash. But it might be too late for 2.18 release. Also can you attach a gif about the behavior before and after?

mjkkirschner:
yeah agree regarding 2.18 - this is more a comment in general about splash screen / API consideration in general for the future.

reddyashish:
Before a lot of splash screens were starting and it would crash. 

splash D4R

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -305,15 +305,44 @@ public Result ExecuteCommand(DynamoRevitCommandData commandData)
             //subscribe to the assembly load
             AppDomain.CurrentDomain.AssemblyLoad += AssemblyLoad;
 
+            // Launch main Dynamo directly when ShowUiKey is true.
```
QilongTang, line 308: `when ShowUiKey is false`?

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -305,15 +305,44 @@ public Result ExecuteCommand(DynamoRevitCommandData commandData)
...
+            if (CheckJournalForKey(commandData, JournalKeys.ShowUiKey, false))
+            {
+                try
+                {
+                    extCommandData = commandData;
+                    LoadDynamoWithoutSplashScreen();
+                }
+                catch (Exception ex)
```
QilongTang, line 316: It seems the catch part is common between both condition, should we put the whole logic into try-catch so we only have one single try catch?

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -449,6 +478,50 @@ private void LoadDynamoView()
             splashScreen.DynamicSplashScreenReady -= LoadDynamoView;
         }
 
+        // Start main Dynamo directly without splash screen.
+        private void LoadDynamoWithoutSplashScreen()
```
QilongTang, line 482: I guess this function simply reflects logic before we enable the splash screen in D4R right?

reddyashish: yes

src/DynamoRevit/DynamoRevit.cs
reddyashish: when ShowUiKey is false, it will go to else block in line 392(new file). We are changing the default parameter in CheckJournalForKey() to check if that flag is true or false or not present.

src/DynamoRevit/DynamoRevit.cs
reddyashish: Yes changed it, now we have a single try-catch block.

Mikhinja (approved):

reddyashish:
@Mikhinja Can you merge this if it looks good to you. I don't have write access.

PR #2918, opened 2023-04-27
New node Element.GetGeometry

dnenov:
### Purpose

Part of the new nodes PRs, this small PR adds a new Element.GetGeometry node, which incorporates an input for Detail Level.

getgeometry

- added new node Element.GetGeometry that has detail level
- 
### Declarations

### Reviewers

@reddyashish 
@Amoursol


src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -585,6 +586,56 @@ public object[] Geometry()
             return converted.ToArray();
         }
 
+        /// <summary>
+        /// Returns all geometry associated with an element. Ignores transforms when used with linked elements.
+        /// <param name="detailLevel">Detail level</param>
+        /// <returns>List of geometry from the element</returns>
+        /// </summary>
+        public object[] GetGeometry(string detailLevel = "Medium")
```
mjkkirschner, line 594: doesn't an enum make more sense here? What about localized envs? How will users know what strings to use etc? I think there are other examples in the repo - maybe?

mjkkirschner:
wheres the tests ;)?

src/Libraries/RevitNodes/Elements/Element.cs
dnenov: It actually works with the current `Detail Level` dropdown node in Dynamo which yields a string output. It looks like this:

image

The output of the `Detail Level` node is not language dependent, so it made sense to do it this way. Let me know if that still bothers you though!

dnenov:
Recreated to master branch https://github.com/DynamoDS/DynamoRevit/pull/2924, closing this PR.

PR #2910, opened 2023-03-30
DynamoRevit LinkElementNodes

dariaivanciucova:
### Purpose

This PR contains new nodes for Dynamo Revit - all falling under LinkElement category.
image
image

- LinkElement.GetGeometry
- LinkElement.GetLocation
- LinkElement.BoundingBox
- LinkElement.LinkInverseTransform
- LinkElement.LinkTransform
- LinkElement.ByRayBounce

### Declarations

Check these if you believe they are true

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] Snapshot of UI changes, if any.

### Reviewers

@Amoursol
@hwahlstrom

### FYIs

@dnenov

PR #2922, opened 2023-05-25
Re dynamorevit linkinstancenodes

dariaivanciucova:
### Purpose

Re-do of https://github.com/DynamoDS/DynamoRevit/pull/2909 PR.

This PR contains several new nodes for Dynamo Revit. The nodes fall under **LinkInstance** and **Document** category.

- Document.GetLinkInstances
- LinkInstance.AllElementsOfCategory
- LinkInstance.AllElementsOfCategoryInView
- LinkInstance.AllElementsOfClass
- LinkInstance.AllElementsAtLevel
- LinkInstance.ByName
- LinkInstance.ByTitle
- LinkInstance.Document
- LinkInstance.ElementById
- LinkInstance.UniqueId

image
image

image

### Declarations

### Reviewers

@Amoursol 
@hwahlstrom

### FYIs

@dnenov


dariaivanciucova:
> Note! Currently targeting Revit 2024 dlls, to be able to build. There are changes inside multiple project files where the references have been updated. Potentially these need to be removed from the PR before merging, but on a few occasions, there are valid changes to the project files that need to be kept.

This should now be resolved

src/DynamoRevitIcons/DynamoRevitIcons.csproj
```diff
@@ -34,7 +34,9 @@
   </PropertyGroup>
   <ItemGroup>
     <EmbeddedResource Include="DSRevitNodesUIImages.resx" />
-    <EmbeddedResource Include="RevitNodesImages.resx" />
+    <EmbeddedResource Include="RevitNodesImages.resx">
```
reddyashish, line 37: Oh so this was originally in the other PR https://github.com/DynamoDS/DynamoRevit/pull/2910/files#diff-42a24daedb6c64bb3a1c76755ee376806e5502d21e4a904a55360fe74f876a91R37 here. Is that the only change?

src/Libraries/RevitNodes/RevitNodes.csproj
```diff
@@ -280,4 +281,4 @@
   <Target Name="AfterBuild">
   </Target>
   -->
-</Project>
```
reddyashish, line 283: Can we add EOL here to be consistent with other files.

reddyashish:
Thanks @dariaivanciucova Looks good with couple of comments.

dariaivanciucova:
@reddyashish tested - all works as expected

reddyashish:
@Mikhinja All good for this PR. It can be merged.

src/Libraries/RevitNodes/Elements/LinkInstance.cs
```diff
@@ -0,0 +1,425 @@
...
+        }
+
+
+        /// <summary> Retrieves all elements of a given category in a link instance </summary>
+        /// <param name="linkInstance">Revit link Instance</param>
+        /// <param name="category">Element category</param>
+        /// <returns name="linkedElements[]">All elements of the category</returns>
+        public static List<Element> AllElementsOfCategory(Element linkInstance, Revit.Elements.Category category)
+        {
+            RevitLinkInstance revitlinkInstance = linkInstance.InternalElement as RevitLinkInstance;
```
Mikhinja, line 52: If the first input, `linkInstance`, will not be a RevitLinkInstance then this will be `null` and the error message might be misleading.

This is a visible node in Dynamo, so the user may create a Dynamo graph that could be incorrect. In that case the user will inspect the input and see that it is not `null`, but a regular `Element` or something else, yet the error they will receive will complain about `null`.

Not a big deal, since the tooltip should be informative enough, I only softly recommend adding a separate error message for that case.

src/Libraries/RevitNodes/Elements/LinkInstance.cs
```diff
@@ -0,0 +1,425 @@
...
+            BuiltInCategory bic = (BuiltInCategory)System.Enum.Parse(typeof(BuiltInCategory),
```
Mikhinja, line 53: Why not `BuiltInCategory bic = (BuiltInCategory)category.InternalCategory.Id.Value` ?

dnenov:
Hi @Mikhinja - would it be OK if we were to merge this PR, since it's quite old now and we are having major issues when building against the old RevitAPI.dll and RevitAPIUI.dll and the special version of dynamocore.dll .. basically it's quite the nightmare to make a clean PR, but keep a stash full of references to allow us to work on it. I was hoping that after merging this PR, we can have a clean and crisp build against the latest `Master` that doesn't have any of these issues. We can then take work on all the comments, if that's OK.
image

Mikhinja (approved):

PR #2933, opened 2023-07-26
Pass the Preferences as part of the Initializacion configuration

jesusalvino:
### Purpose

As part of the task https://jira.autodesk.com/browse/DYN-5816 the DynamoModel that the class RevitDynamoModel inherits, needs to pass the explicit Preferences as part of its initialization configuration.

### Declarations

### Reviewers
@QilongTang 
@reddyashish

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -607,7 +607,8 @@ namespace Dynamo.Applications
                 AuthProvider = new RevitOAuth2Provider(new DispatcherSynchronizationContext(Dispatcher.CurrentDispatcher)),
                 ExternalCommandData = commandData,
                 UpdateManager = revitUpdateManager,
-                ProcessMode = isAutomationMode ? TaskProcessMode.Synchronous : TaskProcessMode.Asynchronous
+                ProcessMode = isAutomationMode ? TaskProcessMode.Synchronous : TaskProcessMode.Asynchronous,
+                Preferences = PreferenceSettings.Instance
             });
         }
 
--- a/test/Libraries/RevitTestServices/RevitSystemTestBase.cs
+++ b/test/Libraries/RevitTestServices/RevitSystemTestBase.cs
@@ -276,7 +276,8 @@ namespace RevitTestServices
                         SchedulerThread = new TestSchedulerThread(),
                         PackageManagerAddress = "https://www.dynamopackages.com",
                         ExternalCommandData = commandData,
-                        ProcessMode = RevitTaskProcessMode
+                        ProcessMode = RevitTaskProcessMode,
+                        Preferences = PreferenceSettings.Instance
                     });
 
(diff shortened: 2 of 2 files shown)


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -607,7 +607,8 @@ private static RevitDynamoModel InitializeCoreModel(DynamoRevitCommandData comma
                 AuthProvider = new RevitOAuth2Provider(new DispatcherSynchronizationContext(Dispatcher.CurrentDispatcher)),
                 ExternalCommandData = commandData,
                 UpdateManager = revitUpdateManager,
-                ProcessMode = isAutomationMode ? TaskProcessMode.Synchronous : TaskProcessMode.Asynchronous
+                ProcessMode = isAutomationMode ? TaskProcessMode.Synchronous : TaskProcessMode.Asynchronous,
+                Preferences = PreferenceSettings.Instance
```
QilongTang, line 611: Do we need to pass the PathResolver?

jesusalvino: > Do we need to pass the PathResolver?

yes as the previously been doing in the startconfiguration, now We assigning it to the PathManager in the DynamoModel ctor, the only update is need to pass the Preferences.

QilongTang (approved):

QilongTang:
@Mikhinja PTAL, please refer to https://github.com/DynamoDS/Dynamo/pull/14118

mjkkirschner:
Does this cause the multiple instance bug shown on slack?

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
```diff
@@ -276,7 +276,8 @@ protected override void StartDynamo(TestSessionConfiguration testConfig)
                         SchedulerThread = new TestSchedulerThread(),
                         PackageManagerAddress = "https://www.dynamopackages.com",
                         ExternalCommandData = commandData,
-                        ProcessMode = RevitTaskProcessMode
+                        ProcessMode = RevitTaskProcessMode,
+                        Preferences = PreferenceSettings.Instance
```
mjkkirschner, line 280: have you tried running all RTF tests locally?

jesusalvino:
> Does this cause the multiple instance bug shown on slack?

I Don't think so since I have tested a branch before this change, but looks like it's an issue that only pass for my local setup

test/Libraries/RevitTestServices/RevitSystemTestBase.cs
jesusalvino: > have you tried running all RTF tests locally?

I have tried but even when I put the paths and the proper tests it got me errors, I have tested with some Dynamo flows under the DynamoRevit pointing to DynamoCore 4.8

Mikhinja (approved):

Mikhinja (approved):

PR #2955, opened 2023-08-24
DYN-6173 - fix bug where multiple instances of dynamo can be opened.

mjkkirschner:
### Purpose

- When the splash screen closes either because the user closed it, or the user launched dynamo, the DynamoButton was enabled in the Revit ribbon.
- This PR uses a new flag added to the splash screen to determine if the button in the ribbon should be enabled or not.
- modifies the DynamoRevit build so that it does not overwrite the Dynamo.config file if it already exists so when developing locally you don't need to keep setting the dynamo core path on each DynamoRevit build.
- switches refs to DynamoCoreNodes instead of WPFUILibrary nuget package.

- [ ] when the nuget is published, remove the dynamic call.

### Declarations

Check these if you believe they are true

- [ ] The code base is in a better state after this PR
- [ ] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] Snapshot of UI changes, if any.

PR #2958, opened 2023-08-25
Follow up PR of #2955

mjkkirschner:
continues: 
https://github.com/DynamoDS/DynamoRevit/pull/2955

removes the dynamic call, updates the nugets to use the new flag api.
Also removes the package ref on ionic zip, we stopped relying on it a long time a ago.


src/Config/packages-template.aget
```diff
@@ -13,8 +13,7 @@
       "Newtonsoft.Json": "8.0.3",
       "NUnit": "2.6.3",
       "Prism": "4.1.0",
-      "RestSharp": "106.12.0",
-      "Unofficial.Ionic.Zip": "1.9.1.8"
```
QilongTang, line 17: @Mikhinja Please confirm it is fine to remove such dependency as part of DynamoRevit 2.19 release

QilongTang (approved):

src/Config/packages-template.aget
mjkkirschner: @QilongTang I checked the DynamoRevit addin and do not see this binary there.

mjkkirschner: Also note that this PR is to the master branch - which I assume will not be subject to such reasons for 2 reasons.
1. potential 3.0
2. GL25 is supposed to resolve patch installer insanity.

Mikhinja: Doesn't look like it's actually used in DynamoRevit nodes, not sure why it's there.

Mikhinja (approved):

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -1100,9 +1100,7 @@ private static void OnSplashScreenClosed(object sender, EventArgs e)
             view.Closed -= OnSplashScreenClosed;
             //if the user explicitly closed the splashscreen, then we should let them boot
             //dynamo back up.
-            //TODO use new api by updating dynamocorewpf nuget.
-            //After it's published, remove the use of dynamic.
-            if(sender is Dynamo.UI.Views.SplashScreen ss && ((dynamic)ss).CloseWasExplicit)
+            if(sender is Dynamo.UI.Views.SplashScreen ss && ss.CloseWasExplicit)
```
Mikhinja, line 1103: Do we need these changes in Revit 2024.2 as well?

src/Config/packages-template.aget
QilongTang: hmm @mjkkirschner I thought this change is for R2024.2?

src/DynamoRevit/DynamoRevit.cs
mjkkirschner: yes, or users will be able to open multiple instances of Dynamo.

PR #2989, opened 2023-10-23
update rest sharp to 109.0.1

mjkkirschner:
as RestSharp 106.12 no longer receives security updates, update to 109.0.1

I could not update to 110.2 as originally planned as I ran into issues with Dynamo's current .net 6 build and the System.Text.Json package.
for now I have an issue out to the dotnet runtime maintainers to see if this is a bug or as designed:
https://github.com/dotnet/runtime/issues/93780

anyhow, 109.0.1 should get security updates and it does not depend on a specific version of STJ so we load whatever the runtime provides.

Testing simple cases in DynamoRevit works.

This PR also just fixes a few compiler warnings in some new nodes.


src/Config/packages.aget
```diff
@@ -9,13 +9,13 @@
...
       "DynamoVisualProgramming.ZeroTouchLibrary": "3.0.0-beta6354",
       "DynamoVisualProgramming.DynamoSamples": "3.0.0-beta6354",
-      "Greg": "2.5.0.5464",
-      "GregRevitAuth": "1.3.8591.28369",
+      "Greg": "3.0.0.2955",
+      "GregRevitAuth": "2.0.8696.26508",
       "pythonnet_py38_win": "2.5.1",
       "Newtonsoft.Json": "13.0.1",
       "NUnit": "3.13.3",
       "Prism": "4.1.0",
```
mjkkirschner, line 17: @pinzart90 can we remove this?

mjkkirschner: hmm, let me see what happens if I just remove this.

mjkkirschner: ok, I will try to deal with it here:
https://github.com/DynamoDS/DynamoRevit/pull/2990

Mikhinja (approved):

PR #2990, opened 2023-10-23
remove direct dep on prism

mjkkirschner:
instead of using the prism DelegateCommand type, use the DynamoUI type which implements the same interface.


pinzart90 (approved):

sm6srw (approved):

mjkkirschner:
Screenshot 2023-10-23 at 4 49 08 PM

it does seem to work.

PR #2997, opened 2023-11-02
Mikhinja/revit 212138/fix wrongly updated node search tests

Mikhinja:
### Purpose

Fix a few tests for node search that I wrongly updated when upgrading to Dynamo Core 3.0 and NUnit 3.0.

### Declarations

### Reviewers

@mjkkirschner

### DIFF
--- a/test/Libraries/RevitIntegrationTests/RegressionTests.cs
+++ b/test/Libraries/RevitIntegrationTests/RegressionTests.cs
@@ -100,7 +100,7 @@ namespace RevitSystemTests
             Assert.NotNull(homespace, "The current workspace is not a HomeWorkspaceModel");
 
             // Iterate through all loaded nodes in library & add Revit UI nodes to ws
-            var nodeList = Model.CurrentWorkspace.Nodes.OfType<NodeSearchElement>();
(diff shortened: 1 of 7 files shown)


test/Libraries/RevitIntegrationTests/RegressionTests.cs
```diff
@@ -159,7 +159,7 @@ public void NoDuplicatePortsOnUINodes()
             Assert.NotNull(homespace, "The current workspace is not a HomeWorkspaceModel");
 
             // Iterate through all loaded nodes in library & add Revit UI nodes to ws
-            var nodeList = Model.CurrentWorkspace.Nodes.OfType<NodeSearchElement>();
+            var nodeList = Model.SearchModel.Search(string.Empty, Model.LuceneUtility);
```
mjkkirschner, line 162: this returns all nodes in search?

mjkkirschner (approved):

test/Libraries/RevitIntegrationTests/RegressionTests.cs
Mikhinja: It did not, I fixed it, but I still have other issues with these tests. I'll commit the current state, but it's not yet ready.

PR #3010, opened 2023-11-20
Drop SOAP formatter and move to JSON based serialization

mjkkirschner:
### Purpose

This work is mostly done by @saintentropy - and uses the new JSON serialization based APIs for trace. This means strings are the new API type and serialization/deserialization to whatever format should be handled by the integrators as in this PR.

I have not been able to run these tests locally as RTF is out of date 😢 - but I did get this to at least build.

@Mikhinja let's please merge this ASAP and get a test run on so we can iterate on the failures we know will be here, I will help with that once I have a list of failures.

Also note that DynamoCore version is updated here to include the new APIs we need so that will also need to be updated on the Revit side.

### Declarations

### Reviewers

@Mikhinja 

### FYIs

@saintentropy


src/Libraries/RevitNodes/Elements/DirectShape.cs
```diff
@@ -40,13 +40,6 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
             this.syncId = syncId;
             this.materialId = materialId.Value;
         }
-
-        public DirectShapeState(SerializationInfo info, StreamingContext context) :
```
Mikhinja, line 44: Won't this cause a breakage without obsolete message? Not that we need to keep this, but perhaps document somewhere for package developers?

Mikhinja (approved):

PR #3013, opened 2023-11-29
AGD-3139 Migrate requested selection nodes form GD to Revit

saintentropy:
### Purpose

The purpose of this PR is to migrate some of the nodes delivered with the GD Package to Revit as OOB nodes.  User feedback has come in that in some scenarios, GD package is not present and these nodes can not be reliably utilized in shared graphs.  The specific nodes included in this PR are:

1. Pick Revit Elements, Similar to the Select Revit Element but utilizes the Multiple Selection UI (ie click multiple times and complete selection with finish button) vs Drag Select which limits precision. 
2. Select Family Instance
3. Select Family Instances
4. Pick Family Instances.

The family instance node serves as to fill a gap between the general Select (any) Model Element nodes which provide no filtering and the Selection By Category which may be two precise in some cases.  The Family Instance selection nodes filter the selection in a more general way then the category filter can support. 
### Declarations

### Reviewers

@BogdanZavu @Mikhinja 

### FYIs

@twastvedt


src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -576,7 +600,7 @@ protected override IEnumerable<Reference> ExtractSelectionResults(Reference sele
      IsVisibleInDynamoLibrary(false)]
     public class DSAnalysisResultSelection : ElementSelection<Element>
     {
-        private const string message = "Select an analysis result.";
+        private static string message = Resources.SelectAnalysisResultsMessage;
```
saintentropy, line 603: Opportunistic localization additions.  We have added the _message_ string in some cases but not all.

src/Libraries/RevitNodesUI/Selection.cs
```diff
@@ -981,6 +1005,36 @@ public DSModelElementsSelection(IEnumerable<string> selectionIdentifier, IEnumer
         { }
     }
 
+    [NodeName("Pick Model Elements")]
```
BogdanZavu, line 1008: Or perhaps "Select Model Elements with the Pick Editor"

BogdanZavu:
Since we only have api tests which don't perform a real selection anyway and mostly our addition is just launching the pick editor on top of the base classes I think we can say we are covered. We can add something later if needed.

BogdanZavu (approved):

Mikhinja (approved):

src/Libraries/RevitNodesUI/Selection.cs
saintentropy: Yes we will keep the shorter for now

src/Libraries/RevitNodesUI/RevitSelectionHelper.cs
```diff
@@ -218,6 +218,66 @@ private static IEnumerable<T> RequestElementSelection(string selectionMessage, I
...
+                elements.Add((T)e);
+            }
+
+            return elements;
+        }
+
+        #endregion
+    }
+
     internal class ElementSelectionFilter<T> : ISelectionFilter
```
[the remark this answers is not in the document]
saintentropy: Existing nodes work unchanged.  The 4 added nodes are augmentation to the existing.

PR #3029, opened 2024-01-18
Skip handling Dispatcher unhandled exception if already handled in DynamoCore

aparajit-pratap:
### Purpose

This PR depends on https://github.com/DynamoDS/Dynamo/pull/14840 first being merged into Dynamo 3.0.x. If not for 3.0.x, it can wait until the Dynamo PR is merged into 3.1.

It skips certain steps in the Dispatcher's unhandled exception handler if the corresponding handler is already invoked in DynamoViewModel. This new handler in DVM has been added in https://github.com/DynamoDS/Dynamo/pull/14840.

### Declarations

### Reviewers

### FYIs


src/DynamoRevit/DynamoRevit.cs
```diff
@@ -1052,11 +1052,12 @@ public static string GetRevitContext(DynamoRevitCommandData commandData)
...
-                RevitDynamoModel.OnRequestsCrashPrompt(
-                    RevitDynamoModel,
-                    new CrashPromptArgs(args.Exception.Message + "\n\n" + args.Exception.StackTrace));
-                RevitDynamoViewModel.Exit(false); // don't allow cancellation
+                if (!args.Handled)
+                {
+                    DynamoModel.IsCrashing = true;
+                    RevitDynamoModel.OnRequestsCrashPrompt(new CrashErrorReportArgs(args.Exception));
+                    RevitDynamoViewModel.Exit(false); // don't allow cancellation
+                }
```
pinzart90, line 1060: Maybe we should still mark the arg.handled as true in the finally block? The docs still say it prevents some internal work thing to be triggered

aparajit-pratap: Yes, it's still being marked true inside the finally block.

sm6srw: @aparajit-pratap Thanks, so this PR will solve the remaining part of DYN-6066 for now?

aparajit-pratap: @sm6srw yes, it fulfills the remaining part of DYN-6066 but I'm still testing something more for my task DYN-6492 as mentioned during standup.

sm6srw (approved):

src/DynamoRevit/DynamoRevit.cs
```diff
@@ -1019,54 +1018,6 @@ public static string GetRevitContext(DynamoRevitCommandData commandData)
...
-                RevitDynamoModel.OnRequestsCrashPrompt(
-                    RevitDynamoModel,
-                    new CrashPromptArgs(args.Exception.Message + "\n\n" + args.Exception.StackTrace));
-                RevitDynamoViewModel.Exit(false); // don't allow cancellation
-            }
-            catch { }
-            finally
-            {
-                args.Handled = true;
-                DynamoRevitApp.DynamoButtonEnabled = true;
```
pinzart90, line 1065: what about the button ?

aparajit-pratap: The button is reenabled even without this.

pinzart90: Is it somewhere in the view.CLose events ?
THere would be a chance that if an exception is thrown in the exception handler, then the button close will not be called at all

aparajit-pratap: Yes, it is re-enabled in `OnDynamoViewClosed` [here](https://github.com/DynamoDS/DynamoRevit/blob/fb7c649f0a130869b596bf4ab88dc1cc17c52eb1/src/DynamoRevit/DynamoRevit.cs#L1085).

Mikhinja: But why the API breaking change so late, in a minor build (3.0.2 vs 3.0.0), and why after Revit API freeze?

This would require more testing on Revit side, that various CER scenarios still work and we are not missing crash reports, possibly even more code changed.

Is this API change in 3.0.2 mandatory or can we revert it, or make other builds based on 3.0.0 with only the critical fixes?

Mikhinja: Yes, that change is what I was actually referring to. I merged it together with Dynamo 3.0.2.

pinzart90 (approved):

PR #3058, opened 2024-03-27
DYN-6775 Localizing SplashScreen

RobertGlobant20:
### Purpose

The buttons from the Static SplashScreen are being localized in DynamoSandbox but in Revit 2026 were not being localized due that the function OnDetectLanguage() was not being called. Then my fix was just calling this method so the SetLabels methods will be also called and the buttons will be in the expected language.

### Declarations

### Reviewers

@QilongTang

### DIFF
--- a/src/DynamoRevit/DynamoRevit.cs
+++ b/src/DynamoRevit/DynamoRevit.cs
@@ -394,6 +394,7 @@ namespace Dynamo.Applications
         {
             // create core data models
             RevitDynamoModel = InitializeCoreModel(extCommandData);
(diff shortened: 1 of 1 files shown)


RobertGlobant20:
Screenshot showing the SplashScreen buttons in Spanish
image

QilongTang:
FYI: @Mikhinja

RobertGlobant20:
@QilongTang Just to let you know that this PR needs to be merged so the SplashScreen buttons are being localized

QilongTang (approved):

PR #3065, opened 2024-04-23
Improve FilterRule.ByRuleType to create filter rule when empty string or multiline text parameter is provided

AdrianaStanica:
### Purpose
After testing the initial improvement regarding FilterRule.ByRuleType (changing the data type of the value according to the data type of the parameter value in C# so that we avoid invalid data type casting in Revit), we found out that we unintentionally throw an exception when comparing a string value with an empty project parameter of type string or multiline text (which in theory should work to create the filter rule). But empty string parameters have no default value, so extracting the value data type led to an exception. We found a workaround for this, and now we use the parameter's storage type to get its data type in this situation.

Jira issue: REVIT-219696
### Declarations

### Reviewers

@Mikhinja


Mikhinja (approved):

mjkkirschner:
test?

AdrianaStanica:
We will consider adding a regression test separately @mjkkirschner

PR #3066, opened 2024-04-30
Fix regression in FamilyType.ByGeometry Node

saintentropy:
### Purpose

This PR is to fix a regression in the FamilyType.ByGeometry node.  Specifically a change in the Revit API behavior with `Delete(elementID)` method is causing the node to throw an exception.  This is regression between Revit 2024 and Revit 2025 API behavior.  In this case the D4R code temporarily imports an SAT file to get geometry to use in the FamilyType creation.  The code deletes the SAT import once the geometry has been converted to a Revit Solid.  The Revit API change in behavior impacts the ability to `Delete` the SAT import.  The workaround is to unpin the imported Element so that the Delete will not throw.

While there are many use of the `Delete` element method in the D4R code, it appears the behavior change is limited to the SAT import.

### Declarations

### Reviewers

@mjkkirschner @sm6srw 

### FYIs

@Mikhinja


sm6srw:
LGTM.

mjkkirschner (approved):

PR #3085, opened 2024-08-09
Get/Set Parameter Improvements

dimven-adsk:
### List of Affected Nodes/Modules

Element.GetParameterByName
Element.SetParameterByName

### Current Performance
There is a two-fold penalty with the current method. The first one is because every single parameter is fetched for every element and the second is because the entire definition is fetched for each of the parameters in the resulting sequence when all we need is the parameter's name

devenv_qRNt6WqQKC

### Proposed Performance
Both of these can be avoided by using the GetParameters(string) method instead. That way fetching the definition is completely avoided and the parameter filtering happens in the native side.

devenv_oxm0oEVlf8

### Dynamo Tuneup Comparison
On a model with ~55k elements fetching a single parameter value we drop form 14000ms to 6000ms

image

image

### Checklist

- [x] There are no public function signature changes
- [x] The code changes have been documented
- [x] The overall behavior does not change


mjkkirschner:
@dimven-adsk please don't commit binaries, especially those that are 32 megs, it's going to cause issues with the repo even if you don't merge them.

src/DynamoRevit/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "profiles": {
```
mjkkirschner, line 2: I probably would not commit this unless you think it's valuable for everyone?

src/Libraries/RevitNodes/Elements/Element.cs
```diff
@@ -436,10 +436,7 @@ private Autodesk.Revit.DB.Parameter GetParameterByName(string parameterName)
             // 3. If it exist: Use the first writable parameter
             // 4. Otherwise: Use the first read-only parameter
             //
-            var allParams =
-            InternalElement.Parameters.Cast<Autodesk.Revit.DB.Parameter>()
-                .Where(x => string.CompareOrdinal(x.Definition.Name, parameterName) == 0)
-                .OrderBy(x => x.Id.Value);
+            var allParams = InternalElement.GetParameters(parameterName).OrderBy(x => x.Id.Value);
```
mjkkirschner, line 439: cool, so if I understand correctly, you skip the dynamo function and use the RevitAPI - this is usually a good pattern for sure, is there anything weird with element binding this could cause?

mjkkirschner:
Do the tests still pass locally?

src/DynamoRevit/Properties/launchSettings.json
twastvedt: Seems generally useful to me...

twastvedt:
> @dimven-adsk please don't commit binaries unnecessarily, especially those that are 32 megs, it's going to cause issues with the repo even if you don't merge them.

Can we add a packagereference with a hintPath to Program Files? Maybe that's out of scope of this PR...

lib/Revit 2025/net8/RevitAPI.dll
Mikhinja, line 1: It shouldn't be necessary to commit these binaries. The internal build script handles downloading these in this folder structure. There already are references to these binaries in projects, I recommend against committing specific binary files.

dimven-adsk: Will do. I was following the readme suggestions :)

src/Libraries/RevitNodes/Elements/Element.cs
dimven-adsk: No, I don't believe so.

dimven-adsk:
I am sorry, my intention was to post this PR on the dedicated fork as to not interfere with the current performance cycle. Please let's continue the discussion here:

https://github.com/dimven-adsk/DynamoRevit/pull/1

PR #3102, opened 2024-10-10
DYN-7400 : remove python runtime dependency

BogdanZavu:
### Purpose
We want to remove Python.Runtime dependency at D4R level in order to easily handle multiple python engines.
The dictionary marshaler has been moved at DynamoCore level - will add reference to the new DynamoServices package that contains the coresponding API once [this pr](https://github.com/DynamoDS/Dynamo/pull/15537) is merged and a new package is published.

### Declarations

### Reviewers
@Mikhinja @twastvedt


twastvedt (approved):
Makes sense to me, but I'm no marshaling expert.

Mikhinja (approved):

src/DynamoRevit/Models/RevitDynamoModel.cs
```diff
@@ -495,56 +505,16 @@ private void OnPythonEngineCollectionChanged(object sender, NotifyCollectionChan
...
-                        });
-            }
-            Func<object, object> unwrap = UnwrapElementMarshaler.Marshal;
             // Turn off element binding during python script execution
             ElementBinder.IsEnabled = false;
-            // register UnwrapElement method
-            scopeSet("UnwrapElement", unwrap);
+            if (UnwrapElementMarshaler != null)
+            {
+                Func<object, object> unwrap = UnwrapElementMarshaler.Marshal;
```
Mikhinja, line 514: Do nested `dict` and `list` objects work appropriately?

Do we know if there were existing workarounds related to python wrapping done by 3rd party packages (nodes) or included in python scripts?

BogdanZavu: I think yes. I've been looking into some graphs from the tests in D4R that make use of that code and it seems to work as expected.

Mikhinja:
If this requires Dynamo 3.4 then I recommend waiting for Dynamo 3.4 to be integrated first before merging

src/Config/packages.aget
```diff
@@ -2,13 +2,13 @@
...
     "references": {
       "DynamicLanguageRuntime": "1.2.2",
-      "DynamoVisualProgramming.Core": "3.3.0.6316",
-      "DynamoVisualProgramming.DynamoCoreNodes": "3.3.0.6316",
-      "DynamoVisualProgramming.DynamoServices": "3.3.0.6316",
-      "DynamoVisualProgramming.Tests": "3.3.0.6316",
-      "DynamoVisualProgramming.WpfUILibrary": "3.3.0.6316",
-      "DynamoVisualProgramming.ZeroTouchLibrary": "3.3.0.6316",
-      "DynamoVisualProgramming.DynamoSamples": "3.3.0.6316",
+      "DynamoVisualProgramming.Core": "3.4.0-beta6616",
```
Mikhinja, line 5: In order to merge in Revit we need to use a build that has retention "keep" in artifactory. This build currently has retention policy 30 ( days ).

BogdanZavu: Good catch! I changed the policy to keep.

PR #3152, opened 2025-04-04
d4r for design automation

pinzart90:
Minimal changes to get D4R working on Design automation
This PR is based on RC3.3.0_Revit2025 because the latest Revit on DA seems to be 2025

I opted to use "if def"s in order to minimize the changes and avoid breaking changes. As a follow up task we should do a proper separation of Core and UI functionality in all the projects (this will most likely mean breaking changes).
I will create a couple of jira tasks with proposed changes to pursue this further.


src/DADynamoConsole/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿using System.Reflection;
+
+Assembly.LoadFrom(@"C:\Program Files\Autodesk\Revit 2025\RevitAPI.dll");
```
mjkkirschner, line 3: are these console applications / entry points used? How does DA call them? Won't these not run the in context of Revit but as separate processes?

src/Libraries/RevitNodes/RevitNodeModel.cs
```diff
@@ -11,4 +11,4 @@ public RevitNodeModel() { }
         [JsonConstructor]
         public RevitNodeModel(IEnumerable<PortModel> inPorts, IEnumerable<PortModel> outPorts) : base(inPorts, outPorts) { }
     }
```
pinzart90, line 13: @mjkkirschner do you think this a proper place to have a NodeModel base class ?

pinzart90: It is used in RevitNodesUI (valid for cloud context too)

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+    public class DAEntrypoint : IExternalDBApplication
+    {
+        private DynamoModel model;
+        internal static DynamoPlayerLoggerConfiguration logConfig = new DynamoPlayerLoggerConfiguration() { DynamoLogLevel = Dynamo.Logging.LogLevel.Console, LogLevel = DynamoPlayer.LogLevel.Information };
+
+        private string LoadMessage;
+        private string WorkItemFolder;
+
+        private static EventHandlerProxy proxy;
+        public static EventHandlerProxy EventHandlerProxy
```
mjkkirschner, line 27: /// summary

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            get { return proxy; }
+        }
+
+        public static List<IUpdater> Updaters = new List<IUpdater>();
```
mjkkirschner, line 32: why is this static, public etc?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        private void CheckIfPythonExists(string inputFolder)
+        {
+            var pyDir = "python-3.9.12-embed-amd64";
```
mjkkirschner, line 36: maybe this should be a bit more tolerant of different python versions?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        private static void CurrentDomain_UnhandledException(object sender, UnhandledExceptionEventArgs e)
+        {
+            Console.WriteLine($"Unhandled exception: {e}");
+        }
+
+        public ExternalDBApplicationResult OnStartup(ControlledApplication application)
+        {
+            AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
```
mjkkirschner, line 85: Maybe we should just stop using AppDomain - it's just a pointer to the default ALC - right? Maybe lets be more specific, especially if Revit starts using different ALCs for Dynamo/Addins?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+            AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
+            AppDomain.CurrentDomain.AssemblyResolve += DynamoPathResolver.ResolveAssembly;
+
+            Console.WriteLine("<<!>> Starting to load D4DA");
+
+            try
+            {
+                RevitServices.Transactions.TransactionManager.SetupManager(new RevitServices.Transactions.AutomaticTransactionStrategy());
+                ElementBinder.IsEnabled = true;
```
mjkkirschner, line 94: I am not sure element binding really makes sense in the context of design automation, it really only makes sense in interactive workflows.

I would disable it by default or potentially have a config option to turn it on instead of defaulting to on.

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+
+                RevitServicesUpdater.Initialize(Updaters);
```
mjkkirschner, line 96: IMO updaters seem similar to element binding, the use case I am familiar with for `updaters` is that when updaters raise changes Dynamo will kick off node executions, but this doesn't make much sense in a non-interactive single run session.

Let's chat about this more - we should review how updaters are used in DynamoRevit and what use cases we need to support in Design Automation.

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+
+                DesignAutomationBridge.DesignAutomationReadyEvent += HandleDesignAutomationReadyEvent;
```
mjkkirschner, line 98: I guess unsubscribing this doesn't really matter since DA will shutdown Revit between runs?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+            foreach (var xx in Directory.EnumerateFiles(WorkItemFolder))
+                Console.WriteLine(xx);
+
+            var dynTempDir = Path.Combine(WorkItemFolder, "dyn_tmp");
+            Console.WriteLine($"<<!>> Work folder is '{WorkItemFolder}'");
+            var app = e.DesignAutomationData?.RevitApp;
+            Console.WriteLine("<<!>> Preparing Dynamo model. Vers 1");
+
+            var hostloc = typeof(Autodesk.Revit.ApplicationServices.Application).Assembly.Location;
+            var asmLocation = Path.GetDirectoryName(hostloc);
```
mjkkirschner, line 124: this isn't going to work with ASC (autodesk shared components)
@sm6srw can you confirm?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+            Console.WriteLine($"using asm at location {asmLocation}");
+            Console.WriteLine($"Is Loaded {LoadMessage}");
+            var curFolder = Directory.GetParent(Assembly.GetExecutingAssembly().Location).FullName;
```
mjkkirschner, line 127: I'm confused what directory this is supposed to be pointing to - current in what sense?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+            var cmdArgs = new Dynamo.Applications.StartupUtils.CommandLineArguments();
+            cmdArgs.ASMPath = asmLocation;
+            cmdArgs.DisableAnalytics = true;
+            cmdArgs.ServiceMode = true;
+            cmdArgs.AnalyticsInfo = new() { HostName = "Revit_DA" };
+            cmdArgs.ImportedPaths = [Path.Combine(curFolder, "nodes", "RevitNodes.dll"), Path.Combine(curFolder, "nodes", "DSRevitNodesUI.dll")];
```
mjkkirschner, line 133: okay, you're going to need to give me some more info as to why you think this approach is better than what Dynamo/DynamoRevit does normally to resolve dependencies?

Or is the crux, that you want to avoid subclassing DynamoModel for some reason?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+            // need this for cloud, does not work on local
```
mjkkirschner, line 134: what does not work in local?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+            cmdArgs.UserDataFolder = Path.Combine(dynTempDir, "Dynamo Revit");
+            cmdArgs.CommonDataFolder = Path.Combine(dynTempDir, "Dynamo");
+
+            // Startup a new project, maybe an option we can have ?
+            //app.NewProjectDocument(Autodesk.Revit.DB.UnitSystem.Metric);
```
mjkkirschner, line 139: If I were a user executing against dynamo as a service - I would expect this information to somehow have been supplied via a web UI or via part of the graph authoring experience ahead of time.

For example:

1. I authored the graph in a metric project, so that should be taken into account when run in this context against another project.
2. I set this graph to run against a particular project already so we should just open that project.


Do you have a user use case in mind where a graph should be run on a brand new project?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            DocumentManager.Instance.PrepareForAutomation(app);
+
+            //true, asmLocation, "Revit_DA"
+            model = Dynamo.Applications.StartupUtils.MakeCLIModel(cmdArgs);
+
+            LoadMessage = model != null ? "loaded" : "no loaded";
+            Console.WriteLine($"Checking python in {curFolder}");
+            CheckIfPythonExists(curFolder);
```
mjkkirschner, line 148: can you explain why you are doing this as opposed to letting Dynamo load python the normal way?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+            var testFolder = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var res = dynHandler.HandleRoute("POST", "/v1/graph/run", runContent);
+            var output = res.Result;
+
+            var result = Newtonsoft.Json.JsonConvert.SerializeObject(output);
+            Console.WriteLine(result);
+
+            File.WriteAllText(Path.Combine(WorkItemFolder, "result.json"), result);
+
+            model.RunCompleted += Model_RunCompleted;
```
mjkkirschner, line 173: isn't this too late to attach this event? Hasn't the graph already been run via HandleRoute?

mjkkirschner (commented):
part 1 of my review - mostly questions. Will continue pt2 when I can.

src/DADynamoApp/DADynamoApp.csproj
```diff
@@ -0,0 +1,47 @@
...
+    
+    <ItemGroup>
+        <Reference Include="RevitAPI">
+            <HintPath>$(REVITAPI)\RevitAPI.dll</HintPath>
+            <Private>False</Private>
+        </Reference>
+    </ItemGroup>
+    <ItemGroup>
+        <PackageReference Include="Autodesk.Forge.DesignAutomation.Revit" Version="2025.0.1" />
+        <PackageReference Include="DynamoVisualProgramming.Core" Version="3.3.0.6316" ExcludeAssets="runtime" GeneratePathProperty="true" />
```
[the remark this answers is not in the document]
mjkkirschner: ah, I see your description now.

src/Config/CS_SDK.props
```diff
@@ -1,11 +1,10 @@
 <Project>
 	<Import Project="$(SolutionDir)Config/user_local.props" />
 	<PropertyGroup>
-		<Platforms>NET70;NET80</Platforms>
```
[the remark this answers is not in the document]
pinzart90: CICD can invoke a build either with platform values set, or with other underlying properties (UI vs Core)

src/DADynamoApp/DAEntrypoint.cs
sm6srw: This will not work on desktop Revit. But I am not sure if they use ASC in design automation. We don't do the ASC lookup in D4R normally. Revit tells us where it is. Does that not work in this context?

src/DADynamoApp/DADynamoApp.csproj
```diff
@@ -0,0 +1,47 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+    <ImportGroup Label="PropertySheets">
+        <Import Project="$(SolutionDir)Config\CS_SDK.props" />
+    </ImportGroup>
+    
+  <PropertyGroup>
+      <TargetFramework>net8.0-windows</TargetFramework>
```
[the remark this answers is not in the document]
pinzart90: yes, visual studio keeps putting it in here...

src/DADynamoApp/DADynamoApp.csproj
```diff
@@ -0,0 +1,47 @@
...
+  </ItemGroup>
+    
+    <ItemGroup>
+        <Reference Include="RevitAPI">
+            <HintPath>$(REVITAPI)\RevitAPI.dll</HintPath>
+            <Private>False</Private>
+        </Reference>
+    </ItemGroup>
+    <ItemGroup>
+        <PackageReference Include="Autodesk.Forge.DesignAutomation.Revit" Version="2025.0.1" />
```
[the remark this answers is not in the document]
pinzart90: seems like a 2026 version of this package was pushed two weeks ago. Not sure if the versions exactly match.

src/DADynamoApp/DADynamoApp.csproj
```diff
@@ -0,0 +1,47 @@
...
+        <PackageReference Include="DynamoVisualProgramming.Core" Version="3.3.0.6316" ExcludeAssets="runtime" GeneratePathProperty="true" />
+        <PackageReference Include="DynamoPlayer" Version="6.0.2" ExcludeAssets="runtime" GeneratePathProperty="true" />
+    </ItemGroup>
+    <ItemGroup>
+      <ProjectReference Include="..\Libraries\RevitServices\RevitServices.csproj" />
+    </ItemGroup>
+    <ItemGroup>
+        <Reference Include="DynamoPlayer.Models">
```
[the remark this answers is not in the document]
pinzart90: because the player package does not respect the dotnet norms. It is built like a revit nuget (weird dll paths)

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -0,0 +1,182 @@
...
+using RevitServices.EventHandler;
+using System.IO.Compression;
+using System.Reflection;
+using System.Runtime.Loader;
+using System.Diagnostics;
+
+namespace DADynamoApp
+{
+    [Autodesk.Revit.Attributes.Regeneration(Autodesk.Revit.Attributes.RegenerationOption.Manual)]
+    [Autodesk.Revit.Attributes.Transaction(Autodesk.Revit.Attributes.TransactionMode.Manual)]
```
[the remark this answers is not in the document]
pinzart90: I am not sure. @Mikhinja could you shed some light ? what would be better to use ?
Is automatic transaction mode auto triggered after every change ?

src/DADynamoApp/DAEntrypoint.cs
pinzart90: I guess, if the workers are stateless and thorw away.

src/DADynamoApp/DAEntrypoint.cs
pinzart90: that was my understanding, but does not hurt to cleanup

src/DADynamoApp/DAEntrypoint.cs
pinzart90: Is there a Revit API that gives us the ASC location?

src/Config/CS_SDK.props
```diff
@@ -1,11 +1,10 @@
...
+		<Platforms>NET70;NET80;NET80_DA</Platforms>
 		<PlatformTarget >x64</PlatformTarget>
-		<TargetFramework Condition="'$(Platform)' == 'NET80'">net8.0-windows</TargetFramework>
-		<TargetFramework Condition="'$(Platform)' == 'NET70'">net7.0-windows</TargetFramework>
```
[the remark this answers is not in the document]
Mikhinja: Confirm, removable.

src/DADynamoApp/DAEntrypoint.cs
Mikhinja: Yes, `Autodesk.Revit.ApplicationServices.ControlledApplication.SharedComponentsLocation`.

src/DADynamoApp/DADynamoApp.csproj
```diff
@@ -0,0 +1,47 @@
...
+    
+  <PropertyGroup>
+      <TargetFramework>net8.0-windows</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+  </PropertyGroup>
+    
+  <ItemGroup>
+    <Compile Remove="obj\**" />
```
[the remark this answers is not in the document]
pinzart90: old code from when I was still experimenting with this project, will remove

src/DADynamoApp/DAEntrypoint.cs
pinzart90: no reason, changing

src/DADynamoApp/DAEntrypoint.cs
pinzart90: As far as i know, AppDomain.CurrentDomain is the only API that can add an `UnhandledException ` handler

src/DADynamoApp/DAEntrypoint.cs
pinzart90: I could not figure out how to make the dynamo userDataFolder and commonDataFolder to work with the DA work directory both in cloud and locally (I probably just lost my patience)

src/DADynamoApp/DAEntrypoint.cs
pinzart90: maybe, but I'm not doing much with it 
Will remove it for now

src/DADynamoApp/DAEntrypoint.cs
pinzart90: I changed some of this, but I am loading python from a subfolder (wherever the DA engine unzips the pythonLibs zip)
Dynamo has an issue where it expects all python libs should be in the same folder as DynamoCore.dll (else it throws an error)

src/DADynamoApp/DAEntrypoint.cs
pinzart90: I don't have a use case for NewProjectDocument yet. 
Maybe if the user wants to generate data from scratch somehow.

src/DADynamoApp/DAEntrypoint.cs
pinzart90: reverted most of this

src/DADynamoApp/DAEntrypoint.cs
pinzart90: Changed most of this, I think I was refering to the current working directory

src/DADynamoApp/DAEntrypoint.cs
pinzart90: I am not checking this anymore

PR #3174, opened 2025-05-20
Malinaa/revit 236351/add validations D4R tests

MalinaAlice:
I added some new validations and some things I have to modify because the nodes were obsolete.


test/Libraries/RevitIntegrationTests/LengthTests.cs
```diff
@@ -30,10 +32,10 @@ public void Length()
             Assert.AreEqual(5, ViewModel.Model.CurrentWorkspace.Connectors.Count());
 
             //check length of String
-
-            Assert.AreEqual(GetPreviewValue("9311b424-7887-4332-be51-82795d3e3ce6"), -27.341145833333318d);
-
-
+            var length = GetPreviewValue("c0ecff8e98204341a73b04de6e99c90e");
+            var lengthValue = (length is KeyValuePair<string, object> kvp) ? kvp.Value : length;
```
Mikhinja, line 36: Why is the node value uncertain, when is it not a number and when is it a dictionary? Is it a KeyValuePair or is it a dictionary? Is the test graph not stable in its results?

Mikhinja (approved):

PR #3180, opened 2025-06-04
Regression test to check for new Dynamo .dll files

teodorR:
Create a regression test to fail if new .dll files are added. It should contain a .txt to keep the file list, so that it is easy to update. It needs to be run when integrating a new Dynamo Core version.

### Purpose

(FILL ME IN) This section describes why this PR is here. Usually it would include a reference 
to the tracking task that it is part or all of the solution for.

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/test/Libraries/RevitIntegrationTests/RegressionTests.cs
+++ b/test/Libraries/RevitIntegrationTests/RegressionTests.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Text;
 using Dynamo.Applications.Models;
 using Dynamo.Graph.Nodes;
 using Dynamo.Graph.Workspaces;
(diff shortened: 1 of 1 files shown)


test/Libraries/RevitIntegrationTests/RegressionTests.cs
```diff
@@ -268,9 +270,88 @@ public void NoDuplicatePortsOnUINodes()
...
-        // Helper Functions
+      [Test]
+      public void NoUnexpectedDlls_ShouldExist()
+      {
+         // Folder of the test
+         string testAssemblyPath = Assembly.GetExecutingAssembly().Location;
+         string testDirectory = Path.GetDirectoryName(testAssemblyPath);
+
+         // Revit addin build output where the dlls are located
+         string relativeBuildPath = Path.Combine(testDirectory, @"..\..\..\..\..\..\Revit\Debugx64\Addins\DynamoForRevit");
```
vpavel04, line 281: can we start from the path of current executable? This test does not work if we compile Releasex64 or other configuration

vpavel04 (changes requested):

vpavel04 (approved):

PR #3252, opened 2025-11-12
Add cloud model support

pinzart90:
What changed:
Ability to load Revit Cloud Models ("type":"versions:autodesk.bim360:C4RModel")

Documentation 
https://aps.autodesk.com/blog/new-sample-support-access-revit-cloud-model-revit-design-automation-engine 

Not implemented yet: Cloud Shared models.
Does not work for regular rvt files stored on ACC (only supported in Autocad for now) - https://aps.autodesk.com/en/docs/design-automation/v3/developers_guide/reference-downloading/

TODO:
To access cloud models using Automation, you need to provide the user context through a special workitem argument; adsk3LeggedToken, scope=code:all needs to be added to your authorization call.
I need to add an example in the scripts/postman collections

### DIFF
--- a/src/DADynamoApp/DAEntrypoint.cs
+++ b/src/DADynamoApp/DAEntrypoint.cs
@@ -237,41 +237,40 @@ namespace DADynamoApp
             Console.WriteLine($"{nameof(saveRvt)} is set to {saveRvt}");
             if (saveRvt)
             {
-                if (doc.IsModelInCloud)
+                try
                 {
-                    if (doc.IsWorkshared) // work-shared/C4R model
+                    if (doc.IsModelInCloud)
                     {
-                        // Syncronize with central
-                        SynchronizeWithCentralOptions swc = new SynchronizeWithCentralOptions();
-                        swc.SetRelinquishOptions(new RelinquishOptions(/*relinquishAll*/true));// Should this be configurable?
-                        doc.SynchronizeWithCentral(new TransactWithCentralOptions(), swc);
-                    }
-                    else 
-                    {// Single user cloud model
-                        var newLoc = setupReq?.SaveCloudModelLocation;
-                        if (newLoc != null)
+                        if (doc.IsWorkshared) // work-shared/C4R model
                         {
-                            doc.SaveAsCloudModel(newLoc.AccountId, newLoc.ProjectId, newLoc.FolderId, newLoc.ModelName ?? Path.GetFileName(doc.PathName));
+                            // Syncronize with central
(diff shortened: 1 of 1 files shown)


mjkkirschner:
I'm confused by the linked docs - and how you're saving the model.

It appears that the Revit cloud model (workshared model?) should usually be synced / published - but in this PR you save a new copy of it to the output - is that correct - that might be a totally reasonable thing for us to do here - but it seems it could be unexpected for a Revit user - right?

At this point I am mostly just trying to understand all these different terms and workflows for Revit cloud models.

pinzart90:
> I'm confused by the linked docs - and how you're saving the model.
> 
> It appears that the Revit cloud model (workshared model?) should usually be synced / published - but in this PR you save a new copy of it to the output - is that correct - that might be a totally reasonable thing for us to do here - but it seems it could be unexpected for a Revit user - right?
> 
> At this point I am mostly just trying to understand all these different terms and workflows for Revit cloud models.

I updated the PR to handle cloud shared models. It is still not clear to me how to handle conflicts with central (in the example it shows that the DA model should relinquish all changes)

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -133,10 +131,38 @@ public void HandleDesignAutomationReadyEvent(object sender, DesignAutomationRead
...
+            SetupDARequest? setupReq = null;
+            var setupReqPath = Path.Combine(WorkItemFolder, "setup.json");
+            if (File.Exists(setupReqPath))
+            {
+                try
+                {
+                    var setupRequest = File.ReadAllText(setupReqPath);
+                    Console.WriteLine(setupRequest);
+
+                    setupReq = JsonConvert.DeserializeObject<SetupDARequest>(setupRequest, new JsonSerializerSettings() { TypeNameHandling = TypeNameHandling.Auto });
```
mjkkirschner: is the `setup.json` fully under our control?

src/DADynamoApp/SetupDARequest.cs
```diff
@@ -10,7 +57,19 @@ internal class SetupDARequest
         //
         // Summary:
         // Save the revit document to the default result.rvt file.
-        [JsonProperty("saveRvt", Required = Required.DisallowNull, NullValueHandling = NullValueHandling.Ignore)]
-        public bool SaveRvt { get; set; } = false;
```
mjkkirschner: is this class something you have designed or is it something inherited by us from DA?

PR #3256, opened 2025-11-17
DYN-9845 : set categories value from id

BogdanZavu:
### Purpose
We want to be able to use the category built-in id if possible.
The usage context here is that the AI ( through DynamoMCP ) is aware of the built-in ids but it is more error prone when it comes to categories names since those are localized.

### Declarations

### Reviewers
@Mikhinja 

### FYIs
@johnpierson

### DIFF
--- a/src/Libraries/RevitNodesUI/RevitDropDown.cs
+++ b/src/Libraries/RevitNodesUI/RevitDropDown.cs
@@ -745,14 +745,8 @@ namespace DSRevitNodesUI
                 //In fact this should be the way to do it since display names can change with localization.
                 if (SelectedIndex < 0)
                 {
-                    var allCategories = Enum.GetValues(typeof(BuiltInCategory))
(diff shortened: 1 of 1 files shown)


src/Libraries/RevitNodesUI/RevitDropDown.cs
```diff
@@ -732,9 +732,36 @@ protected override bool UpdateValueCore(UpdateValueParams updateValueParams)
...
+                catch
+                {
+                    SelectedIndex = -1;
+                }
+
+                //Value could also refer to BuiltInCategory enum name.
+                //In fact this should be the way to do it since display names can change with localization.
+                if (SelectedIndex < 0)
+                {
+                    var allCategories = Enum.GetValues(typeof(BuiltInCategory))
```
Mikhinja: Why not use a variant of `Enum.Parse(value, true)` that specifies to ignore case?

PR #3306, opened 2026-03-29
Default Output folder

pinzart90:
### Purpose

Add default output folder

### Declarations

### Reviewers

### FYIs

### DIFF
--- a/src/DADynamoApp/DAEntrypoint.cs
+++ b/src/DADynamoApp/DAEntrypoint.cs
@@ -281,9 +281,7 @@ namespace DADynamoApp
             DocumentManager.Instance.PrepareForDesignAutomation(app);
 
             var loadedLibGVersion = ASMPrealoaderUtils.PreloadAsmFromRevit(asmLocation, DynamoPath);
-            Console.WriteLine($"{loadedLibGVersion}");
             var geometryFactoryPath = ASMPrealoaderUtils.GetGeometryFactoryPath(DynamoPath, loadedLibGVersion);
-            Console.WriteLine($"{geometryFactoryPath}");
 
             PreInstallPythonDependencies();
 
--- a/src/DADynamoApp/DAEntrypoint.cs
+++ b/src/DADynamoApp/DAEntrypoint.cs
@@ -444,7 +444,7 @@ namespace DADynamoApp
             }
             catch (Exception ex)
             {
-                Console.WriteLine($"Failed to create output folder. {ex.Message}");
+                Console.WriteLine($"Failed to delete the empty output folder. {ex.Message}");
             }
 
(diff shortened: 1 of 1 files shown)


src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -214,11 +214,13 @@ public void HandleDesignAutomationReadyEvent(object sender, DesignAutomationRead
...
             // Ensure we have a pre-built graph output folder.
+            var graphOutputFolder = Path.Combine(WorkItemFolder, "output");
             try
             {
-                var graphOutputFolder = "output";
+                Console.WriteLine("Checking for output folder");
                 if (!Directory.Exists(graphOutputFolder))
                 {
+                    Console.WriteLine("Output folder does not exist. Creating..");
                     Directory.CreateDirectory(graphOutputFolder);
```
awigginsadsk, line 229: What happens if CreateDirectory fails here?

src/DADynamoApp/DAEntrypoint.cs
```diff
@@ -433,6 +433,20 @@ public void HandleDesignAutomationReadyEvent(object sender, DesignAutomationRead
...
+                // If the out put folder exists and is empty, then delete it so we don't generate empty output zip files.
+                if (Directory.Exists(graphOutputFolder) && !Directory.EnumerateFileSystemEntries(graphOutputFolder).Any())
+                {
+                    Console.WriteLine("The output folder is empty.");
+                    Directory.Delete(graphOutputFolder);
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Failed to create output folder. {ex.Message}");
```
awigginsadsk: Shouldn't this be "Failed to *delete* output folder"? Or am I missing some context?

awigginsadsk (changes requested):

src/DADynamoApp/DAEntrypoint.cs
pinzart90: The output folder is basically a feature that we provide to DAAS_DA users.
It offers an out of the box place where graphs can produce data/content (ex. images, xml reports etc). If creation of the folder fails, then most likely the graph execution will have nodes that fail and will report those nodes back to the user. ALso the output.zip file will not be sent back to the user.
Because it has a very low chance of failing, I thought a log failure should suffice. This ootb "output" folder feature may be scrapped for something more generic in the future

src/DADynamoApp/DAEntrypoint.cs
pinzart90: yes it should. Thanks !

awigginsadsk (approved):

PR #3431, opened 2026-09-03
Fix typo: rename ScheduleFilter.FiledId to FieldId (#3430)

### LINKED ISSUE #3430: Typographic Error from Dynamo 3.3.0.6316 - ScheduleFilter.FiledId
### Dynamo Version

3.3.0.6316 

### Host

Revit 2025

### Operating System

Microsoft Windows 11

### What did you do?

The out of the box node ScheduleFilter.FiledId is incorrectly named. This should be FieldId

### What did you expect to see?

The node will provide the FieldId as an int. 

### What did you see instead?

Name of node reports a FiledId

johnpierson:
### Purpose

Fixes #3430.

The out-of-the-box node property `Revit.Schedules.ScheduleFilter.FiledId` is misspelled. It should be `FieldId`.

### Changes

- Rename the public property `FiledId` → `FieldId` in `ScheduleFilter.cs`, and update the internal uses (`ToString`, `Equals`, `GetHashCode`). The `ToString` label already read `FieldId:`, so runtime output is unchanged.
- Add a `<priorNameHint>` migration (`Revit.Schedules.ScheduleFilter.FiledId` → `Revit.Schedules.ScheduleFilter.FieldId`) so existing graphs referencing the old node signature remap automatically. This keeps the rename backward-compatible rather than breaking saved graphs.
- Rename the node help `.dyn`, its example image, and the fallback docs across all locales to the corrected name, and update their contents (also corrected a stray `ScheduleFilter.FileId` reference in the en-US doc).

### Notes

Pure rename of a public API surface, guarded by a migration entry that follows the existing property-rename precedent in `RevitNodes.Migrations.xml` (e.g. `FamilySymbol.Name` → `FamilyType.Name`).


Mikhinja (approved):

mjkkirschner:
This breaks the API for c# package authors if this does not go into a new major version release - is it intended for a new year? (maybe yes because it's to master?)

Would be good to have copilot reviews on this repo - and to teach it about API compatibility and when Revit breaks api compat.

johnpierson:
I feel like there is no way C# package authors are using this method as they would be using raw Revit API when authoring in C#. And if they are, it is pretty poor practice honestly. This API call specifically is very very minor and if people are working with fields within schedules in Revit, even smaller.

While it may be breaking, I think that is fine as I really do not see a lot of folks using this as is.

mjkkirschner:
@johnpierson there may be other consequences to an API break like this than just this single method not working. It would be good to test it thoroughly with a package that references the old version of this binary but gets loaded in a version of Revit that bundles the new version. (the whole binary or types in some cases may not load)

